PR #9, opened 2019-04-25
CI updates

chfast:


circle.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - checkout
       - run:
           name: "Update submodules"
-          command: git submodule update --init
+          command: git submodule update --init --recursive
```
axic: You know https://github.com/chfast/intx/issues/68 would solve this :wink:

chfast: Right, maybe will do it before first intx release.

PR #21, opened 2019-04-30
Make public header compatible with C 

axic:
Can squash commits if needed. Most of this is from Hera.


include/evmone/evmone.h
```diff
@@ -2,8 +2,20 @@
 // Copyright 2018-2019 The evmone Authors.
 // Licensed under the Apache License, Version 2.0.
 
+#ifndef __EVMONE_H
```
chfast: Remove the `__` prefix.

axic: Done.

PR #24, opened 2019-04-30
Implement SELFBALANCE opcode from EIP-1884

axic:
Ref https://eips.ethereum.org/EIPS/eip-1884

Need to actually add all the changes to EVMC apparently.

### DIFF
--- a/test/unittests/evm_state_test.cpp
+++ b/test/unittests/evm_state_test.cpp
@@ -214,6 +214,29 @@ TEST_F(evm_state, balance)
     EXPECT_EQ(result.output_data[5], 0x01);
 }
 
+TEST_F(evm_state, selfbalance)
+{
+    host.accounts[msg.destination].set_balance(0x0504030201);
+    // NOTE: adding push here to balance out the stack pre-Istanbul (needed to get undefined
+    // instruction as a result)
+    auto code = bytecode{} + push(1) + OP_SELFBALANCE + mstore(0) + ret(32 - 6, 6);
+
+    rev = EVMC_CONSTANTINOPLE;
+    execute(code);
+    EXPECT_EQ(result.status_code, EVMC_UNDEFINED_INSTRUCTION);
+
+    rev = EVMC_ISTANBUL;
+    execute(code);
+    EXPECT_GAS_USED(EVMC_SUCCESS, 23);
+    ASSERT_EQ(result.output_size, 6);
+    EXPECT_EQ(result.output_data[0], 0);
+    EXPECT_EQ(result.output_data[1], 0x05);
+    EXPECT_EQ(result.output_data[2], 0x04);
+    EXPECT_EQ(result.output_data[3], 0x03);
+    EXPECT_EQ(result.output_data[4], 0x02);
+    EXPECT_EQ(result.output_data[5], 0x01);
+}
+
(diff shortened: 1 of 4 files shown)


lib/evmone/execution.cpp
```diff
@@ -1518,6 +1525,7 @@ exec_fn_table create_op_table_petersburg() noexcept
 exec_fn_table create_op_table_istanbul() noexcept
 {
     auto table = create_op_table_petersburg();
+    table[OP_BALANCE] = op_selfbalance;
```
axic: Need to use the correct opcode.

chfast: Just use `0x46`.

axic: Well, the bigger work left for 1884 is repricing the opcodes and that requires a change in EVMC, because the instruction metrics live there, if I'm not wrong?

chfast:
> The invalid opcode test is failing because `0x46` is not defined in `evm_instruction_names`.

Add `if` in the test.

axic:
Will make this one depend on https://github.com/ethereum/evmc/pull/372

axic:
@chfast @gumb0 any idea why this fails with segfault?

chfast:
> @chfast @gumb0 any idea why this fails with segfault?

evmone fails because there is not instruction table for Berlin but EVMC_MAX_REVISION points to Berlin now.

axic:
Rebased and fixed the previous issues, however now for some reason `stack underflow` is reported where it should report `unknown instruction`.

gumb0:
It just checks stack requirements earlier (at the beginning of the block) than instruction returns `unknown instruction`, and the stack is empty.
I think if you add push before, it will be fixed.

gumb0:
Rebased.

test/unittests/evm_state_test.cpp
```diff
@@ -177,6 +177,27 @@ TEST_F(evm_state, balance)
     EXPECT_EQ(result.output_data[5], 0x01);
 }
 
+TEST_F(evm_state, selfbalance)
+{
+    accounts[msg.destination].set_balance(0x0504030201);
+    auto code = bytecode{} + push(1) + OP_SELFBALANCE + mstore(0) + ret(32 - 6, 6);
```
axic: There's no argument to selfbalanace.

chfast: But when `OP_SELFBALANCE` is undefined, there is an argument missing to `mstore(0)`.

axic: But shouldn't it just abort because of invalid instruction?

chfast: Only in naive implementation of EVM. Here we check stack requirements for the whole block of instructions before executing them.

axic: I get that, but the concept is wrong then. evmone fails with a strange error code on invalid input then?

chfast: Not sure this concept you mean, but yes, it can return other error code.

axic: This suggest to me that evmone cannot handle badly formatted EVM input, which has a stack underflow.

chfast:
Please add CHANGELOG entry.

PR #31, opened 2019-05-06
Test utils

### LINKED ISSUE #30: Improve from_hex()
- [ ] Make it case insensitive (support `A-F` chars).
- [ ] Throw exception on invalid input
- [ ] Add unit tests

chfast:
Closes #30


test/utils/utils.cpp
```diff
@@ -10,6 +10,37 @@
 #include <iomanip>
 #include <iostream>
 
+bytes from_hex(std::string_view hex)
```
axic: Would it be an OK idea moving this to evmc so it can also be used in Hera?

Or perhaps creating yet another library? :D

chfast: Not sure... I'd rather copy&paste sources...

PR #41, opened 2019-05-07
Use improved intx and EVMC API

chfast:


lib/evmone/execution.cpp
```diff
@@ -996,7 +969,7 @@ void op_call(execution_state& state, instr_argument arg) noexcept
         msg.gas += 2300;  // Add stipend.
     }
 
-    auto result = state.host->host->call(state.host, &msg);
+    auto result = state.host.call(msg);
     state.return_data.assign(result.output_data, result.output_size);
```
axic: Can you group `if (result.output_data)` and this line?

PR #49, opened 2019-05-15
Upgrade intx to 0.2

chfast:


lib/evmone/execution.cpp
```diff
@@ -212,20 +212,19 @@ void op_iszero(execution_state& state, instr_argument) noexcept
 
 void op_and(execution_state& state, instr_argument) noexcept
 {
-    // TODO: Add operator&= to intx.
-    state.item(1) = state.item(0) & state.item(1);
+    state.item(1) &= state.item(0);
```
axic: Tiny speedups incoming? :)

chfast: No. The intx API cleanup.

axic: Yeah but potentially intx could do the `and` in place without copies -> speed up if sufficiently large number of ands/shifts/etc. are done.

chfast: Maybe there is some space for improvement, but `x &= y` is defined `x = x & y` do no functional change here. Syntax sugar only.

axic: But it would be possible to improve it in `intx` without the need to change it here, after this was merged.

chfast: Yes, that's correct.

PR #56, opened 2019-05-20
utils: Add hexx encoding format

chfast:


axic:
This is hexadecimal RLE 😉

chfast:
> This is hexadecimal RLE wink

Any better name suggestions?

PR #57, opened 2019-05-20
test: Add solidity ABI decoder unit test

chfast:


test/unittests/execution_test.cpp
```diff
@@ -1635,4 +1635,15 @@ TEST_F(execution, memory_access)
             }
         }
     }
+}
+
+TEST_F(execution, solidity_abi_decoder)
+{
```
bshastry: This is not the bytecode of the solidity abi decoder. It is a simple contract that contains a function that returns 1.

chfast: ok. I'm not going to merge it. I don't want unit test complicated like this one.

PR #60, opened 2019-05-22
Coverage of state tests

chfast:


chfast:
State tests coverage already included in codecov.

PR #77, opened 2019-06-27
Use new evmc::result constructor

chfast:


chfast:
Rebased. I think we can include this.

axic:
Is this speed critical though? Could this influence benchmarks a lot?

chfast:
> Is this speed critical though? Could this influence benchmarks a lot?

It should not matter - EVMC is doing the same thing except it checks the result of malloc(). Benchmarks show not performance difference.

PR #85, opened 2019-07-03
EVM test tool

chfast:
This adds EVM testing tool for EVMC-compatible VM implementations.
It is build out of generic evmone execution tests.

E.g.
```
evm-test ./aleth-interpreter.so
```

Do not confuse with evmone-unittests which contains these test + internal unit tests and has evmone linked "statically".

The name `evmone-test` was also considered.

TODO:
- [x] Extend EVMC C++ API
- [x] Add entry in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,18 @@ bin/evmone-unittests
 bin/evmone-bench
 ```
 
+### Tools
+
+#### evm-test
+
+The **evm-test** executes a collection of unit tests on 
+any EVMC-compatible Ethereum Virtual Machine implementation.
+The collection of tests comes from the evmone project.
+
+```bash
+evm-test ./evmone.so
+```
+
 ## Maintainer
 
 Paweł Bylica [@chfast]
--- a/test/unittests/main.cpp
+++ b/test/unittests/main.cpp
@@ -1,0 +1,179 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2019 The evmone Authors.
+// Licensed under the Apache License, Version 2.0.
+
+#include "vm_loader.hpp"
+#include <evmc/loader.h>
+#include <gtest/gtest.h>
+#include <iostream>
+#include <string>
+#include <vector>
+
+/// The loaded EVMC module.
+static evmc::vm evmc_module;
+
+evmc::vm& get_vm() noexcept
+{
+    return evmc_module;
+}
+
+/// Simple and copy&paste distributable CLI parser.
+///
+/// TODO: Originally taken from EVMC and modified here. Copy it back.
+class cli_parser
+{
+public:
+    using preprocessor_fn = void (*)(int*, char**);
+
+    const char* const application_name = nullptr;
+    const char* const application_version = nullptr;
+    const char* const application_description = nullptr;
+
+    std::vector<std::string> arguments_names;
+    std::vector<std::string> arguments;
+
+    preprocessor_fn preprocessor = [](int*, char**) {};
+
+    cli_parser(const char* app_name, const char* app_version, const char* app_description,
+        std::vector<std::string> args_names) noexcept
+      : application_name{app_name},
+        application_version{app_version},
+        application_description{app_description},
+        arguments_names{std::move(args_names)}
+    {
+        arguments.reserve(this->arguments_names.size());
+    }
+
+    /// Sets the preprocessor and enables preprocessing.
+    ///
+    /// The preprocessor runs on provided arguments before the parsing is done.
+    /// It is allowed to modify the arguments and/or generate other output.
+    ///
+    /// @param fn The preprocessor function.
+    void set_preprocessor(preprocessor_fn fn) noexcept { preprocessor = fn; }
+
+    /// Parses the command line arguments.
+    ///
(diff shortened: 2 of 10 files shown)


axic:
Any way moving this back to evmc?

chfast:
> Any way moving this back to evmc?

That would be very inconvenient for developing evmone as you don't want to have unit tests in other project.

axic:
The name is misleading then, should be `evmone-test`.

chfast:
> The name is misleading then, should be `evmone-test`.

It tests EVMs.

axic:
So it is not evmone specific then?

chfast:
> So it is not evmone specific then?

Description updated.

axic:
So then my question is reasonable: any possibility moving this out of evmone at some point to aid other VMs using evmc?

chfast:
> So then my question is reasonable: any possibility moving this out of evmone at some point to aid other VMs using evmc?

The answer is still no. The `evm-test` tool contributes to the EVMC ecosystem, but it would be very inconvenient to have these tests in some other repo. Also, it might be bad to extend EVMC with "testing" feature.

I'm going to get the reference to this tool from ethereum/testing.

axic:
So it is basically a "state test over evmc" tool. Still, it would make sense to have it outside of evmone to encourage other EVMC compatible EVMs to use it. Not saying it has to be in evmc, since it may not belong there.

chfast:
> So it is basically a "state test over evmc" tool. Still, it would make sense to have it outside of evmone to encourage other EVMC compatible EVMs to use it. Not saying it has to be in evmc, since it may not belong there.

Yes it make sense for other projects but not for this project. And is not worth the trouble and my time at this point.

README.md
```diff
@@ -43,6 +43,18 @@ bin/evmone-unittests
 bin/evmone-bench
 ```
 
+### Tools
+
+#### evm-test
```
axic: At the bare minimum extending this section on build instruction if it is possible to only build this without evmone to be used by other projects would be useful.

chfast: This is not possible right now. I need another round of CMake changes to re-tune build options.

test/unittests/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+        auto cli = cli_parser{"EVM Test", PROJECT_VERSION,
+            "Testing tool for EVMC-compatible Ethereum Virtual Machine implementations.\n"
+            "Powered by the evmone project.\n\n"
+            "EVMC:   https://github.com/ethereum/evmc\n"
+            "evmone: https://github.com/ethereum/evmone",
+            {"MODULE"}};
+        cli.set_preprocessor(testing::InitGoogleTest);
+
+        if (const auto error_code = cli.parse(argc, argv, std::cout, std::cerr); error_code <= 0)
```
gumb0: cool, C++17

test/unittests/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+            return error_code;
+
+        const auto& evmc_config = cli.arguments[0];
+        evmc_loader_error_code ec;
+        evmc_module = evmc::vm{evmc_load_and_configure(evmc_config.c_str(), &ec)};
+
+        if (ec != EVMC_LOADER_SUCCESS)
+        {
+            if (const auto error = evmc_last_error_msg())
```
gumb0: This could be just
```
const auto error = evmc_last_error_msg();
std::cerr << "EVMC loading error: " << (error ? error : ec) << "\n";
```

test/unittests/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+#include "vm_loader.hpp"
+#include <evmc/loader.h>
+#include <gtest/gtest.h>
+#include <iostream>
+#include <string>
+#include <vector>
+
+/// The loaded EVMC module.
+static evmc::vm evmc_module;
```
gumb0: It would be less confusing if this and `get_evm` below were in a separate cpp file, then it would be more clear that it's another implementation for `get_evm` of `vm_loader.hpp`
(some setter I guess would be required then, too)

test/unittests/main.cpp
```diff
@@ -0,0 +1,179 @@
+// evmone: Fast Ethereum Virtual Machine implementation
```
gumb0: It would probably be better if this file were in a separate directory, not `unittests`?

test/unittests/main.cpp
chfast: The `error` is a `char*`.

test/unittests/main.cpp
chfast: It was planned like that, but it gets tricky because `evmc_module` is set from `main()`.

test/unittests/main.cpp
chfast: There are more changes to dir structure needed, so I left it for another time.

Preferably the `evm-test` should be in `tools/`, not in `test/` but because the `evm-unitests` is shared between `evm-test` and `evmone-unittests` it might be overkill.

Maybe just get rid of `test/` dir entirely...

test/unittests/main.cpp
gumb0: Ah I see

PR #89, opened 2019-07-11
Improve evm.memory_access test

gumb0:


gumb0:
@chfast 2Gb + 2Gb case fails currently for evmone, too

gumb0:
Passes with 30M gas

test/unittests/evm_test.cpp
```diff
@@ -1171,16 +1173,24 @@ TEST_F(evm, memory_access)
 
             code += bytecode{t.opcode};
 
-            execute(code);
+            auto const gas = 30000000;
```
chfast: Can you try `8796294610952`. That's the max value causing all test to go OOG

gumb0: evmone passes, interpreter returns `EVMC_INVALID_MEMORY_ACCESS`

chfast: Is this solved now?

gumb0: No, interpreter checks the size of returndata first and returns `EVMC_INVALID_MEMORY_ACCESS` before checking for required memory size

chfast: Can we apply something like this then? https://github.com/ethereum/evmone/pull/92

It also looks `RETURNDATACOPY` might require its own test for `OUT_OF_GAS` not being `INVALID_MEMORY_ACCESS`.

gumb0: Yes that, helps.

I rebased this PR on #92

chfast: Ready now?

gumb0: Yes

PR #91, opened 2019-07-15
Push arguments to stack in undefined_instructions test.

gumb0:
aleth-interpreter first checks whether there's enough arguments for instruction (according to latest revision requirements) before checking whether this instruction is defined for current revision.

So this was failing for some instructions like bit shifts, that were defined in later revisions.

### DIFF
--- a/test/unittests/evm_test.cpp
+++ b/test/unittests/evm_test.cpp
@@ -946,18 +946,22 @@ TEST_F(evm, shift_overflow)
 
 TEST_F(evm, undefined_instructions)
 {
+    auto latest_metrics = evmc_get_instruction_metrics_table(EVMC_MAX_REVISION);
+
     for (auto i = 0; i <= EVMC_MAX_REVISION; ++i)
     {
-        auto r = evmc_revision(i);
-        auto names = evmc_get_instruction_names_table(r);
+        rev = evmc_revision(i);
+        auto names = evmc_get_instruction_names_table(rev);
 
         for (uint8_t opcode = 0; opcode <= 0xfe; ++opcode)
         {
             if (names[opcode] != nullptr)
                 continue;
 
-            auto res = vm.execute(*this, r, {}, &opcode, sizeof(opcode));
-            EXPECT_EQ(res.status_code, EVMC_UNDEFINED_INSTRUCTION) << hex(opcode);
+            auto code = latest_metrics[opcode].num_stack_arguments * push(0) + evmc_opcode(opcode);
+
+            execute(30000000, code);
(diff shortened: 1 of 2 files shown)


test/unittests/evm_test.cpp
```diff
@@ -946,18 +946,22 @@ TEST_F(evm, shift_overflow)
...
         for (uint8_t opcode = 0; opcode <= 0xfe; ++opcode)
         {
             if (names[opcode] != nullptr)
                 continue;
 
-            auto res = vm.execute(*this, r, {}, &opcode, sizeof(opcode));
-            EXPECT_EQ(res.status_code, EVMC_UNDEFINED_INSTRUCTION) << hex(opcode);
+            auto code = latest_metrics[opcode].num_stack_arguments * push(0) + evmc_opcode(opcode);
+
+            execute(30000000, code);
```
gumb0: I haven't dig into this too much yet, but it seemed that with default gas (which is `std::numeric_limits<int64_t>::max()` analyze phase has some kind of integer overflow (because for undefined opcodes gas price is defined as `-1`), then execution end with `EVMC_OUT_OF_GAS` instead of `EVMC_UNDEFINED_INSTRUCTION`

chfast: That should be noticeable if you build with UBSan, `-DSANITIZE=undefined`.

chfast:
> aleth-interpreter first checks whether there's enough arguments for instruction (according to latest revision requirements) before checking whether this instruction is defined for current revision.

Now I think we might want to change that in aleth-interpreter. It should be rather simple if it uses `evmc::instructions` tables.

test/unittests/evm_test.cpp
gumb0: Yeah it reports it
```
/home/andrei/dev/evmone/lib/evmone/instructions.cpp:1164:25: runtime error: signed integer overflow: 9223372036854775807 - -1 cannot be represented in type 'long int'
```

gumb0:
@chfast I've made a fix here for integer overflow in analysis.cpp. If it looks ok, I can submit it in a separate PR

chfast:
> @chfast I've made a fix here for integer overflow in analysis.cpp. If it looks ok, I can submit it in a separate PR

Create a new PR with a new unit tests that shows the UBSan warning (CI already has UBSan builds). Then add another commit with the fix.

I believe we should also change `-1` to `0` in EVMC so create an issue for it.

gumb0:
https://github.com/ethereum/evmone/pull/93

gumb0:
Fixed in aleth-interpreter https://github.com/ethereum/aleth/pull/5679

PR #93, opened 2019-07-16
Fix integer overflow in analysis around gas cost of undefined instruction

gumb0:
Without the fix it's not only UBSAN warning, but also the test fails, returning `EVMC_OUT_OF_GAS` instead of `EVMC_UNDEFINED_INSTRUCTION`.

Overflow happens not at the place of fix but later during execution here
https://github.com/ethereum/evmone/blob/47a1d47ae96b19a247d3923311735cce118a347b/lib/evmone/instructions.cpp#L1164

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -84,7 +84,8 @@ code_analysis analyze(
         auto& instr = jumpdest ? analysis.instrs.back() : analysis.instrs.emplace_back(fns[c]);
 
         auto metrics = instr_table[c];
-        block->gas_cost += metrics.gas_cost;
+        if (metrics.gas_cost > 0)  // can be -1 for undefined instruction
+            block->gas_cost += metrics.gas_cost;
         auto stack_req = metrics.num_stack_arguments - block->stack_diff;
         block->stack_diff += (metrics.num_stack_returned_items - metrics.num_stack_arguments);
         block->stack_req = std::max(block->stack_req, stack_req);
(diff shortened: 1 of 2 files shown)


chfast:
I found similar test in the `fuzzer` branch. After cleaning up here is the unit test: https://github.com/ethereum/evmone/commit/27ef5b04cf66e9e31b72de5e679090646bec4839. Please integrate with this PR.

lib/evmone/analysis.cpp
```diff
@@ -84,7 +84,8 @@ code_analysis analyze(
         auto& instr = jumpdest ? analysis.instrs.back() : analysis.instrs.emplace_back(fns[c]);
 
         auto metrics = instr_table[c];
-        block->gas_cost += metrics.gas_cost;
+        if (metrics.gas_cost > 0)  // can be -1 for undefined instruction
```
axic: If we know this instruction will be hit, shouldn't the cost be set to max instead (e.g. out of gas)?

gumb0: I think then it would stop with `EVMC_OUT_OF_GAS` at the beginning of the block, instead of expected `EVMC_UNDEFINED_INSTRUCTION`

chfast: I believe we should set it to `0` as this is real gas cost of executing it. Setting it to "max" is a nice hack, but I don't waste 64-bits in the table just for it.

chfast: https://github.com/ethereum/evmc/issues/356.

PR #94, opened 2019-07-16
utils: Decode bytecode into human-readable string

chfast:
Adds a procedure to convert a bytecode into a human-readable bytecode expression which is also compilable  back to bytecode with bytecode helpers.

TODO:
- [x] Add basic unit tests.

### DIFF
--- a/test/unittests/bytecode_test.cpp
+++ b/test/unittests/bytecode_test.cpp
@@ -48,3 +48,27 @@ TEST(bytecode, repeat)
 
     EXPECT_EQ(0 * OP_STOP, "");
 }
+
+TEST(bytecode, decode)
+{
+    const auto code = push(0x01e240) + OP_DUP1 + OP_GAS + "cc" + OP_REVERT;
+    EXPECT_EQ(decode(code, EVMC_FRONTIER),
+        "bytecode{} + OP_PUSH3 + \"01e240\" + OP_DUP1 + OP_GAS + \"cc\" + \"fd\"");
+    EXPECT_EQ(decode(code, EVMC_BYZANTIUM),
+        "bytecode{} + OP_PUSH3 + \"01e240\" + OP_DUP1 + OP_GAS + \"cc\" + OP_REVERT");
+}
+
+TEST(bytecode, decode_push_trimmed_data)
+{
+    const auto code1 = bytecode{} + OP_PUSH2 + "0000";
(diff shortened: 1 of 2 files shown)


test/utils/bytecode.hpp
```diff
@@ -163,4 +163,35 @@ inline bytecode sstore(bytecode index, bytecode value)
...
+        if (const auto name = names[opcode]; name)
+        {
+            s += std::string{" + OP_"} + name;
+
+            if (opcode >= OP_PUSH1 && opcode <= OP_PUSH32)
+            {
+                auto push_data = std::string{};
+                auto push_data_size = opcode - OP_PUSH1 + 1;
+                while (push_data_size-- && ++it != bytecode.end())
+                    push_data += to_hex({&*it, 1});
```
gumb0: Would be nice to have `to_hex` overload for a single byte parameter

gumb0: On the other hand, couldn't you do something like
```
auto push_data_start = ++it;
push_data_size = std::min(push_data_size, bytecode.end() - push_data_start);
push_data += to_hex({&*push_data_start, push_data_size});
it += push_data_size;
```
maybe too complicated...

test/unittests/bytecode_test.cpp
```diff
@@ -48,3 +48,27 @@ TEST(bytecode, repeat)
...
+    const auto code1 = bytecode{} + OP_PUSH2 + "0000";
+    EXPECT_EQ(decode(code1, EVMC_FRONTIER), "bytecode{} + OP_PUSH2 + \"0000\"");
+
+    const auto code2 = bytecode{} + OP_PUSH2 + "00";
+    EXPECT_EQ(decode(code2, EVMC_FRONTIER), "bytecode{} + OP_PUSH2 + \"00\"");
+
+    const auto code3 = bytecode{} + OP_PUSH2;
+    EXPECT_EQ(decode(code3, EVMC_FRONTIER), "bytecode{} + OP_PUSH2");
+
+    const auto code4 = bytecode{} + OP_PUSH2 + "";
```
gumb0: How is this different from `code3` ?

chfast: Not really different. Just wanted to show the `""` will be dropped. In previous implementation `""` was added for every push with empty data.

test/utils/bytecode.hpp
chfast: Great change actually. Allowed me to remove the additional iterator end check.

PR #95, opened 2019-07-17
cmake: Build evmone-standalone library on Linux

chfast:


bshastry:
This works as expected in my fuzzing setup.

PR #96, opened 2019-07-17
cmake: Use evmc-vmtester helper

chfast:


axic:
@chfast is this still relevant?

chfast:
This works. I think it is fine to include it.

PR #98, opened 2019-07-18
Implement EIP-663 variant C

### LINKED ISSUE #84: Prototype EIP-663 (unlimited swapdup)
See https://eips.ethereum.org/EIPS/eip-663

It would be nice to see some benchmarks in order to decide between those options.

chfast:
This implement the variant C of the [EIP-663](https://eips.ethereum.org/EIPS/eip-663) by adding 4 new instructions for EVM stack manipulation: `DUPN`, `SWAPN`, `DUPSN`, `SWAPSN`.

The function change is only done in the analysis phase when the instruction argument (stack item index) is decode. During execution the same procedure is used as for legacy `DUP` and `SWAP` instructions.

Closes #84.

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -15,6 +15,18 @@ bool is_terminator(uint8_t c) noexcept
     return c == OP_JUMP || c == OP_JUMPI || c == OP_STOP || c == OP_RETURN || c == OP_REVERT ||
            c == OP_SELFDESTRUCT;
 }
+
+int load_1byte_immediate_value(const uint8_t* code, size_t code_size, size_t& index) noexcept
+{
+    return (index + 1 < code_size) ? code[++index] : 0;
+}
+
+int load_2byte_immediate_value(const uint8_t* code, size_t code_size, size_t& index) noexcept
+{
+    const auto hi_byte = load_1byte_immediate_value(code, code_size, index);
+    const auto lo_byte = load_1byte_immediate_value(code, code_size, index);
+    return (hi_byte << 8) | lo_byte;
+}
 }  // namespace
 
 int code_analysis::find_jumpdest(int offset) const noexcept
@@ -84,8 +96,8 @@ code_analysis analyze(
         auto& instr = jumpdest ? analysis.instrs.back() : analysis.instrs.emplace_back(fns[c]);
 
         const auto metrics = instr_table[c];
-        const auto instr_stack_req = metrics.num_stack_arguments;
-        const auto instr_stack_change = metrics.num_stack_returned_items - instr_stack_req;
+        auto instr_stack_req = int{metrics.num_stack_arguments};
+        auto instr_stack_change = int{metrics.num_stack_returned_items - instr_stack_req};
 
         if (metrics.gas_cost > 0)  // can be -1 for undefined instruction
             block->gas_cost += metrics.gas_cost;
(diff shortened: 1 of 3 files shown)


lib/evmone/analysis.cpp
```diff
@@ -15,6 +15,18 @@ bool is_terminator(uint8_t c) noexcept
     return c == OP_JUMP || c == OP_JUMPI || c == OP_STOP || c == OP_RETURN || c == OP_REVERT ||
            c == OP_SELFDESTRUCT;
 }
+
+int load_1byte_immediate_value(const uint8_t* code, size_t code_size, size_t& index) noexcept
+{
+    return (index + 1 < code_size) ? code[++index] : 0;
```
gumb0: Does this mean that code not having an argument behaves the same as argument 0? And costs the same gas? Is this correct? (how is this with the code not having enough bytes after PUSH?)

lib/evmone/analysis.cpp
```diff
@@ -124,7 +135,41 @@ code_analysis analyze(
             instr.arg.p.number = static_cast<int>(i);
         else if (c >= OP_LOG0 && c <= OP_LOG4)
             instr.arg.p.number = c - OP_LOG0;
-        else if (is_terminator(c))
+
+        if (rev >= EVMC_ISTANBUL)
```
gumb0: I think this should be better `else if`, so that it's not even considered if it's some other opcode.

lib/evmone/analysis.cpp
```diff
@@ -15,6 +15,18 @@ bool is_terminator(uint8_t c) noexcept
     return c == OP_JUMP || c == OP_JUMPI || c == OP_STOP || c == OP_RETURN || c == OP_REVERT ||
            c == OP_SELFDESTRUCT;
 }
+
+int load_1byte_immediate_value(const uint8_t* code, size_t code_size, size_t& index) noexcept
```
gumb0: Minor stylistic suggestion: I would leave updating of `index` on the calling side, i.e. pass it by value here. Then it would be easier to understand what's happending with it, when you look the calling side

lib/evmone/analysis.cpp
chfast: It's the same case as implicit zero bytes after any PUSH instructions.

lib/evmone/analysis.cpp
chfast: Yeah, but we update the index conditionally. We can rename `load_` into `consume_` though.

test/unittests/evm_test.cpp
```diff
@@ -78,6 +78,154 @@ TEST_F(evm, sub_and_swap)
...
+
+    execute(full_stack_code + OP_POP + OP_DUPN + "0100" + ret_top());
+    EXPECT_STATUS(EVMC_SUCCESS);
+    EXPECT_OUTPUT_INT(256 + 2);
+
+    execute(full_stack_code + OP_POP + OP_DUPN + "03fd" + ret_top());
+    EXPECT_STATUS(EVMC_SUCCESS);
+    EXPECT_OUTPUT_INT(1023);
+
+    execute(full_stack_code + OP_DUPN + "03fe" + OP_SWAPSN + "00" + OP_RETURN);
```
gumb0: why does it not use `ret_top()` and how does it return anything without storing in memory?

gumb0:
Unit test suggestions:
- check the case with implicit zero bytes (check at least that it completes sucessfully, maybe check that it's the same gas used as with actual zero args)
- check that it's undefined isntructions before Istanbul
- Two test cases from the EIP spec

test/unittests/evm_test.cpp
chfast: The `ret_top()` requires additional push therefore more stack space. We only check the output size which matches the last duped item.

lib/evmone/analysis.cpp
chfast: Reworked.

lib/evmone/analysis.cpp
```diff
@@ -15,6 +16,20 @@ bool is_terminator(uint8_t c) noexcept
     return c == OP_JUMP || c == OP_JUMPI || c == OP_STOP || c == OP_RETURN || c == OP_REVERT ||
            c == OP_SELFDESTRUCT;
 }
+
+std::pair<int, size_t> load_1byte_immediate_value(
+    const uint8_t* code, size_t code_size, size_t index) noexcept
+{
+    return {(index + 1 < code_size) ? code[++index] : 0, index};
```
gumb0: This doesn't look right, because the order of evaluation of `pair` constructor arguments is not defined, is it?

chfast: I thought this is now specified in C++17, but it is not. To be fixed.

gumb0: (re the need to update `index` conditionally: even if would update it unconditionally on the  calling side and go beyond the code size, it will correctly break out of the loop, so updating here doesn't seem necessary to me)

chfast: I though about this, but sounds risky. It would not work if we used `!=` in the loop condition, or used an iterator instead of an index. Quite fragile to have the correctness of the implementation of this helper depend on some other loop implementation outside.

chfast:
Can I close this?

PR #123, opened 2019-08-13
article: Efficient gas calculation algorithm for EVM

chfast:
TODO:
- [x] Add link to the article in README

### DIFF
--- a/docs/efficient_gas_calculation_algorithm.md
+++ b/docs/efficient_gas_calculation_algorithm.md
@@ -11,13 +11,14 @@ Let's start by defining some basic and universal instructions' parameters.
 1. Base gas cost.
    
    This is the static gas cost of instructions. Some instructions may have 
-   additional cost depending on their operand values - these have to be
-   handled individually during the instruction execution.
+   _additional_ cost depending on their operand values and/or the environment
+   — these have to be handled individually during the instruction execution.
+   The _final_ cost is never less than the _base_ cost.
    
 2. Stack height requirement.
 
    This is the minimum stack height (number of items on the stack) 
-   required for the instruction execution.
+   required prior to the instruction execution.
    
 3. Stack height change.
 
@@ -36,45 +37,49 @@ Examples:
 | CALL    | 700           | 7                        | -6                  |
 
 
-## Basic blocks
+## Basic instruction blocks
 
-A _basic block_ is a sequence of instructions that are executed "straight-line"
-without being interrupted by jumps. I.e. they are nodes in the _control flow graph_.
+A _basic instruction block_ is a sequence of "straight-line" instructions 
+without jumps and jumpdests in the middle.
+Jumpdests are only allowed at the entry, jumps at the exit.
+Basic blocks are nodes in the _control flow graph_.
+See [Basic Block] in Wikipedia.
 
-The name "basic block" has been taken from LLVM. Other names are: just "block"
-in wasm or "subroutine" in subroutine-threaded interpreters.
+In EVM there are simple rules to identify basic instruction block boundaries:
 
-In EVM there are simple rules to identify basic block boundaries:
+1. A basic instruction block _starts_ right before:
+   - the first instruction in the code,
+   - `JUMPDEST` instruction.
 
-1. The following instructions _end_ a basic block:
+2. A basic instruction block _ends_ after the "terminator" instructions:
    - `JUMP`,
    - `JUMPI`,
    - `STOP`,
    - `RETURN`,
    - `REVERT`,
-   - `SELFDESTRUCT`,
-   - an instruction directly preceding `JUMPDEST`.
+   - `SELFDESTRUCT`.
 
-2. The following instructions _start_ a basic block:
-   - `JUMPDEST`,
-   - the first instruction in the code,
-   - an instruction directly after an instruction that has ended previous basic block.
+A basic instruction block is a shortest sequence of instructions such that 
+a basic block starts before the first instruction and ends after the last.
 
+In some cases multiple of the above rules can apply to single basic instruction 
+block boundary.
 
 ## Algorithm
 
-The algorithm for calculating gas and checking stack requirements precomputes
-the values for basic blocks and during execution the checks are done once per block.
+The algorithm for calculating gas and checking stack requirements pre-computes
+the values for basic instruction blocks and during execution the checks 
+are done only once per instruction block.
 
 ### Collecting requirements for basic blocks
 
-For a basic block we need to collect following information:
+For a basic block we need to collect the following information:
 
-- total base gas required by instructions,
-- the stack height required (the minimum stack height needed to execute 
-  all instructions in the block),
-- ~~the start-to-end stack height change~~,
-- the relative maximum stack height.
+- total **base gas cost** of all instructions,
+- the **stack height required** (the minimum stack height needed to execute all 
+  instructions in the block),
+- the **maximum stack height growth** relative to the stack height at block 
+  start.
 
 This is done as follows:
 
@@ -83,26 +88,26 @@ This is done as follows:
 
 ```python
 class Instruction:
-    gas_required = 0
+    base_gas_cost = 0
     stack_required = 0
     stack_change = 0
 
 class BasicBlock:
-    gas_required = 0
+    base_gas_cost = 0
     stack_required = 0
-    stack_change = 0  # FIXME: We don't have too keep it.
-    stack_max = 0
+    stack_max_growth = 0
 
 def collect_basic_block_requirements(basic_block):
+    stack_change = 0
     for instruction in basic_block:
-        basic_block.gas_required += instruction.gas_required
+        basic_block.base_gas_cost += instruction.base_gas_cost
         
-        current_stack_required = instruction.stack_required - basic_block.stack_change
+        current_stack_required = instruction.stack_required - stack_change
         basic_block.stack_required = max(basic_block.stack_required, current_stack_required)
         
-        basic_block.stack_change += instruction.stack_change
+        stack_change += instruction.stack_change
         
-        basic_block.stack_max = max(basic_block.stack_max, basic_block.stack_change)
+        basic_block.stack_max_growth = max(basic_block.stack_max_growth, stack_change)
 ```
 
 ### Checking basic block requirements
@@ -116,14 +121,14 @@ class ExecutionState:
     stack = []
 
 def check_basic_block_requirements(state, basic_block):
-    state.gas_left -= basic_block.gas_required
+    state.gas_left -= basic_block.base_gas_cost
     if state.gas_left < 0:
         raise OutOfGas()
     
     if len(state.stack) < basic_block.stack_required:
         raise StackUnderflow()
     
-    if len(state.stack) + basic_block.stack_max > 1024:
+    if len(state.stack) + basic_block.stack_max_growth > 1024:
         raise StackOverflow()
 ```
 
@@ -131,7 +136,7 @@ def check_basic_block_requirements(state, basic_block):
 
 ### EVM may terminate earlier
 
-Because requirements for a whole basic blocks are checked up front, the instructions
+Because requirements for a whole basic block are checked up front, the instructions
 that have observable external effects might not be executed although they would be
 executed if the gas counting would have been done per instruction.
 This is not a consensus issue because the execution terminates with a "hard" exception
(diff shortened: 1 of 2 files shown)


docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+## Instructions metadata
+
+Let's start by defining some basic and universal instructions' parameters.
+
+1. Base gas cost.
+   
+   This is the static gas cost of instructions. Some instructions may have 
+   additional cost depending on their operand values - these have to be
```
chriseth: What about those whose gas costs also depend on the state?

chfast: The same, you only include the minimal possible cost here. E.g. for `SSTORE` the base cost is 0 (old HFs) or 5000. And later might go to 200 with EIP-2200.

chriseth: Can we assume the base gas cost to never be higher than the final gas cost?

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+   handled individually during the instruction execution.
+   
+2. Stack height requirement.
+
+   This is the minimum stack height (number of items on the stack) 
+   required for the instruction execution.
```
chriseth:
```suggestion
   required prior to the instruction execution.
```

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+| DUP4    | 3             | 4                        | 1                   |
+| SWAP1   | 3             | 2                        | 0                   |
+| ADDRESS | 2             | 0                        | 1                   |
+| CALL    | 700           | 7                        | -6                  |
+
+
+## Basic blocks
+
+A _basic block_ is a sequence of instructions that are executed "straight-line"
+without being interrupted by jumps. I.e. they are nodes in the _control flow graph_.
```
chriseth: What about calls and e.g. stop?

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+
+The name "basic block" has been taken from LLVM. Other names are: just "block"
```
chriseth: Isn't this just the generic term in compiler theory?

docs/efficient_gas_calculation_algorithm.md
chriseth: Ah stop is mentioned below, but it would be interesting to know about call, since a call does not really influence the internal control flow.

docs/efficient_gas_calculation_algorithm.md
holiman: yes

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+in wasm or "subroutine" in subroutine-threaded interpreters.
+
+In EVM there are simple rules to identify basic block boundaries:
+
+1. The following instructions _end_ a basic block:
```
chriseth:
```suggestion
1. A basic block _ends_ after the following instructions:
```

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+   - `JUMP`,
+   - `JUMPI`,
+   - `STOP`,
+   - `RETURN`,
+   - `REVERT`,
+   - `SELFDESTRUCT`,
+   - an instruction directly preceding `JUMPDEST`.
+
+2. The following instructions _start_ a basic block:
```
chriseth:
```suggestion
2. A basic block _starts_ right before the following instructions:
```

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+   - `JUMPDEST`,
+   - the first instruction in the code,
+   - an instruction directly after an instruction that has ended previous basic block.
+
+
```
chriseth:
```suggestion
A basic block is a shortest sequence of instructions such that a basic block starts before the first instruction and ends after the last.
```

chriseth: (hopefully, this is still well-defined...)

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+In EVM there are simple rules to identify basic block boundaries:
+
+1. The following instructions _end_ a basic block:
+   - `JUMP`,
+   - `JUMPI`,
+   - `STOP`,
+   - `RETURN`,
+   - `REVERT`,
+   - `SELFDESTRUCT`,
+   - an instruction directly preceding `JUMPDEST`.
```
chriseth:
```suggestion
   - the last instruction in the program
   - an invalid instruction
```

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+
+2. The following instructions _start_ a basic block:
+   - `JUMPDEST`,
```
chriseth:
```suggestion
```
(this should be redundant together with the two rules below)

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+   - the first instruction in the code,
+   - an instruction directly after an instruction that has ended previous basic block.
+
+
+## Algorithm
+
+The algorithm for calculating gas and checking stack requirements precomputes
+the values for basic blocks and during execution the checks are done once per block.
```
chriseth:
```suggestion
the values for basic blocks and during execution the checks are done only once per block.
```

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Collecting requirements for basic blocks
+
+For a basic block we need to collect following information:
```
chriseth:
```suggestion
For a basic block we need to collect the following information:
```

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+
+- total base gas required by instructions,
+- the stack height required (the minimum stack height needed to execute 
+  all instructions in the block),
+- ~~the start-to-end stack height change~~,
```
chriseth: Why is this crossed out?

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+- the relative maximum stack height.
```
chriseth: What exactly is this?

docs/efficient_gas_calculation_algorithm.md
chfast: When writing the article, I realized we don't have to keep this value for execution. It is only needed temporarily to calculate other values. But I want to apply this change to evmone first to make sure. I believe @zac-williamson proposed this change in #11.

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+    
+    if len(state.stack) + basic_block.stack_max > 1024:
+        raise StackOverflow()
+```
+
+## Misc
+
+### EVM may terminate earlier
+
+Because requirements for a whole basic blocks are checked up front, the instructions
```
chriseth:
```suggestion
Because requirements for a whole basic block are checked up front, the instructions
```

docs/efficient_gas_calculation_algorithm.md
chriseth: Can you add that to the text?

chfast: > Can we assume the base gas cost to never be higher than the final gas cost?

In some way this is the definition of it.

docs/efficient_gas_calculation_algorithm.md
chfast: As mentioned in Misc, for calls we only need to know the "current" gas counter, which is not directly available in EVMs following this algorithm. Because of that, in evmone we add back the cost of all instructions after the call.

docs/efficient_gas_calculation_algorithm.md
chfast: I never had proper education in this field. If the term is well known, I should simplify this paragraph.

docs/efficient_gas_calculation_algorithm.md
chfast: We actually ignore invalid instructions by assuming they cost 0.

docs/efficient_gas_calculation_algorithm.md
chfast: I don't believe this is true. You can have a `JUMPDEST` anywhere.

docs/efficient_gas_calculation_algorithm.md
chfast: The maximum stack height increase in the basic block relatively to the stack height on which the block execution starts. If `stack_max` is 5 and the block execution starts at stack height 10 the absolute stack height will reach 15.

docs/efficient_gas_calculation_algorithm.md
chfast: Change in evmone: https://github.com/ethereum/evmone/pull/125. So this is not needed.

docs/efficient_gas_calculation_algorithm.md
chriseth: I'm just saying that 'additional cost' could be interpreted as 'cost that has to be added' in which case you could also add a negative value. Furthermore, the sentence only explicitly states "depending on their operand values". I think it would be clearer if you also state "on their operand values or the environment", in which case there is no doubt left. Just my two cents :)

docs/efficient_gas_calculation_algorithm.md
chriseth: Not sure about the well-known-ness, but llvm certainly did not introduce it :)

docs/efficient_gas_calculation_algorithm.md
chriseth: sure, but you say: a basic block starts right before ... an instruction directly after an intruction that has ended the previous basic block. And above: a basic block ends after ... an instruction directly preceding "jumpdest". Maybe just nitpicking :)

docs/efficient_gas_calculation_algorithm.md
chriseth: So maybe: "The maximum stack height relative to the stack height at block start"? Also is that per instruction in the block or is it a single value?

chfast: Per block.

gumb0: Maybe "the maximum stack height change over the block"?
I would also call it `stack_change_max` in the pseudocode

docs/efficient_gas_calculation_algorithm.md
```diff
@@ -0,0 +1,156 @@
...
+Because requirements for a whole basic block are checked up front, the instructions
+that have observable external effects might not be executed although they would be
+executed if the gas counting would have been done per instruction.
+This is not a consensus issue because the execution terminates with a "hard" exception
+anyway (and all effects are reverted) but might produce unexpected traces 
+or terminate with a different exception type.
+
+### Current "gas left" value
+
+In EVMJIT also `GAS` and _call_ instructions begin a basic block. This is because
```
gumb0: just a rewording
```suggestion
In EVMJIT additional instructions that begin a basic block are `GAS` and any of the _call_ instructions. This is because
```

docs/efficient_gas_calculation_algorithm.md
chfast: @axic proposed `stack_max_growth`, and I renamed is like that in the code.

docs/efficient_gas_calculation_algorithm.md
holiman: I'd prefer if you called it 'basic segment', and talk about segments. Because we're talking about blockchains here, and in extension maybe doing paralell gas-calculation of future blocks, and things become a bit messy if we have two definitions for 'block'.

docs/efficient_gas_calculation_algorithm.md
holiman: I'm wondering about the attackability of this -- in hte case where a contract consists of 24k `JUMPDEST`s. I guess that's up to the implementation.

chfast: I have a test/benchmark for this one. Other variations about it are definitely needed.

evmone's top speed is ~1Ggas/s. In this case where we do gas/stack checking for every instruction the execution is ~0.5Ggas/s.

The bigger problem is the analysis of this code takes ~0.1 ms what means it should cost ~1000 gas. This is the biggest concern so far.

docs/efficient_gas_calculation_algorithm.md
chfast: This is pretty concrete term: https://en.wikipedia.org/wiki/Basic_block.

Maybe "basic instruction block" then?

docs/efficient_gas_calculation_algorithm.md
chfast: Added both suggestions in the best way I was capable of.

PR #125, opened 2019-08-13
Do not keep information about block stack change

chfast:
This information is only needed in analysis to compute other block stack requirements parameters.

### DIFF
--- a/lib/evmone/analysis.hpp
+++ b/lib/evmone/analysis.hpp
@@ -162,10 +162,14 @@ struct instr_info
 
 struct block_info
 {
+    /// The total base gas cost of all instructions in the block.
     int64_t gas_cost = 0;
-    int stack_req = 0;     ///< The required stack height to execute the block.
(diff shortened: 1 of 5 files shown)


lib/evmone/analysis.hpp
```diff
@@ -162,10 +162,14 @@ struct instr_info
...
     int64_t gas_cost = 0;
-    int stack_req = 0;     ///< The required stack height to execute the block.
-    int stack_max = 0;     ///< The relative maximum stack height in the block.
-    int stack_change = 0;  ///< The relative stack height change after the execution of the block.
+
+    /// The stack height required to execute the block.
+    int stack_req = 0;
+
+    /// The maximum stack height relative to the stack height at block start.
+    int stack_max = 0;
```
axic: Perhaps `stack_max_growth` represents this variable better?

chfast: Better name. Renamed.

PR #113, opened 2019-08-01
Test for EIP-663 with jumpdest

gumb0:
I've added a test on top of [EIP-663 implementation](https://github.com/ethereum/evmone/pull/98) for the problem called out by @chriseth in https://ethereum-magicians.org/t/eip-663-unlimited-swap-and-dup-instructions/3346/10

Basically some JUMPDESTs that were considered valid before EIP-663, with it are considered a part of the multibyte opcode and not a JUMPDEST. So some old contracts might be broken.

PR #136, opened 2019-08-26
Tests demonstrating unlimited SWAP/DUP problem with jumpdests

gumb0:
Tests extracted from https://github.com/ethereum/evmone/pull/113

Without EIP-663 implementation result in Petersburg and Istanbul is the same.

### DIFF
--- a/test/unittests/evm_test.cpp
+++ b/test/unittests/evm_test.cpp
@@ -95,6 +95,40 @@ TEST_F(evm, sub_and_swap)
     EXPECT_EQ(result.output_data[31], 1);
 }
 
+TEST_F(evm, swapsn_jumpdest)
+{
+    // Test demonstrating possible problem with introducing multibyte SWAP/DUP instructions as per
+    // EIP-663 variants B and C.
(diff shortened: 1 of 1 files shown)


test/unittests/evm_test.cpp
```diff
@@ -95,6 +95,40 @@ TEST_F(evm, sub_and_swap)
...
+    // EIP-663 variants B and C.
+    // When SWAPSN is implemented execution will fail with EVMC_BAD_JUMP_DESTINATION.
+    const auto swapsn = "b3";
+    const auto code = push(4) + OP_JUMP + swapsn + OP_JUMPDEST + push(0) + ret_top();
+
+    rev = EVMC_PETERSBURG;
+    execute(code);
+    EXPECT_STATUS(EVMC_SUCCESS);
+
+    rev = EVMC_ISTANBUL;
```
chfast: Can you add one more execution with `EVMC_MAX_REVISION` to detect possible problems in future revisions?

gumb0: @chfast added

PR #140, opened 2019-08-27
EVM unittests grouping

chfast:


chfast:
> Haven't verified that only state tests were moved to `evm_state_tests`, but the idea looks good.

You can check the test count in logs. But I checked, there are still 116 of them.

PR #141, opened 2019-08-28
Fixes and regression tests

chfast:


lib/evmone/instructions.cpp
```diff
@@ -1065,7 +1065,7 @@ const instr_info* op_create(const instr_info* instr, execution_state& state) noe
     state.return_data.clear();
 
     if (state.msg->depth >= 1024)
-        return nullptr;
+        return ++instr;
```
gumb0: What does this change do?

gumb0: Oh so it stopped on invalid `CREATE` going over max stack depth, I see

PR #142, opened 2019-08-29
Implement EIP-2200

gumb0:
https://github.com/ethereum/EIPs/pull/2200

### DIFF
--- a/lib/evmone/instructions.cpp
+++ b/lib/evmone/instructions.cpp
@@ -499,6 +499,14 @@ const instruction* op_sstore(const instruction* instr, execution_state& state) n
     if (state.msg->flags & EVMC_STATIC)
         return state.exit(EVMC_STATIC_MODE_VIOLATION);
 
+    if (state.rev >= EVMC_ISTANBUL)
+    {
+        const auto correction = state.current_block_cost - instr->arg.number;
+        const auto gas_left = state.gas_left + correction;
+        if (gas_left <= 2300)
+            return state.exit(EVMC_OUT_OF_GAS);
+    }
+
     const auto key = intx::be::store<evmc::bytes32>(state.stack.pop());
     const auto value = intx::be::store<evmc::bytes32>(state.stack.pop());
     auto status = state.host.set_storage(state.msg->destination, key, value);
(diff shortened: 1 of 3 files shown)


test/unittests/evm_state_test.cpp
```diff
@@ -95,42 +95,72 @@ TEST_F(evm_state, sstore_cost)
...
+            EXPECT_EQ(gas_used, 20812);
+        else
+            EXPECT_EQ(gas_used, 25012);
 
         // Modified again:
         storage.clear();
         storage[v1] = {v1, true};
         execute(sstore(1, push(2)));
         EXPECT_EQ(result.status_code, EVMC_SUCCESS);
-        EXPECT_EQ(gas_used, rev == EVMC_CONSTANTINOPLE ? 206 : 5006);
```
axic: Shouldn't all these have `>= CONSTANTINOPLE`?

E.g. `(rev >= istanbul) elseif (rev == constantinople) else ..` (because it was disabled in petersburg)

gumb0: right, changed now to cover post-Istanbul revisions

gumb0:
"OOG if gas below stipend" is not implemented yet

lib/evmone/instructions.cpp
```diff
@@ -491,20 +491,38 @@ const instr_info* op_sstore(const instr_info* instr, execution_state& state) noe
     if (state.msg->flags & EVMC_STATIC)
         return state.exit(EVMC_STATIC_MODE_VIOLATION);
 
+    if (state.rev >= EVMC_ISTANBUL)
+    {
+        const auto correction = state.current_block_cost - instr->arg.p.number;
+        const auto gas_left = static_cast<uint64_t>(state.gas_left + correction);
```
gumb0: This doesn't seem to work right: for `PUSH1 0 PUSH1 0 SSTORE` with 2306 starting gas, at the beginning of SSTORE `state.gas_left` is 2300 and `correction` is 6

gumb0: got it, `arg.p.number` should be filled for SSTORE in analysis

PR #143, opened 2019-08-29
bytecode: Add helper for constructing call opcodes

chfast:


test/unittests/evm_calls_test.cpp
```diff
@@ -342,7 +341,11 @@ TEST_F(evm_calls, call_then_oog)
     accounts[call_dst] = {};
     call_result.status_code = EVMC_FAILURE;
     call_result.gas_left = 0;
-    execute(1000, "6040600060406000600060aa60fef180018001800150");
+
+    const auto code =
+        ::call(0xaa).gas(254).value(0).input(0, 0x40).output(0, 0x40) + 4 * add(OP_DUP1) + OP_POP;
```
chfast: Here `call()` clashes with `.call()` from `MockedHost`. This can be solved by compositing (instead of inheriting) of `MockedHost`. This should also improve general readability by having `host.accounts[x]`, `host.call_result`, etc.

test/utils/bytecode.hpp
```diff
@@ -181,6 +181,81 @@ inline bytecode sload(bytecode index)
...
+    bytecode m_value = 0;
+    bytecode m_input = 0;
+    bytecode m_input_size = 0;
+    bytecode m_output = 0;
+    bytecode m_output_size = 0;
+
+public:
+    explicit call_instruction(bytecode address) : m_address{std::move(address)} {}
+
+    auto& gas(bytecode g)
```
gumb0: Seems it would be more logical and better constraining the caller, if you had `uint64_t` parameter here (and in other methods), and `m_gas = bytecode{g};` in the body.

chfast: I don't want to complicate this. This is rather a tool for presenting code.

chfast: Also you can do `.gas(calldataload(0))`.

test/unittests/evm_calls_test.cpp
halfalicious: This new syntax looks great!

PR #144, opened 2019-08-29
Block info optimization

chfast:
Requires #153.

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -157,9 +157,8 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size) 
                                        static_cast<int16_t>(block.stack_req) :
                                        std::numeric_limits<int16_t>::max();
             const auto stack_max_growth = static_cast<int16_t>(block.stack_max_growth);
-            analysis.blocks.emplace_back(block_info{block.gas_cost, stack_req, stack_max_growth});
-            analysis.instrs[block.begin_block_index].arg.number =
-                static_cast<int>(analysis.blocks.size() - 1);
+            analysis.instrs[block.begin_block_index].arg.block = {
+                block.gas_cost, stack_req, stack_max_growth};
 
 
             // Create new block.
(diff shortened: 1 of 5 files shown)


lib/evmone/analysis.cpp
```diff
@@ -11,6 +11,19 @@ namespace evmone
...
 {
+struct block_analysis
+{
+    int gas_cost = 0;
+
+    int stack_req = 0;
+    int stack_max_growth = 0;
+    int stack_change = 0;
+
+    size_t first_instruction_index = 0;
```
gumb0: Maybe it's `begin_block_index`?
As I understand it can be either `OPX_BEGINBLOCK` or `OP_JUMPDEST`, but "first instructions" sounds a bit like a first real instruction in the block

chfast: This is effectively always `OPX_BEGINBLOCK` as we skip `OP_JUMPDEST` from instruction table. Anyway, I'll rename it and document more.

chfast: Renamed.

test/unittests/analysis_test.cpp
```diff
@@ -43,16 +41,15 @@ TEST(analysis, stack_up_and_down)
...
     EXPECT_EQ(analysis.instrs[2].fn, op_table[OP_DUP1]);
     EXPECT_EQ(analysis.instrs[8].fn, op_table[OP_POP]);
     EXPECT_EQ(analysis.instrs[18].fn, op_table[OP_PUSH1]);
 
-    ASSERT_EQ(analysis.blocks.size(), 1);
-    EXPECT_EQ(analysis.blocks[0].gas_cost, 7 * 3 + 10 * 2 + 3);
-    EXPECT_EQ(analysis.blocks[0].stack_req, 3);
-    EXPECT_EQ(analysis.blocks[0].stack_max_growth, 7);
+
+    EXPECT_EQ(analysis.instrs[0].arg.block.gas_cost, 7 * 3 + 10 * 2 + 3);
```
gumb0: I'd create an alias like `auto& block = analysis.instrs[0].arg.block;` for less repetition

chfast: Done.

PR #149, opened 2019-09-03
Add "loop full of jumpdests" unit test

chfast:


test/unittests/evm_other_test.cpp
```diff
@@ -30,3 +31,19 @@ TEST_F(evm_other, evmone_block_stack_req_overflow)
...
     EXPECT_STATUS(EVMC_STACK_UNDERFLOW);
 }
+
+TEST_F(evm_other, loop_full_of_jumpdests)
+{
+    // The code is a simple loop with a counter taken from the input or a constant (325) if the
+    // input is zero. The loop body contains of only JUMPDESTs, as much as the code size limit
+    // allows.
+
+    const auto code = push(15) + not_(0) + mul(325, iszero(dup1(calldataload(0)))) + OP_OR +
```
gumb0: I'm lost about why `dup1` is needed here.
It duplicates calldata on the stack, then top calldata is converted into either 1 or 0 with `ISZERO`, then it gets to MUL, so multiplies this 1 or 0 by another calldata, is this what's going on?

chfast: Explained in comments.

gumb0: Aaah, it took me some time to realize that the source of my confusion was that `mul(a, b)` pushes `b` first and then `a`, not the other way around... kind of makes sense now

PR #150, opened 2019-09-03
Micro benchmarks support

chfast:


test/bench/bench.cpp
```diff
@@ -110,9 +108,9 @@ struct benchmark_case
 };
 
 
-void load_benchmark(fs::path path)
+void load_benchmark(fs::path path, const fs::path& subdir)
```
gumb0: why pass one `fs::path` by value and another by const ref?

test/bench/bench.cpp
```diff
@@ -110,9 +108,9 @@ struct benchmark_case
...
 {
-    auto base_name = path.stem().string();
+    const auto base_name = (subdir / path.stem()).string();
```
gumb0: Why `subdir` parameter is now needed? can't full path be passed in `path` only?

test/bench/bench.cpp
chfast: Because the one passed by value is modified inside the function. I will change it though because it also negatively affects the readability in the place where it is modified.

test/bench/bench.cpp
chfast: I want the tests to be named `subdir/filename/...`. I could also pass the root dir but this way seems easier.

gumb0: I think I got it, you want the test names not to have the full path, but relative one. But it's confusing. Maybe this parameter shouldn't be `fs::path` at all (but string) if it's used only for test name construction? (could still be `fs::path` in `load_benchmarks_from_dir`)
Maybe I'd call it `test_name_prefix` and move `(subdir / path.stem()).string()` part to the calling side.

test/bench/bench.cpp
```diff
@@ -175,15 +174,27 @@ void load_benchmark(fs::path path)
     }
 }
 
-void load_benchmarks_from_dir(const char* path)
+void load_benchmarks_from_dir(const fs::path& path, const fs::path& subdir = {})
```
gumb0: Maybe it's `(const fs::path& absolute_path, const fs::path& relative_path = {})`

test/bench/bench.cpp
```diff
@@ -175,15 +174,27 @@ void load_benchmark(fs::path path)
...
+        if (e.is_directory())
+            subdirs.emplace_back(e);
+        else if (e.path().extension() != inputs_extension)
+            files.emplace_back(e);
     }
+
+    std::sort(std::begin(subdirs), std::end(subdirs));
+    std::sort(std::begin(files), std::end(files));
+
+    for (auto& f : files)
```
gumb0: can be `auto const&`

test/bench/bench.cpp
chfast: Make sense.

chfast: Done.

test/bench/bench.cpp
chfast: It is actually not. The first one can be also an relative path. Refactored differently as suggested in other comment.

PR #153, opened 2019-09-09
Analysis refactoring

chfast:
Requires https://github.com/ethereum/evmone/pull/166.

Refactors analysis to have new block creation in single place. It also removes some unneeded variables and code.

TODO
- [x] Squash commits.

Bechmarks:
```
blake2b_huff/analysis                           -0.0866         -0.0866            37            34            37            34
blake2b_shifts/analysis                         -0.0941         -0.0940            20            18            20            18
sha1_divs/analysis                              -0.0930         -0.0929             4             4             4             4
sha1_shifts/analysis                            -0.0826         -0.0826             4             3             4             3
weierstrudel/analysis                           -0.0622         -0.0622            45            42            45            42
micro/loop_with_many_jumpdests/analysis         -0.0163         -0.0163           329           324           329           324
```

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -18,6 +18,8 @@ inline constexpr uint64_t load64be(const unsigned char* data) noexcept
 code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size) noexcept
 {
     const auto& fns = get_op_table(rev);
+    const auto opx_beginblock_fn = fns[OPX_BEGINBLOCK];
+
     code_analysis analysis;
 
     const auto max_instrs_size = code_size + 1;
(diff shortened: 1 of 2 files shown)


lib/evmone/analysis.cpp
```diff
@@ -54,38 +54,20 @@ code_analysis analyze(
     block_info* block = nullptr;
 
     int block_stack_change = 0;
-    int instr_index = 0;
+
+    // Create new block.
+    block = &analysis.blocks.emplace_back();
+    block_stack_change = 0;
+    analysis.instrs.emplace_back(fns[OPX_BEGINBLOCK]).arg.p.number =
```
gumb0: This was clearer when it was on two separate lines.

lib/evmone/analysis.cpp
```diff
@@ -177,31 +170,21 @@ code_analysis analyze(
...
-
-        case OP_JUMP:
-        case OP_JUMPI:
-        case OP_STOP:
-        case OP_RETURN:
-        case OP_REVERT:
-        case OP_SELFDESTRUCT:
-            block = nullptr;
-            break;
+        if (create_new_block || (code_pos != code_end && *code_pos == OP_JUMPDEST))
```
gumb0: Maybe you could set `create_new_block` to `true` in `if (opcode == OP_JUMPDEST)` above, then this would be simpler

lib/evmone/analysis.cpp
```diff
@@ -147,18 +152,6 @@ code_analysis analyze(
             break;
         }
 
-        case ANY_DUP:
-            // TODO: This is not needed, but we keep it
```
gumb0: Why is this not needed anymore?

chfast: The information provided here is not needed because now we have individual functions for each DUP instruction. This is slightly faster. Think of it as this information is encoded in the function pointer. This will also save us space as we only need single function pointer per DUP instruction instead of function pointer + an immediate value.

lib/evmone/analysis.cpp
chfast: That's right, but if `create_new_block` is already `true` we don't have to perform other checks.

lib/evmone/analysis.cpp
chfast: Done.

lib/evmone/analysis.cpp
chfast: Renamed that to `is_terminator`. This is consistent with future changes in #158.

chfast:
Added benchmark results.

axic:
Would it make sense merging all the test changes separately or they are different due to refactoring?

chfast:
> Would it make sense merging all the test changes separately or they are different due to refactoring?

Will do.

chfast:
#166.

chfast:
How does it look now?

lib/evmone/analysis.cpp
```diff
@@ -150,31 +143,22 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
...
-        case OP_SELFDESTRUCT:
-            block = nullptr;
-            break;
+        if (is_terminator || (code_pos != code_end && *code_pos == OP_JUMPDEST))
+        {
+            // Create new basic block if
+            // this is a terminating instruction or the next instruction is a JUMPDEST.
+            block = &analysis.blocks.emplace_back();
+            block_stack_change = 0;
+            analysis.instrs.emplace_back(fns[OPX_BEGINBLOCK]).arg.number =
```
gumb0: here maybe split into two lines, too

PR #154, opened 2019-09-09
Put instruction immediate values in the instruction table

chfast:
Requires #144, #153.

### DIFF
--- a/test/utils/dump.cpp
+++ b/test/utils/dump.cpp
@@ -28,7 +28,7 @@ void dump(const evmone::code_analysis& analysis)
 
         if (c == OPX_BEGINBLOCK)
         {
-            block = &instr.arg.block;
+            block = &instr.block;
 
             const auto get_jumpdest_offset = [&analysis](size_t index) noexcept
             {
@@ -59,7 +59,7 @@ void dump(const evmone::code_analysis& analysis)
                   << gas_cost;
 
         if (c >= OP_PUSH1 && c <= OP_PUSH8)
-            std::cout << '\t' << instr.arg.small_push_value;
+            std::cout << '\t' << instr.small_push_value;
 
         std::cout << '\n';
     }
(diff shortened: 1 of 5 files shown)


axic:
Needs a rebase?

gumb0:
Needs rebase

test/utils/dump.cpp
```diff
@@ -28,7 +28,7 @@ void dump(const evmone::code_analysis& analysis)
 
         if (c == OPX_BEGINBLOCK)
         {
-            block = &instr.arg.block;
+            block = &instr.block;
```
gumb0: I didn't get how this works without changing the way you iterate over `analysis.instrs`
(`instr` points here to the union containing `fn`, not `block_info`, right?)

gumb0: Also I'm not sure what `c` is here actually. Least significant byte of function pointer?

gumb0 (commented):
Looks fine overall, just more documentation would be great, and I'm not sure if `dump` is correct.
Also why bother with these complications, if it doesn't give performance gains?

chfast:
> Also why bother with these complications, if it doesn't give performance gains?

You have to do the work to be able to benchmark it later. This way is more memory efficient - we allocate space for instruction arguments (immediate values) only when needed. This may be important factor when we decide to cache the evmone loaded programs (not to repeat the analysis for the same contracts).

I still don't see big improvements, ~1-2% only. I believe my CPU is good enough in fetching memory quickly enough - old version has the same memory layout, just wastes some space.

Old:
```
 Performance counter stats for 'bin/evmone-bench-master ../../test/benchmarks':

        21 502 116      cache-references                                            
         1 119 073      cache-misses              #    5,204 % of all cache refs    
   127 269 289 930      cycles                                                      
   352 107 262 713      instructions              #    2,77  insn per cycle         

      28,971433332 seconds time elapsed
```

New:
```
 Performance counter stats for 'bin/evmone-bench ../../test/benchmarks':

        17 453 822      cache-references                                            
         1 046 808      cache-misses              #    5,998 % of all cache refs    
   127 385 133 793      cycles                                                      
   358 617 530 718      instructions              #    2,82  insn per cycle         

      28,987604886 seconds time elapsed
```

The above shows that the new version has similar number of cache misses, it just uses less memory in general.

chfast:
I'm leaving this for next release.

kmk142789 (approved):

PR #159, opened 2019-09-10
Drop the call kind runtime argument

chfast:


axic:
Is it such a big speed bump its worth the code duplication (in the binary)?

It doesn't matter for compiling to C, but it could if we actually manage compiling it to Wasm. However it is minor I suppose.

axic:
Couldn't you merge delegatecall/staticcall into call now with `if constexpr` ?

chfast:
This is only done for removing this additional immediate value. I don't have any benchmarks for calls. There is a way to de-duplicate the code again (and for all other calls) by using a proxy function. But I want to keep the information about the call kind effectively in the function pointer.

```
template <kind>
void op_call(...)
{
    op_call_impl(kind, ...);
}
```

chfast:
> Couldn't you merge delegatecall/staticcall into call now with `if constexpr` ?

I could if I had more time. https://github.com/ethereum/evmone/issues/160

chfast:
I'm merging this leaving the mentioned issues for later discussion and future work.

PR #164, opened 2019-09-10
CHAINID opcode - 64 bit version

gumb0:
https://eips.ethereum.org/EIPS/eip-1344

Currently based on EVMC branch changing ChainID size to 64 bit https://github.com/ethereum/evmc/pull/420


gumb0:
Rebased.

Last commit adds Istanbul's opcode repricing to op_table, it should be better extracted to a separate PR.

chfast:
Closing, because the 256-bit variant is only needed at the moment.

PR #167, opened 2019-09-10
Drop op_table parameter in analysis()

chfast:
I believe the `instruction_table` is more precise name, but `op_table` is shorter and might be less mistaken with the table of instructions in the loaded program.

### DIFF
--- a/lib/evmone/analysis.hpp
+++ b/lib/evmone/analysis.hpp
@@ -212,6 +212,8 @@ inline int find_jumpdest(const code_analysis& analysis, int offset) noexcept
 }
 
 EVMC_EXPORT code_analysis analyze(
-    const exec_fn_table& fns, evmc_revision rev, const uint8_t* code, size_t code_size) noexcept;
+    evmc_revision rev, const uint8_t* code, size_t code_size) noexcept;
+
+EVMC_EXPORT const exec_fn_table& get_op_table(evmc_revision rev) noexcept;
 
 }  // namespace evmone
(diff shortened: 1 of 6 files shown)


test/unittests/analysis_test.cpp
```diff
@@ -11,33 +11,24 @@
...
-const auto fake_fn_table = []() noexcept
-{
-    evmone::exec_fn_table fns;
-    for (size_t i = 0; i < fns.size(); ++i)
-        fns[i] = (evmone::exec_fn)i;
-    return fns;
-}
-();
-
+const auto& op_table = get_op_table(rev);
```
gumb0: Danger of static init fiasco? in case `op_tables` in instructions.cpp is initialized after this

chfast: The problem do not apply to `constexpr` globals, but still I can move it inside the function.

lib/evmone/analysis.hpp
```diff
@@ -212,6 +212,8 @@ inline int find_jumpdest(const code_analysis& analysis, int offset) noexcept
 }
 
 EVMC_EXPORT code_analysis analyze(
-    const exec_fn_table& fns, evmc_revision rev, const uint8_t* code, size_t code_size) noexcept;
+    evmc_revision rev, const uint8_t* code, size_t code_size) noexcept;
+
+EVMC_EXPORT const exec_fn_table& get_op_table(evmc_revision rev) noexcept;
```
gumb0: Why do you make it exported?

chfast: Used in analysis tests (evmone is usually built as shared library).

test/unittests/analysis_test.cpp
gumb0: Ah right, haven't noticed `constexpr`

chfast: Good notice anyway.

PR #168, opened 2019-09-10
New metrics table

chfast:
Closes #151 

This adds modified instruction metrics to op_table (previously containing only function pointers to instruction implementations). This should improve cache locality, but no performance improvement noticed.


gumb0:
> This adds modified instruction metrics to op_table

How are they modified?

lib/evmone/analysis.cpp
```diff
@@ -45,24 +43,20 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
     while (code_pos != code_end)
     {
         const auto opcode = *code_pos++;
-
-        const auto metrics = instr_table[opcode];
-        const auto instr_stack_req = metrics.num_stack_arguments;
-        const auto instr_stack_change = metrics.num_stack_returned_items - instr_stack_req;
+        const auto opcode_info = op_tbl[opcode];
```
gumb0: Why not reference?

chfast: > > This adds modified instruction metrics to op_table
> 
> How are they modified?

It contains "stack height change" instead of "number of returned items".
This allows using the value directly in https://github.com/ethereum/evmone/pull/168/files#diff-79f19d4cf3a33e26add73cd9c0ad79cbR50 instead of computing it.

chfast: > Why not reference?

I'd prefer the values are loaded in single go, but when checking assembly - it does not matter: values are loaded individually anyway.

chfast: Actually, using reference is ~27% faster in one case.
```
Comparing bin/evmone-bench to bin/evmone-bench-ref
blake2b_huff/analysis                           -0.0002         -0.0002            40            40            40            40
blake2b_shifts/analysis                         -0.2770         -0.2770            26            19            26            19
sha1_divs/analysis                              -0.0052         -0.0052             4             4             4             4
sha1_shifts/analysis                            -0.0104         -0.0104             4             4             4             4
weierstrudel/analysis                           -0.0290         -0.0290            48            46            48            46
micro/loop_with_many_jumpdests/analysis         -0.0359         -0.0360           149           143           149           143
```

PR #173, opened 2019-09-12
Add classification of EVM instructions (op_kind)

chfast:
Requires #168.

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -77,19 +77,12 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size) 
 
         auto& instr = analysis.instrs.back();
 
-        bool is_terminator = false;  // A flag whenever this is a block terminating instruction.
-        switch (opcode)
+        switch (opcode_info.kind)
         {
-        case OP_JUMP:
-        case OP_JUMPI:
-        case OP_STOP:
-        case OP_RETURN:
-        case OP_REVERT:
-        case OP_SELFDESTRUCT:
-            is_terminator = true;
+        default:
             break;
 
-        case ANY_SMALL_PUSH:
+        case op_kind::small_push:
         {
             const auto push_size = size_t(opcode - OP_PUSH1 + 1);
             const auto push_end = code_pos + push_size;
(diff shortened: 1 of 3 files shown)


chfast:
This looks nice, but in the end makes the analysis loop much slower. I would need more hours of staring on the `perf` output to figure out why. By the way, I discovered an optimization in form of #178.

cc @gcolvin

lib/evmone/analysis.hpp
```diff
@@ -170,13 +170,26 @@ enum intrinsic_opcodes
     OPX_BEGINBLOCK = OP_JUMPDEST
 };
 
+/// Defines classification of EVM instructions.
+enum class op_kind
+{
+    regular,             ///< Totally uninteresting instruction.
```
gumb0: :laughing:

lib/evmone/analysis.hpp
```diff
@@ -170,13 +170,26 @@ enum intrinsic_opcodes
...
+
 struct op_table_entry
 {
     exec_fn fn;
     int16_t gas_cost;
     int8_t stack_req;
     int8_t stack_change;
+    op_kind kind;
 };
+static_assert(sizeof(op_table_entry) <= 2 * sizeof(void*));
```
gumb0: if this is important, maybe `op_kind`'s underlying type should be specified?

lib/evmone/analysis.cpp
```diff
@@ -77,19 +77,12 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
...
-        case OP_JUMPI:
-        case OP_STOP:
-        case OP_RETURN:
-        case OP_REVERT:
-        case OP_SELFDESTRUCT:
-            is_terminator = true;
+        default:
             break;
 
-        case ANY_SMALL_PUSH:
```
gumb0: Can `ANY_SMALL_PUSH` and `ANY_LARGE_PUSH` definitions be deleted now?

lib/evmone/analysis.cpp
```diff
@@ -77,19 +77,12 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
...
+        switch (opcode_info.kind)
         {
-        case OP_JUMP:
-        case OP_JUMPI:
-        case OP_STOP:
-        case OP_RETURN:
-        case OP_REVERT:
-        case OP_SELFDESTRUCT:
-            is_terminator = true;
+        default:
```
gumb0: I'd say explicit cases here would more clearly show the intention
```suggestion
        op_kind::regular:
        op_kind::terminator:
```

axic:
Is this PR still useful/relevant?

PR #178, opened 2019-09-16
Make small push value parsing endianness independent

chfast:
This also makes it faster.

Skylake Mobile:
```
blake2b_huff/analysis                           -0.1002         -0.1002            28            25            28            25
blake2b_shifts/analysis                         -0.2091         -0.2090            15            12            15            12
sha1_divs/analysis                              -0.0715         -0.0714             3             3             3             3
sha1_shifts/analysis                            -0.1620         -0.1620             3             2             3             2
weierstrudel/analysis                           -0.1195         -0.1195            34            30            34            30
micro/loop_with_many_jumpdests/analysis         +0.0189         +0.0190           113           115           113           115
```

Haswell:
```
blake2b_huff/analysis                           -0.1524         -0.1523            30            25            30            25
blake2b_shifts/analysis                         -0.2739         -0.2739            16            12            16            12
sha1_divs/analysis                              -0.1160         -0.1160             3             3             3             3
sha1_shifts/analysis                            -0.1585         -0.1586             3             2             3             2
weierstrudel/analysis                           -0.1736         -0.1735            35            29            35            29
micro/loop_with_many_jumpdests/analysis         -0.0195         -0.0195           110           108           110           108
```

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -23,13 +23,6 @@ struct block_analysis
     explicit block_analysis(size_t index) noexcept : begin_block_index{index} {}
 };
 
-inline constexpr uint64_t load64be(const unsigned char* data) noexcept
-{
-    return uint64_t{data[7]} | (uint64_t{data[6]} << 8) | (uint64_t{data[5]} << 16) |
-           (uint64_t{data[4]} << 24) | (uint64_t{data[3]} << 32) | (uint64_t{data[2]} << 40) |
-           (uint64_t{data[1]} << 48) | (uint64_t{data[0]} << 56);
-}
-
 code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size) noexcept
 {
     static constexpr auto stack_req_max = std::numeric_limits<int16_t>::max();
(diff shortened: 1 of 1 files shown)


lib/evmone/analysis.cpp
```diff
@@ -91,22 +84,23 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
...
-            auto insert_pos = &value_bytes[sizeof(value_bytes) - push_size];
-
-            // TODO: Consier the same endianness-specific loop as in ANY_LARGE_PUSH case.
+            uint64_t value = 0;
+            auto insert_bit_pos = (push_size - 1) * 8;
             while (code_pos < push_end && code_pos < code_end)
-                *insert_pos++ = *code_pos++;
-            instr.arg.small_push_value = load64be(value_bytes);
+            {
+                value |= uint64_t{*code_pos++} << insert_bit_pos;
```
gumb0: Could it still be in a separate helper function maybe?

chfast: The `load64be` was only doing `bswap`. I would need to extract whole `case:`.

chfast: I pushed an example how it could look like, but it is up to 30% slower.

gumb0: Wow then let's leave it, if it's so much slower (why, can't be inlined sometimes?)

`push_end` change looks fine, though

chfast: I think returning multiple values and using tuple and `std::tie` messes it up. But I'm only guessing. I'd need much more time to investigate.

PR #179, opened 2019-09-17
Coverage build continues

chfast:


CMakeLists.txt
```diff
@@ -55,14 +57,15 @@ if(EVMONE_FUZZING)
         message(FATAL_ERROR "Clang 8+ compiler is required for fuzzing")
     endif()
 
-    option(EVMONE_FUZZING_COVERAGE "Fuzzing coverage" OFF)
-
     set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build evmone as a shared library" FORCE)
     set(EVMONE_TESTING ON CACHE BOOL "Build tests and test tools" FORCE)
 
-    if(EVMONE_FUZZING_COVERAGE)
-        add_compile_options(-fprofile-instr-generate -fcoverage-mapping)
```
gumb0: These are never added now? why?

chfast: These are added by `-DCMAKE_BUILD_TYPE=Coverage`.

gumb0: Is it supported directly by cmake?

chfast: Cmake supports custom build types. Here the "Coverage" build type is defined (#176) and it adds the mentioned flags.

gumb0: Ah I think I see, cmake gets flags for the custom build type from the variable containing that build type name, here `CMAKE_CXX_FLAGS_COVERAGE`

gumb0:
```
cable_set_build_type(DEFAULT Release CONFIGURATION_TYPES Release Debug)
```
Should this not list `Coverage`?

chfast:
> Should this not list `Coverage`?

No. The `CONFIGURATION_TYPE` is for MSVC, but there is no support for MSVC in the "Coverage" build type (it would just build without any additional flags).

PR #184, opened 2019-09-18
fuzzer: Correct the LICENSE

chfast:
Change the license of evmone-fuzzer from Apache 2.0 to GPL 3.0.
This is required because evmone-fuzzer "links to" aleth-interpreter (GPL 3.0).

### DIFF
--- a/test/fuzzer/CMakeLists.txt
+++ b/test/fuzzer/CMakeLists.txt
@@ -1,4 +1,4 @@
-# evmone: Fast Ethereum Virtual Machine implementation
+# evmone-fuzzer: LibFuzzer based testing tool for EVMC-compatible EVM implementations.
 # Copyright 2019 The evmone Authors.
 # Licensed under the Apache License, Version 2.0.
 include(ExternalProject)
--- a/test/fuzzer/README.md
+++ b/test/fuzzer/README.md
@@ -1,0 +1,21 @@
+# evmone-fuzzer
+
+> [LibFuzzer] powered testing tool for [EVMC]-compatible EVM implementations.
+
+## License
+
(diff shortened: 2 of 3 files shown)


axic:
> This is required because evmone-fuzzer "links to" aleth-interpreter (GPL 3.0).

It is not required, Apache 2.0 is compatible with GPL 3.0. This source still can be Apache licensed, but the resulting linked binary will be GPL 3.0.

chfast:
> It is not required, Apache 2.0 is compatible with GPL 3.0. This source still can be Apache licensed, but the resulting linked binary will be GPL 3.0.

This seems right. Anyway, you suggest to remove changes in sources, but leave README and LICENSE?

axic:
> Anyway, you suggest to remove changes in sources, but leave README and LICENSE?

To simplify licensing in this repository I would suggest to keep everything under the same license (Apache 2.0) and mention in the fuzzer README that since aleth-interpreter is statically linked, the resulting fuzzer binary falls under the terms of GPL-3.0.

chfast:
It does not make sense yet. The evmone-fuzzer being a aleth-interepreter derived work, its source code must be released under GPL (Apache is compatible with GPL only in the other direction).

axic:
> The evmone-fuzzer being a aleth-interepreter derived work

Well it is a debatable case, hence why LGPL was created. The best would just linking dynamically through EVMC to both aleth-interpreter and evmone.

chfast:
Better now?

axic:
I think the argument here is that you link against EVMC which is Apache-2.0 and you are a "derived work" (but don't think Apache defines that as strictly as GPL) of that. Aleth-Interpreter also just implements that.

chfast:
I don't care much about it. I just didn't want to totally ignore the issue.

PR #186, opened 2019-09-18
Unittests

chfast:
This completes the code coverage by unit tests: https://4685-160548111-gh.circle-artifacts.com/0/coverage/missing.html.

### DIFF
--- a/test/unittests/evm_calls_test.cpp
+++ b/test/unittests/evm_calls_test.cpp
@@ -434,6 +434,22 @@ TEST_F(evm_calls, call_then_oog)
     EXPECT_EQ(result.status_code, EVMC_OUT_OF_GAS);
 }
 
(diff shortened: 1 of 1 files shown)


test/unittests/evm_calls_test.cpp
```diff
@@ -484,6 +513,7 @@ TEST_F(evm_calls, call_with_value_low_gas)
 
 TEST_F(evm_calls, call_oog_after_balance_check)
 {
+    host.accounts[{}] = {};
```
gumb0: What is this, account with address 0?

gumb0: Ah ok, I see the explanation in the commit message.

Maybe `0x0000000000000000000000000000000000000000_address` would be more cleare

PR #189, opened 2019-09-24
EVMC 7

gumb0:


chfast:
I think this is obsolete now after #191.

PR #190, opened 2019-09-24
CHAINID opcode - 256 bit version

gumb0:
This is a variant of CHAINID implementation that uses current EVMC `master`, which defines chain ID as `evmc_uint256be`

Based on separate PR that updates EVMC https://github.com/ethereum/evmone/pull/189

A version that uses 64 bit chain ID was in https://github.com/ethereum/evmone/pull/164


chfast:
Do you want me to finish this one?

gumb0:
@chfast sure go ahead

PR #191, opened 2019-09-27
Upgrade EVMC to version pre 7.0.0

chfast:


lib/evmone/instructions.cpp
```diff
@@ -1377,6 +1377,11 @@ constexpr op_table create_op_table_constantinople() noexcept
 constexpr op_table create_op_table_istanbul() noexcept
 {
     auto table = create_op_table_constantinople();
+    table[OP_BALANCE] = {op_balance, 700, 1, 0};
+    table[OP_CHAINID] = {op_undefined, 2, 0, 1};
+    table[OP_EXTCODEHASH] = {op_extcodehash, 700, 1, 0};
+    table[OP_SELFBALANCE] = {op_undefined, 5, 0, 1};
```
axic: Shouldn't this and chainid be implemented?

gumb0: They are in separate PRs, it's better to merge this first, and then rebase those

Issue #192, opened 2019-09-27
Add 32-bit support


chfast:
- [x] Add 32-bit build to CircleCI. (Example how to do it in other project: https://github.com/ethereum/evmc/blob/master/circle.yml#L149-L158)
- [x] Fix build issues / compiler warnings.
- [x] Test and fix the 32-bit binaries.

hrishikeshSuresh:
I would like to work on this issue, if it hasn't been already assigned/done.

chfast:
> I would like to work on this issue, if it hasn't been already assigned/done.

It's yours.

chfast:
Thank you!

PR #197, opened 2019-10-07
Add 32-bit gcc support

hrishikeshSuresh:
Related to #192 
I added 32-bit build to circle.yml as given in [https://github.com/ethereum/evmc/blob/master/circle.yml#L149-L158](https://github.com/ethereum/evmc/blob/master/circle.yml#L149-L158), with some differences. I built it using `cmake .. -DEVMONE_TESTING=ON` command, followed by `cmake --build . -- -j` and then ran `bin/evmone-unittests` to test it.
No error/warnings were seen during the execution of these three commands. Have I tested it correctly?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -159,6 +159,18 @@ jobs:
           command: |
             codecov --required --gcov-exec "$GCOV" --gcov-root ~/build
 
+  gcc-32bit:
+    executor: linux
+    environment:
+      CMAKE_OPTIONS: -DTOOLCHAIN=cxx17-32bit
+    steps:
+      - run:
+          name: "Install GCC 32-bit"
+          command: |
+            sudo apt -q update && sudo apt -qy install g++-multilib
+      - build
+      - test
+
   clang-latest-ubsan:
     executor: linux
     environment:
@@ -247,3 +259,4 @@ workflows:
       - clang-latest-ubsan
       - clang-latest-coverage
       - macos-asan
+      - gcc-32bit


hrishikeshSuresh:
I noticed no CI test was performed on gcc-32-bit. Have I missed some file?

hrishikeshSuresh:
> I noticed no CI test was performed on gcc-32-bit. Have I missed some file?

I realised I didn't add gcc-32bit to workflow. Working on it now.

circle.yml
```diff
@@ -159,6 +159,21 @@ jobs:
           command: |
             codecov --required --gcov-exec "$GCOV" --gcov-root ~/build
 
+  gcc-32bit:
+    executor: linux
+    environment:
+      CMAKE_OPTIONS: -DTOOLCHAIN=cxx11-32bit
```
chfast: It does not build because evmone requires c++17. Set the toolchain to cxx17-32bit (if such toolchain file does not exist create it accordingly)

hrishikeshSuresh: done.

circle.yml
```diff
@@ -159,6 +159,18 @@ jobs:
...
 
+  gcc-32bit:
+    executor: linux
+    environment:
+      CMAKE_OPTIONS: -DTOOLCHAIN=cxx17-32bit
+    steps:
+      - run:
+          name: "Install GCC 32-bit"
+          command: |
+            sudo apt -q update && sudo apt -qy install g++-multilib && sudo apt -qy install gcc-multilib
```
chfast: I don't think you need to install `gcc-multilib` when `g++-multilib` is installed.

hrishikeshSuresh: done.

chfast:
Thanks a lot!

PR #199, opened 2019-10-23
cmake: Add version and soversion to evmone shared library

chfast:


axic:
@chfast will this mean only versioned filenames will be created?

chfast:
> @chfast will this mean only versioned filenames will be created?

The main file is with the full version + 2 symlinks:
```
-rw-r--r-- chfast/chfast  193936 2019-11-12 16:39 lib/libevmone.so.0.3.0-dev
lrwxrwxrwx chfast/chfast       0 2019-11-12 16:42 lib/libevmone.so.0.3 -> libevmone.so.0.3.0-dev
lrwxrwxrwx chfast/chfast       0 2019-11-12 16:42 lib/libevmone.so -> libevmone.so.0.3
```

axic:
Can someone then decide not to install the symlinks via the makefile (in a multi version environment)?

chfast:
> Can someone then decide not to install the symlinks via the makefile (in a multi version environment)?

At least I don't know how to do it. I expect that is not possible.

PR #201, opened 2019-10-23
Rework usage of <filesystem>

chfast:


gumb0:
What's the difference between ExternalProject and file dowload approaches?

chfast:
> What's the difference between ExternalProject and file download approaches?

The `file(DOWNLOAD)` is done straight away during configuration. The ExternalProject creates a CMake target and it is done during building. The first one is much simpler in case of single small file to be downloaded.

PR #204, opened 2019-11-06
Upgrade EVMC to version 7.0.0

chfast:
This is proof the release candidate of EVMC 7.0.0 is ok.


chfast:
@gumb0 @axic Do you want to revisit this one?

PR #206, opened 2019-11-12
fuzzer: Update to EVMC7, use Aleth 1.7.0-rc.1

chfast:


test/fuzzer/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ if(NOT fuzzing_coverage)
     # TODO: Aleth reports undefined behaviors, disable it for fuzzing.
     string(REPLACE undefined "" aleth_fuzzing_flags ${fuzzing_flags})
 
-    set(aleth_git_tag v1.7.0-alpha.1)
+    set(aleth_git_tag v1.7.0-rc.1)
```
axic: Shouldn't this be matched by bumpversion too? If not, should also fix the bumpversion config.

axic:
Where is this fuzzer run?

@bshastry would it be possible to submit this to ossfuzz?

test/fuzzer/CMakeLists.txt
axic: Ah never mind, this is the fixed version it loads.

chfast:
> Where is this fuzzer run?
> 
> @bshastry would it be possible to submit this to ossfuzz?

So far it runs on my machine. I need to put some work into Aleth's CMake to allow building aleth-interpreter without all Aleth's dependencies.

test/fuzzer/fuzzer.cpp
```diff
@@ -230,7 +232,7 @@ fuzz_input populate_input(const uint8_t* data, size_t data_size) noexcept
     if (data_size < input_size_16bits)  // Not enough data for input.
         return in;
 
-    in.rev = rev_4bits > EVMC_PETERSBURG ? EVMC_PETERSBURG : static_cast<evmc_revision>(rev_4bits);
+    in.rev = rev_4bits > latest_rev ? latest_rev : static_cast<evmc_revision>(rev_4bits);
```
axic: Why not fuzz for older versions though?

chfast: It does when `rev_4bits < latest_rev`. Only the `latest_rev` happens more frequently.

bshastry:
> Where is this fuzzer run?
> 
> @bshastry would it be possible to submit this to ossfuzz?

I guess I'm missing context. What is being fuzzed here? In general it should be possible to upstream it to ossfuzz if a static binary can be built because the machines on which google fuzzes a binary are different from where they are built.

PR #207, opened 2019-11-13
Fix UB around RETURN

chfast:


lib/evmone/instructions.cpp
```diff
@@ -802,8 +802,9 @@ const instruction* op_return(const instruction*, execution_state& state) noexcep
     if (!check_memory(state, offset, size))
         return nullptr;
 
-    state.output_offset = static_cast<size_t>(offset);
     state.output_size = static_cast<size_t>(size);
+    if (state.output_size != 0)
```
axic: Is this logic covered by the ethereum test suite?

chfast: What logic exactly?

axic: Well memory expansion should be handled in `check_memory`, that is being tested by the test suite.

PR #211, opened 2019-11-21
Optimize comparison instructions

chfast:


lib/evmone/instructions.cpp
```diff
@@ -177,38 +177,36 @@ auto op_signextend(const instruction* pc, execution_state& state) noexcept
...
 
 const instruction* op_slt(const instruction* instr, execution_state& state) noexcept
 {
-    auto x = state.stack[0];
-    auto y = state.stack[1];
-    auto x_neg = static_cast<bool>(x >> 255);
-    auto y_neg = static_cast<bool>(y >> 255);
-    state.stack[1] = (x_neg ^ y_neg) ? x_neg : x < y;
-    state.stack.pop();
+    // TODO: Move this to intx.
```
axic, line 194: @chfast was this included in intx?

chfast: No.

chfast:
This does not seem to change much, but merging this anyway...

PR #212, opened 2019-11-21
Use MockedHost from EVMC 7.1.0

chfast:
- Upgrade EVMC to 7.1.0.
- Use the MockedHost from EVMC (previously copied from here).


axic:
I guess you want to wait for the evmc7.1 release before this is merged?

PR #213, opened 2019-11-22
Update Dockerfile

chfast:
This removed the geth+EVMC from the docker image. I don't want to support it.
This also adds the benchmarking tool to the image.


chfast:
> @chfast since geth+EVMC are not supported (In the Dockerfile), I think the [README](https://github.com/ethereum/evmone#docker) file needs to be updated accordingly.

Updated, thanks.

PR #214, opened 2019-11-25
test: Solidity fuzz repro

chfast:

PR #215, opened 2019-11-26
Test: Add large bytecode generated by solidity

bshastry:
This is another very large test case (>200KB) that causes a different failure than #214 : null dereference in `op_push_small` function.


test/unittests/evm_other_test.cpp
```diff
@@ -56,3 +56,22 @@ TEST_F(evm_other, loop_full_of_jumpdests)
...
...d5b808301925050508060000180356001602003833603038112151562007a995760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007ac05760006000fd5b60208102360382131562007ad45760006000fd5b6000818110151562007ae257fe5b9050602002810180356001608003833603038112151562007b035760006000fd5b808301925050508060600180356001602003833603038112151562007b285760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007b4f5760006000fd5b60208102360382131562007b635760006000fd5b6001818110151562007b7157fe5b9050602002810180356001602003833603038112151562007b925760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007bb95760006000fd5b60208102360382131562007bcd5760006000fd5b6003818110151562007bdb57fe5b9050602002810180356001602003833603038112151562007bfc5760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007c235760006000fd5b60208102360382131562007c375760006000fd5b6000818110151562007c4557fe5b905060200201602062007c5c919081019062031fce565b6bffffffffffffffffffffffff191614151562007c7d576027905062020e6a565b7fbcba2c46958e61fa2fb4042cd7bc5816a710195b000000000000000000000000828060000180356001602003833603038112151562007cbd5760006000fd5b808301925050813590506020" +
```
bshastry: The reason the large string was split is because the clang compiler errors on string literals larger than 65535 bytes. However, what I'm doing here is also wrong. I need to figure out a way to populate code with smaller string literal fragments

test/unittests/evm_other_test.cpp
```diff
@@ -56,3 +56,26 @@ TEST_F(evm_other, loop_full_of_jumpdests)
...
+        "8201915067ffffffffffffffff81111562007ce45760006000fd5b60208102360382131562007cf85760006000fd5b6000818110151562007d0657fe5b9050602002810180356001606003833603038112151562007d275760006000fd5b808301925050508060000180356001602003833603038112151562007d4c5760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007d735760006000fd5b60208102360382131562007d875760006000fd5b6000818110151562007d9557fe5b9050602002810180356001602003833603038112151562007db65760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007ddd5760006000fd5b60208102360382131562007df15760006000fd5b6000818110151562007dff57fe5b9050602002810180356001602003833603038112151562007e205760006000fd5b808301925050508060000180356001602003833603038112151562007e455760006000fd5b8083019250508135905060208201915067ffffffffffffffff81111562007e6c5760006000fd5b60208102360382131562007e805760006000fd5b6000818110151562007e8e57fe5b9050602002810180356001608003833603038112151562007eaf5760006000fd5b808301925050508060600180356001602003833603038112151562007ed
```
bshastry: I'm pretty sure there are more elegant ways of concatenating string literals.

chfast:
Thanks. Can we close it now?

PR #219, opened 2019-12-02
Fix max stack growth overflow

chfast:


axic:
What was happening before the change? Overflowed and never returned "stack_overflow" ?

chfast:
> What was happening before the change? Overflowed and never returned "stack_overflow" ?

The stack overflow is not detected, but program crashes after accessing an out-of-bound stack item.

PR #221, opened 2019-12-03
Fix block gas cost overflow

chfast:
This avoids overflow when calculating a block gas cost using saturation to `UINT_MAX`. To reach the limit you now need a block of 400k instructions and 4G of gas.

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -8,6 +8,14 @@
 
 namespace evmone
 {
+/// Clamps x to the max value of To type.
+template <typename To, typename T>
+inline constexpr To clamp(T x) noexcept
+{
+    constexpr auto max = std::numeric_limits<To>::max();
+    return x <= max ? static_cast<To>(x) : max;
+}
+
 struct block_analysis
 {
     int64_t gas_cost = 0;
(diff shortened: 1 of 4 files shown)


test/unittests/evm_other_test.cpp
```diff
@@ -61,6 +61,52 @@ TEST_F(evm_other, evmone_block_max_stack_growth_overflow)
...
+    // The runtime values of arguments are not important.
+
+    constexpr auto n = 67'097;
+
+    auto code = bytecode{OP_MSIZE};
+    const auto create_pattern = bytecode{} + OP_DUP1 + OP_DUP1 + OP_CREATE;
+    code.reserve(create_pattern.size() * n);
+    for (int i = 0; i < n; ++i)
+        code += create_pattern;
+    EXPECT_EQ(code.size(), 201'292);
```
axic: Is this is a new C++ notation for numbers?

chfast: From C++17 I believe.

axic: Of course they couldn't go with the more widely used standard of underscores :wink:

chfast: I also wander why.

chfast:
Need fix build issues.

lib/evmone/analysis.hpp
```diff
@@ -89,7 +89,7 @@ struct block_info
 {
     /// The total base gas cost of all instructions in the block.
     /// This cannot overflow, see the static_assert() below.
-    int32_t gas_cost = 0;
+    uint32_t gas_cost = 0;
```
gumb0: Why is it unsigned now?

chfast: The basic block can have the cost of 4G instead of 2G.

lib/evmone/analysis.cpp
```diff
@@ -10,7 +10,7 @@ namespace evmone
 {
 struct block_analysis
 {
-    int gas_cost = 0;
+    int64_t gas_cost = 0;
```
gumb0: I think I see what's the point, it's just a bit confusing that this is signed while `block_info::gas_cost` is unsigned. Does it worth changing this to `uint64_t`?

lib/evmone/analysis.cpp
```diff
@@ -25,14 +25,19 @@ struct block_analysis
     /// Close the current block by producing compressed information about the block.
     [[nodiscard]] block_info close() const noexcept
     {
+        using gas_cost_type = decltype(block_info{}.gas_cost);
```
gumb0: Also confusing that "gas_cost" here means two different things, maybe rename this to `final_gas_cost_type` and variable below to `final_gas_cost_max`?

lib/evmone/analysis.cpp
chfast: Unfortunately, the signed type is better here. It is an error (detectable with undefined behavior sanitizer) to overflow signed types while unsigned have defined behavior on overflow.

The good way of thinking here is as follows: we use `int64_t` for gas computation (here and during execution), and `uint32_t` is used only for storing block gas cost.

lib/evmone/analysis.cpp
chfast: I will try to do something about naming here.

chfast: Refactored. No names any more :)

PR #222, opened 2019-12-06
utils: Cleanups

chfast:
- Clean ups #includes in utils
- Preallocates bytes buffer.
- Renames `to_hex()` to `hex()` (last commit, so can be reviewed separately).

### DIFF
--- a/test/unittests/utils_test.cpp
+++ b/test/unittests/utils_test.cpp
@@ -5,10 +5,10 @@
 #include <gtest/gtest.h>
 #include <test/utils/utils.hpp>
 
-TEST(utils, to_hex)
+TEST(utils, hex)
 {
     auto data = bytes{0x0, 0x1, 0xa, 0xf, 0x1f, 0xa0, 0xff, 0xf0};
-    EXPECT_EQ(to_hex(data), "00010a0f1fa0fff0");
+    EXPECT_EQ(hex(data), "00010a0f1fa0fff0");
 }
 
 TEST(utils, from_hex_empty)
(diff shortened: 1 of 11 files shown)


test/unittests/utils_test.cpp
```diff
@@ -5,10 +5,10 @@
 #include <gtest/gtest.h>
 #include <test/utils/utils.hpp>
 
-TEST(utils, to_hex)
+TEST(utils, hex)
 {
     auto data = bytes{0x0, 0x1, 0xa, 0xf, 0x1f, 0xa0, 0xff, 0xf0};
```
christianparpart: There were other places where you added a `const` modifier to not-modifying variables. Maybe here, too? (Depending on your general opinion about constness)

test/utils/utils.cpp
```diff
@@ -3,21 +3,20 @@
...
 
     bytes bs;
+    bs.reserve(hex.length() / 2);
     int b = 0;
     for (size_t i = 0; i < hex.size(); ++i)
     {
-        auto h = hex[i];
+        const auto h = hex[i];
         int v;
         if (h >= '0' && h <= '9')
```
christianparpart: somewhere deeper down in the else-branch, there is a throw statement that isn't documented in the header file (as the rest is). Would it make sense to state that?

chfast: Functions are not documented.

test/unittests/utils_test.cpp
chfast: I usually add `const` but these are just old unit tests, so it is probably not worth doing.

chfast:
> Just comments or suggestions, however, since this PR is meant to be a "cleanup" - what do you think of using a user defined literal (`_hexx`) instead of the `from_hexx()` free function?

Although used a lot in tests, these are meant for user input. Adding literal would be only syntax sugar as they cannot be `constexpr` anyway. We now also have `bytecode` type which makes most of the usage in unit tests obsolete.

PR #227, opened 2019-12-12
Unit test: Add abiv2 revert unit test

bshastry:
This test case adds the deployed byte code generated for an ABIv2 coder solidity test program targeting Istanbul.

The test entry point is public solidity function `test()` whose signature is `f8a8fd6d`. This is passed as input.

The bytecode execution is expected to be successful, but it reverts instead.


bshastry:
Closing this since I have a strong feeling that the bug is in the fuzzer harness and not evmone

Will re open if needed.

Issue #231, opened 2020-05-02
op_exp needs to have a shortcut for 2^e


AlexeyAkhunov:
There are a lot of contracts on the Ethereum mainnet that were compiled with Solidity prior to the shifting operations were available. For those, exponentiation `2^160` is used to emulate `1<<160`. Measurements performed on turbo-geth confirm this. Since we are planning to integrate evmone into turbo-geth, and we have already implemented this shortcut (https://github.com/ledgerwatch/turbo-geth/pull/505), it would be good to maintain that in evmone.

axic:
I think it would make sense considering an EIP for this, otherwise we are introducing yet another imbalance to "costs".

AlexeyAkhunov:
I disagree with the notion that every performance improvement needs to lead to the decreasing gas cost. The other way around - yes

PR #233, opened 2020-05-06
Shortcut for 2^e

AlexeyAkhunov:
There are a lot of contracts on the Ethereum mainnet that were compiled with Solidity prior to the shifting operations were available. For those, exponentiation 2^160 is used to emulate 1<<160. Measurements performed on turbo-geth confirm this. Since we are planning to integrate evmone into turbo-geth, and we have already implemented this shortcut (ledgerwatch/turbo-geth#505), it would be good to maintain that in evmone.

For https://github.com/ethereum/evmone/issues/231

### DIFF
--- a/lib/evmone/instructions.cpp
+++ b/lib/evmone/instructions.cpp
@@ -160,8 +160,10 @@ const instruction* op_exp(const instruction* instr, execution_state& state) noex
         auto exp_bytes = as_bytes(exponent);
         uint8_t bit_to_set = exp_bytes[0];
         exp_bytes[0] = 0;
-        exp_bytes[bit_to_set>>3] = uint8_t(uint8_t(1) << (bit_to_set&7));
-    } else {
+        exp_bytes[bit_to_set >> 3] = uint8_t(uint8_t(1) << (bit_to_set & 7));
+    }
+    else
+    {
         exponent = intx::exp(base, exponent);
     }
     return ++instr;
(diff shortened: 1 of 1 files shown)


lib/evmone/instructions.cpp
```diff
@@ -155,7 +155,17 @@ const instruction* op_exp(const instruction* instr, execution_state& state) noex
     if ((state.gas_left -= additional_cost) < 0)
         return state.exit(EVMC_OUT_OF_GAS);
 
-    exponent = intx::exp(base, exponent);
+    if (base == 2 && exponent < 256)
+    {
+        auto exp_bytes = as_bytes(exponent);
+        uint8_t bit_to_set = exp_bytes[0];
+        exp_bytes[0] = 0;
+        exp_bytes[bit_to_set >> 3] = uint8_t(uint8_t(1) << (bit_to_set & 7));
```
axic, line 163: Why not something like  `exponent = 2 << exponent;` ?

@chfast would either of these solutions have a significant speed difference?

chfast:
Yes, this makes sense, but I think I will move to the implementation of `intx::exp()`...

lib/evmone/instructions.cpp
chfast: Looks like at least in my benchmarks the `1 << exponent` is the fastest (surprisingly). Full story in https://github.com/chfast/intx/pull/146.

PR #235, opened 2020-06-03
Use intx::as_bytes

axic:
I had this stray change in my local branch.

### DIFF
--- a/lib/evmone/analysis.cpp
+++ b/lib/evmone/analysis.cpp
@@ -118,8 +118,7 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size) 
             const auto push_end = code_pos + push_size;
 
             auto& push_value = analysis.push_values.emplace_back();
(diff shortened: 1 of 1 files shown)


lib/evmone/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 include(LibraryTools)
 
 hunter_add_package(intx)
-find_package(intx CONFIG REQUIRED)
+find_package(intx 0.4.0 CONFIG REQUIRED)
```
chfast, line 8: What is the purpose of this? Dependencies versions are pinned by Hunter anyway.

axic: I wanted to ensure at least 0.4.0 is used, which is required for `as_bytes`, but if it is accomplished by hunter, then this is not needed.

chfast: Remove it please. I will handle that separately.

axic: Done.

PR #252, opened 2020-08-10
CREATE(2) should only populate return data on failure

yperbasis:
According to [EIP-211](https://eips.ethereum.org/EIPS/eip-211#specification), CREATE and CREATE2 should only populate return data on failure. See also the [implementation in Go Ethereum](https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L632).

### DIFF
--- a/lib/evmone/instructions.cpp
+++ b/lib/evmone/instructions.cpp
@@ -340,9 +340,16 @@ const instruction* op_create(const instruction* instr, execution_state& state) n
     msg.value = intx::be::store<evmc::uint256be>(endowment);
 
     auto result = state.host.call(msg);
-    state.return_data.assign(result.output_data, result.output_size);
+
     if (result.status_code == EVMC_SUCCESS)
+    {
         state.stack[0] = intx::be::load<uint256>(result.create_address);
+        state.return_data.clear();
+    }
+    else
+    {
+        state.return_data.assign(result.output_data, result.output_size);
+    }
 
     if ((state.gas_left -= msg.gas - result.gas_left) < 0)
         return state.exit(EVMC_OUT_OF_GAS);


axic:
Can you add a unit tests for this in `unittests/evm_calls_test.cpp`?

chfast:
Or you should not pass returned deployment code through. It is up to discussion which side should block the propagation.

https://github.com/ethereum/go-ethereum/pull/17954/files#diff-9633a3aa7f6fd345efaab4431bab4ae8R268-R273

yperbasis:
> Or you should not pass returned deployment code through. It is up to discussion which side should block the propagation.
> 
> https://github.com/ethereum/go-ethereum/pull/17954/files#diff-9633a3aa7f6fd345efaab4431bab4ae8R268-R273

It can be tricky to have a workaround on the host side. I encountered the issue while executing https://etherscan.io/tx/0x3ceeaa187dc8396f86cdfcf434705f4fd64c0f901c3e7268bdae336a794c290c

The deployment code is
```
0      PUSH1  => 21
2      DUP1
3      PUSH1  => 13
5      PUSH1  => 00
7      CODECOPY
8      PUSH1  => 00
10     PUSH1  => 00
12     CREATE
13     POP
14     RETURNDATASIZE
15     PUSH1  => 00
17     SSTORE
18     STOP
19     PUSH3  => 112233
23     PUSH1  => 00
25     MSTORE
26     PUSH1  => 20
28     PUSH1  => 00
30     PUSH1  => 20
32     PUSH1  => 00
34     PUSH1  => 00
36     PUSH1  => 04
38     PUSH2  => 9000
41     CALL
42     POP
43     RETURNDATASIZE
44     PUSH1  => 00
46     SSTORE
47     PUSH1  => 20
49     PUSH1  => 00
51     RETURN
```

The problem is that RETURNDATASIZE incorrectly yields non-0 after a successful CREATE returns. Because of nested calls I don't see how that can easily be mitigated on the host side.

yperbasis:
OK, fair enough, it can be done on the host side.

PR #253, opened 2020-08-31
Optimize away heavy operations discarded by POP

yperbasis:
This change speeds up execution of certain Shanghai attack blocks, which called EXTCODESIZE and immediately discarded it by POP.


lib/evmone/analysis.cpp
```diff
@@ -71,6 +71,28 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
...
+        // Optimize away heavy operations discarded by POP
+        if (code_pos != code_end && *code_pos == OP_POP && opcode_info.stack_req >= 1 &&
+            opcode_info.stack_change == 0)
+        {
+            bool read_only_and_const_cost = false;
+            switch (opcode)
+            {
+            case OP_BALANCE:
+            case OP_EXTCODESIZE:
+            case OP_EXTCODEHASH:
```
axic, line 83: Removing these won't mess with EIP-161 rules of "touching"?

yperbasis: It will not. Per [EIP-106](https://eips.ethereum.org/EIPS/eip-161), an account is considered to be touched when it's involved in a potential value transfer (SUICIDE beneficiary, CALL* value sender or recipient, miner rewards, CREATE(2) endowment).

axic: While it is not the case yet, there are proposals now, which would mean this optimisation would need to be potentially disabled in a later hard fork. I think leaving a comment here for implications would be nice.

I'm still not sure SLOAD is not affected by the net gas metering EIP, haven't checked.

axic: @yperbasis did you progress through istanbul with this change yet? Does any client implement such optimistation?

yperbasis: I've added more comments.

Only the cost of SSTORE, not SSLOAD is affected by [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200).

We're using this optimization in a new C++ Ethereum client [Silkworm](https://github.com/torquem-ch/silkworm) (an evolution of Turbo-Geth). Istanbul blocks work fine with this change.

lib/evmone/analysis.cpp
```diff
@@ -71,6 +71,35 @@ code_analysis analyze(evmc_revision rev, const uint8_t* code, size_t code_size)
...
+                pure_and_const_cost = true;
+            }
+
+            if (pure_and_const_cost)
+            {
+                // Warning: we can only replace discarded instruction with a no-op
+                // if it's pure (has no side effects) and its cost is constant.
+                // For example, an operation that can add an account to the touch set (EIP-161)
+                // cannot be considered pure and optimized away.
+                analysis.instrs.emplace_back(noop);
```
chfast, line 98: Have you tried no pushing any instruction here?

lib/evmone/analysis.cpp
axic: Btw, if you want more inspiration for optimisations, check out Solidity's ruleset together with the `libevmasm/SemanticInformation` file :wink:

lib/evmone/analysis.cpp
yperbasis: @chfast You're right, it's not necessary to push any instruction. I've updated the PR.

chfast:
I'm thinking of adding optimization level option to evmone: `O`.
The baseline would run at `O=0`, the current evmone implementation at `O=2`, the optimizations like this one may be enabled at `O=3`.

Heavily inspired by GCC compiler flags `-O0`, `-O2`, etc.

Thoughts?

yperbasis:
> I'm thinking of adding optimization level option to evmone: `O`.
> The baseline would run at `O=0`, the current evmone implementation at `O=2`, the optimizations like this one may be enabled at `O=3`.
> 
> Heavily inspired by GCC compiler flags `-O0`, `-O2`, etc.
> 
> Thoughts?

Well, it might be useful for debugging evmone, but, assuming everything is implemented correctly, why would anybody want a lower optimization level? I think for GCC it's different because -O3 might increase the compilation time.

chfast:
It is similar case for evmone: it can spend a lot of time on analyzing code which later execution costs 0 gas. The https://github.com/ethereum/evmone/pull/240 adds more pathological bytecodes to benchmark.

```
-----------------------------------------------------------------------------------------------
Benchmark                                        Time           CPU Iterations UserCounters...
-----------------------------------------------------------------------------------------------
micro/beginsub_push1s_0xffff/analysis          126 us        126 us       5531 rate=519.153M/s size=65.505k
micro/beginsub_push1s_0xffff                   129 us        129 us       5415 gas_rate=0/s gas_used=0
micro/beginsubs_0xffff/analysis                600 us        600 us       1118 rate=109.257M/s size=65.536k
micro/beginsubs_0xffff                         601 us        601 us       1133 gas_rate=0/s gas_used=0
micro/jumpdests_0xffff/analysis                346 us        346 us       2035 rate=189.203M/s size=65.535k
micro/jumpdests_0xffff                         456 us        456 us       1538 gas_rate=143.853M/s gas_used=65.535k
micro/loop_with_many_jumpdests/analysis        134 us        134 us       5201 rate=183.95M/s size=24.576k
micro/loop_with_many_jumpdests               13171 us      13171 us         53 gas_rate=606.487M/s gas_used=7.98788M
micro/push1s_0xffff/analysis                   135 us        135 us       5021 rate=484.837M/s size=65.535k
micro/push1s_0xffff                            145 us        145 us       4661 gas_rate=0/s gas_used=0
micro/push32s_0xffff/analysis                   42 us         42 us      16630 rate=1.57434G/s size=65.536k
micro/push32s_0xffff                            45 us         45 us      15079 gas_rate=0/s gas_used=0
micro/zeros_0xffff/analysis                    978 us        978 us        644 rate=67.0035M/s size=65.535k
micro/zeros_0xffff                             984 us        984 us        656 gas_rate=0/s gas_used=0
```

The last case takes ~1 ms to analyze, but the code is effectively `STOP`.

gumb0:
> @chfast Well, it might be useful for debugging evmone, but, assuming everything is implemented correctly, why would anybody want a lower optimization level? 

Not skipping instructions can be useful e.g. for collecting execution trace.

yperbasis:
@chfast @axic Guys, should I spend some time and add a test for this PR, or you think it's a bad idea and it should be closed?

axic:
@yperbasis sorry we were quite busy with some other EVM related work as you can see the tracing/benchmarking PRs. I think we should also look more into the speed differences between baseline vs. advanced and see how much different discarding something like POP would cause on the baseline as well.

Probably it will take us til next week before we can come back with good feedback here.

yperbasis:
I've checked with [Silkworm](https://github.com/torquem-ch/silkworm) and the performance gain of this change in not that significant: 0.1h out of 17h to execute the first 11.36M blocks. So I'm closing this PR.

chfast:
Anyway, I was planning to add some basic optimization as a hook to evmone advanced analysis. Then it could be enabled under `O=3`. Let me know if anyone is interested.

PR #255, opened 2020-10-08
Expose call instructions

chfast:


lib/evmone/instructions.cpp
```diff
@@ -286,66 +193,14 @@ const instruction* op_create(const instruction* instr, execution_state& state) n
 {
     const auto gas_left_correction = state.current_block_cost - instr->arg.number;
 
-    if (state.msg.flags & EVMC_STATIC)
-        return state.exit(EVMC_STATIC_MODE_VIOLATION);
-
-    auto endowment = state.stack[0];
-    auto init_code_offset = state.stack[1];
-    auto init_code_size = state.stack[2];
+    state.gas_left += gas_left_correction;
```
gumb0, line 196: I don't remember how this `gas_left_correction` works.
Isn't this affecting the check later in ` if constexpr (Kind == EVMC_CREATE2)` ?
Before it was comparing `if (state.gas_left < 0)` and now `gas_left` is increased.

lib/evmone/instructions.cpp
```diff
@@ -172,112 +172,19 @@ const instruction* op_return(const instruction*, execution_state& state) noexcep
     return state.exit(status_code);
 }
 
+
```
gumb0, line 175: Nit: extra line

lib/evmone/instructions.cpp
chfast: evmone counts gas in blocks so the `state.gas_left` is usually lower the "spec" gas left. For instructions that require to access the "spec" gas left this expression brings `state.gas_left` to the spec value.

The check `if (state.gas_left < 0)` is only needed for block gas counting because the spec `CREATE` cannot run out of gas (except for initial costs).
Here this check is moved after generic `create<Kind>()` to `if ((state.gas_left -= gas_left_correction) < 0)` where we also convert `state.gas_left` to block gas counting mode.

gumb0: I remember it better now, thanks.

I was referring to the check after charging for salt in CREATE2 case. Before it was checking against block gas, and now it is checking against "spec" gas. As I understand now it has one final check against block gas in the end `op_create` instead of this, and it should end with the same result, except that previously `op_create` could return OOG earlier, but it shouldn't be a problem.

gumb0 (approved):

PR #266, opened 2020-11-02
Add ExecutionState::clear() to allow execution state reuse

yperbasis:


chfast (commented):
To my understanding this only makes the `ExecutionState` a "clearable" type, but state reuse is not part of this change.

yperbasis:
> is

yperbasis:
> To my understanding this only makes the `ExecutionState` a "clearable" type, but state reuse is not part of this change.

That's right. There's an example of reuse in [Silkworm](https://github.com/torquem-ch/silkworm/blob/master/silkworm/execution/evm.cpp#L228).

test/unittests/execution_state_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+    EXPECT_EQ(st.output_offset, 0);
+    EXPECT_EQ(st.output_size, 0);
+    EXPECT_EQ(st.current_block_cost, 0u);
+    EXPECT_EQ(st.analysis, nullptr);
+}
+
+TEST(execution_state, stack_clear)
+{
+    evmone::evm_stack stack;
+    EXPECT_EQ(stack.size(), 0);
```
gumb0, line 134: You could add also `clear()` call when stack is empty.

gumb0 (approved):
This looks fine to me, but it seems your use case would benefit more from the method like `execution_state::reset(msg, rev, host, code, analysis)`

Then you also wouldn't need a default constuctor, that produces `ExecutionState` in invalid state (null `msg` and `host`)

chfast:
> This looks fine to me, but it seems your use case would benefit more from the method like `execution_state::reset(msg, rev, host, code, analysis)`

I was thinking of having `init()` method like that, but this looks better.

yperbasis:
Looks good to me

chfast:
@yperbasis, I have added a change which implements `.reset()` instead of `.clear()` as @gumb0 suggested. Please test it in Silkworm as this is my preferred version.

yperbasis:
> @yperbasis, I have added a change which implements `.reset()` instead of `.clear()` as @gumb0 suggested. Please test it in Silkworm as this is my preferred version.

gas_left in reset should be set to message.gas instead of 0

yperbasis:
It works fine for Silkworm now, thank you

chfast:
> > This looks fine to me, but it seems your use case would benefit more from the method like `execution_state::reset(msg, rev, host, code, analysis)`
> 
> I was thinking of having `init()` method like that, but this looks better.

I made the change. I would welcome another review @gumb0.

lib/evmone/execution_state.hpp
```diff
@@ -84,23 +90,43 @@ struct ExecutionState
...
-    const evmc_revision rev = {};
+    evmc_revision rev = {};
     bytes return_data;
-    const bytes_view code;
+    bytes_view code;
     evmc_status_code status = EVMC_SUCCESS;
     size_t output_offset = 0;
     size_t output_size = 0;
 
+    ExecutionState() noexcept = default;
```
gumb0, line 102: Is it needed?

chfast: I believe Silkwarm first creates a ExecutionState in a pool with `std::make_unique<ExecutionState>()`.

yperbasis: > Is it needed?

Yes, the default constructor is used in [Silkworm](https://github.com/torquem-ch/silkworm/blob/master/silkworm/execution/state_pool.cpp).

gumb0 (approved):

PR #268, opened 2020-11-16
Refactor instruction tables, add compile-time information

chfast:
Convert instruction metrics to compile-time traits. The traits can then be used to construct efficient implementation-specific instruction tables.

### DIFF
--- a/lib/evmone/instructions.cpp
+++ b/lib/evmone/instructions.cpp
@@ -4,6 +4,7 @@
 
 #include "instructions.hpp"
 #include "analysis.hpp"
+#include "instruction_traits.hpp"
 
 namespace evmone
 {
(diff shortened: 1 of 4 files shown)


lib/evmone/instructions.cpp
```diff
@@ -195,190 +196,173 @@ const instruction* opx_beginblock(const instruction* instr, execution_state& sta
...
+    create_op_table<EVMC_BYZANTIUM>(),
+    create_op_table<EVMC_CONSTANTINOPLE>(),
+    create_op_table<EVMC_PETERSBURG>(),
+    create_op_table<EVMC_ISTANBUL>(),
+    create_op_table<EVMC_BERLIN>(),
 };
-static_assert(sizeof(op_tables) / sizeof(op_tables[0]) > EVMC_MAX_REVISION,
-    "op table entry missing for an EVMC revision");
+static_assert(
+    std::size(op_tables) > EVMC_MAX_REVISION, "op table entry missing for an EVMC revision");
```
gumb0, line 365: maybe could be more strict
```suggestion
    std::size(op_tables) == EVMC_MAX_REVISION + 1, "op table entry missing for an EVMC revision");
```

gumb0 (approved):

PR #274, opened 2020-12-15
cmake: Upgrade dependencies

chfast:
- intx to 0.5.1
- ethash to 0.6.0

### DIFF
--- a/cmake/Hunter/config.cmake
+++ b/cmake/Hunter/config.cmake
@@ -1,10 +1,17 @@
 # evmone: Fast Ethereum Virtual Machine implementation
-# Copyright 2018-2020 The evmone Authors.
+# Copyright 2018 The evmone Authors.
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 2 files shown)


cmake/Hunter/config.cmake
```diff
@@ -1,10 +1,17 @@
...
-    VERSION 0.5.0
-    URL https://github.com/chfast/intx/archive/v0.5.0.tar.gz
-    SHA1 7f917cfe9046278d348a5cf36c20f9847d351807
+    VERSION 0.5.1
+    URL https://github.com/chfast/intx/archive/v0.5.1.tar.gz
+    SHA1 743c46a82750143bd302a4394b7008a2112fc97b
+)
+
+hunter_config(
+    ethash
```
axic, line 13: Where was this taken from before?

chfast: From the Hunter defaults.

PR #275, opened 2020-12-16
Refactor evmone-bench

chfast:


test/bench/bench.cpp
```diff
@@ -27,12 +27,51 @@ using namespace benchmark;
...
+            expected_output{std::move(_expected_output)}
+        {}
+    };
+
+    std::string name;
+    bytes code;
+    std::vector<Input> inputs;
+
+    /// Create a benchmark case without input.
+    explicit BenchmarkCase(std::string _name, bytes _code) noexcept
```
gumb0, line 50:
```suggestion
    BenchmarkCase(std::string _name, bytes _code) noexcept
```

test/bench/bench.cpp
```diff
@@ -184,11 +200,18 @@ void load_benchmarks_from_dir(const fs::path& path, const std::string& name_pref
...
+
     for (const auto& f : files)
-        load_benchmark(f, name_prefix);
+        benchmark_cases.emplace_back(load_benchmark(f, name_prefix));
 
     for (const auto& d : subdirs)
-        load_benchmarks_from_dir(d, name_prefix + d.filename().string() + '/');
+    {
+        const auto t = load_benchmarks_from_dir(d, name_prefix + d.filename().string() + '/');
+        benchmark_cases.insert(benchmark_cases.end(), std::cbegin(t), std::cend(t));
```
gumb0, line 211:
```suggestion
        auto t = load_benchmarks_from_dir(d, name_prefix + d.filename().string() + '/');
        benchmark_cases.insert(benchmark_cases.end(), std::make_move_iterator(t.begin()), std::make_move_iterator(t.end()));
```

test/bench/bench.cpp
```diff
@@ -41,69 +80,87 @@ inline evmc::result execute(bytes_view code, bytes_view input) noexcept
...
+            st = state::input;
+            break;
+
+        case state::input:
+            input = from_hexx(l);
+            st = state::expected_output;
+            break;
+
+        case state::expected_output:
+            inputs.emplace_back(input_name, std::move(input), from_hexx(l));
```
gumb0, line 151:
```suggestion
            inputs.emplace_back(std::move(input_name), std::move(input), from_hexx(l));
```

test/bench/bench.cpp
```diff
@@ -271,13 +289,37 @@ int parseargs(int argc, char** argv)
...
+        BenchmarkCase b{code_hex_file, from_hex(code_hex)};
+        b.inputs.emplace_back("", from_hex(input_hex), from_hex(expected_output_hex));
+
+        return {0, {std::move(b)}};
+    }
+}
+
+void register_benchmarks(const std::vector<BenchmarkCase>& benchmark_cases)
+{
+    for (auto& b : benchmark_cases)
```
gumb0, line 301:
```suggestion
    for (const auto& b : benchmark_cases)
```

test/bench/bench.cpp
```diff
@@ -271,13 +289,37 @@ int parseargs(int argc, char** argv)
...
+    {
+        if (registered_vms.count("advanced"))
+        {
+            RegisterBenchmark(("advanced/analyse/" + b.name).c_str(), [&b](State& state) {
+                analyse(state, b.code);
+            })->Unit(kMicrosecond);
+        }
+
+        const auto execute_name = "/execute/" + b.name;
+        for (auto& input : b.inputs)
```
gumb0, line 311:
```suggestion
        for (const auto& input : b.inputs)
```

test/bench/bench.cpp
```diff
@@ -27,12 +27,51 @@ using namespace benchmark;
 
 namespace
 {
-constexpr auto gas_limit = std::numeric_limits<int64_t>::max();
-auto vm = evmc::VM{};
+struct BenchmarkCase
+{
+    struct Input
+    {
+        std::string name;  ///< The input name, contains / prefix for convenience.
```
gumb0, line 34: Wouldn't it be simpler to add all required `/` at the point of call to `RegisterBenchmark` ?

chfast: It was like that, but this allow us to add `""` input if there is none specified so we don't special case for "no inputs" at registration.

gumb0: It could be just a special case at benchmark name construction (unless input name is empty, add `/` + input name)

Now this comment is kind of not true for empty name.

chfast: Fixed.

gumb0 (approved):

PR #278, opened 2020-12-18
EVM instructions synthetic benchmarks

chfast:
See: https://notes.ethereum.org/@chfast/benchmarking-evm-instructions

### DIFF
--- a/test/bench/synthetic_benchmarks.cpp
+++ b/test/bench/synthetic_benchmarks.cpp
@@ -1,0 +1,277 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2020 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "synthetic_benchmarks.hpp"
+#include "helpers.hpp"
+#include "test/utils/bytecode.hpp"
+#include <evmc/instructions.h>
+#include <evmone/instruction_traits.hpp>
+
+using namespace benchmark;
+
+namespace evmone::test
+{
+namespace
+{
+/// Stack limit inside the EVM benchmark loop (one stack item is used for the loop counter).
+constexpr auto stack_limit = 1023;
+
+enum class Mode
+{
+    min_stack = 0,   ///< The code uses as minimal stack as possible.
+    full_stack = 1,  ///< The code fills the stack up to its limit.
+};
+
+/// The instruction grouping by EVM stack requirements.
+enum class InstructionCategory : char
+{
+    nop = 'n',     ///< No-op instruction.
+    nullop = 'a',  ///< Nullary operator - produces a result without any stack input.
+    unop = 'u',    ///< Unary operator.
+    binop = 'b',   ///< Binary operator.
+    push = 'p',    ///< PUSH instruction.
+    dup = 'd',     ///< DUP instruction.
+    swap = 's',    ///< SWAP instruction.
+    other = 'X',   ///< Not any of the categories above.
+};
+
+constexpr InstructionCategory get_instruction_category(evmc_opcode opcode) noexcept
+{
+    const auto trait = instr::traits[opcode];
+    if (opcode >= OP_PUSH1 && opcode <= OP_PUSH32)
+        return InstructionCategory::push;
+    else if (opcode >= OP_SWAP1 && opcode <= OP_SWAP16)
+        return InstructionCategory::swap;
+    else if (opcode >= OP_DUP1 && opcode <= OP_DUP16)
+        return InstructionCategory::dup;
+    else if (trait.stack_height_required == 0 && trait.stack_height_change == 0)
+        return InstructionCategory::nop;
+    else if (trait.stack_height_required == 0 && trait.stack_height_change == 1)
+        return InstructionCategory::nullop;
+    else if (trait.stack_height_required == 1 && trait.stack_height_change == 0)
+        return InstructionCategory::unop;
+    else if (trait.stack_height_required == 2 && trait.stack_height_change == -1)
+        return InstructionCategory::binop;
+    else
+        return InstructionCategory::other;
+}
+
+struct CodeParams
+{
+    evmc_opcode opcode;
+    Mode mode;
+};
+
+/// The less-than comparison operator. Needed for std::map.
+[[maybe_unused]] inline constexpr bool operator<(const CodeParams& a, const CodeParams& b) noexcept
+{
+    return std::tuple(a.opcode, a.mode) < std::tuple(b.opcode, b.mode);
+}
+
+std::string to_string(const CodeParams& params)
+{
+    return std::string{instr::traits[params.opcode].name} + '/' +
+           static_cast<char>(get_instruction_category(params.opcode)) +
+           std::to_string(static_cast<int>(params.mode));
+}
+
+/// Generates the EVM benchmark loop inner code for the given opcode and "mode".
+bytecode generate_loop_inner_code(CodeParams params)
+{
+    const auto [opcode, mode] = params;
+    const auto category = get_instruction_category(opcode);
+    switch (mode)
+    {
+    case Mode::min_stack:
+        switch (category)
+        {
+        case InstructionCategory::nop:
+            // JUMPDEST JUMPDEST ...
+            return stack_limit * 2 * bytecode{opcode};
+
+        case InstructionCategory::nullop:
+            // CALLER POP CALLER POP ...
+            return stack_limit * (bytecode{opcode} + OP_POP);
+
+        case InstructionCategory::unop:
+            // DUP1 NOT NOT ... POP
+            return OP_DUP1 + stack_limit * 2 * bytecode{opcode} + OP_POP;
+
+        case InstructionCategory::binop:
+            // DUP1 DUP1 ADD DUP1 ADD DUP1 ADD ... POP
+            return OP_DUP1 + (stack_limit - 1) * (OP_DUP1 + bytecode{opcode}) + OP_POP;
+
+        case InstructionCategory::push:
+            // PUSH1 POP PUSH1 POP ...
+            return stack_limit * (push(opcode, {}) + OP_POP);
+
+        case InstructionCategory::dup:
+        {
+            // The required n stack height for DUPn is provided by
+            // duplicating the loop counter n-1 times with DUP1.
+            const auto n = opcode - OP_DUP1 + 1;
+            // DUP1 ...  DUPn POP DUPn POP ...  POP ...
+            // \ n-1  /                         \ n-1 /
+            return (n - 1) * OP_DUP1 +                // Required n stack height.
+                   (stack_limit - (n - 1)) *          //
+                       (bytecode{opcode} + OP_POP) +  // Multiple DUPn POP pairs.
+                   (n - 1) * OP_POP;                  // Pop initially duplicated values.
+        }
+
+        case InstructionCategory::swap:
+        {
+            // The required n+1 stack height for SWAPn is provided by duplicating the loop counter
+            // n times with DUP1. This also guarantees the loop counter remains unchanged because
+            // it is always going to be swapped to the same value.
+            const auto n = opcode - OP_SWAP1 + 1;
+            // DUP1 ...  SWAPn SWAPn ...  POP ...
+            // \  n   /                   \  n  /
+            return n * OP_DUP1 +                         // Required n+1 stack height.
+                   stack_limit * 2 * bytecode{opcode} +  // Multiple SWAPns.
+                   n * OP_POP;                           // Pop initially duplicated values.
+        }
+
+        default:
+            break;
+        }
+        break;
+
+    case Mode::full_stack:
+        switch (category)
+        {
+        case InstructionCategory::nullop:
+            // CALLER CALLER ... POP POP ...
+            return stack_limit * opcode + stack_limit * OP_POP;
+
+        case InstructionCategory::binop:
+            // DUP1 DUP1 DUP1 ... ADD ADD ADD ... POP
+            return stack_limit * OP_DUP1 + (stack_limit - 1) * opcode + OP_POP;
+
+        case InstructionCategory::push:
+            // PUSH1 PUSH1 PUSH1 ... POP POP POP ...
+            return stack_limit * push(opcode, {}) + stack_limit * OP_POP;
+
+        case InstructionCategory::dup:
+        {
+            // The required initial n stack height for DUPn is provided by
+            // duplicating the loop counter n-1 times with DUP1.
+            const auto n = opcode - OP_DUP1 + 1;
+            // DUP1 ...  DUPn DUPn ...  POP POP ...
+            // \ n-1  /  \  S-(n-1)  /  \    S    /
+            return (n - 1) * OP_DUP1 +                           // Required n stack height.
+                   (stack_limit - (n - 1)) * bytecode{opcode} +  // Fill the stack with DUPn.
+                   stack_limit * OP_POP;                         // Clear whole stack.
+        }
+
+        default:
+            break;
+        }
+        break;
+    }
+
+    return {};
+}
+
+/// Generates a benchmark loop with given inner code.
+///
+/// This generates do-while loop with 255 iterations and it starts with PUSH1 of 255 as the loop
+/// counter. The while check is done as `(counter += -1) != 0`. The SUB is avoided because it
+/// consumes arguments in unnatural order and additional SWAP would be required.
+///
+/// The loop counter stays on the stack top. The inner code is allowed to duplicate it, but must not
+/// modify it.
+bytecode generate_loop_v1(const bytecode& inner_code)
+{
+    const auto counter = push(255);
+    const auto jumpdest_offset = counter.size();
+    return counter + OP_JUMPDEST + inner_code +  // loop label + inner code
+           push("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") +  // -1
+           OP_ADD + OP_DUP1 +                 // counter += (-1)
+           push(jumpdest_offset) + OP_JUMPI;  // jump to jumpdest_offset if counter != 0
+}
+
+/// Generates a benchmark loop with given inner code.
+///
+/// This is improved variant of v1. It has exactly the same instructions and consumes the same
+/// amount of gas, but according to performed benchmarks (see "loop_v1" and "loop_v2") it runs
+/// faster. And we want the lowest possible loop overhead.
+/// The change is to set the loop counter to -255 and check `(counter += 1) != 0`.
+bytecode generate_loop_v2(const bytecode& inner_code)
+{
+    const auto counter =
+        push("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff01");  // -255
+    const auto jumpdest_offset = counter.size();
+    return counter + OP_JUMPDEST + inner_code +  // loop label + inner code
+           push(1) + OP_ADD + OP_DUP1 +          // counter += 1
+           push(jumpdest_offset) + OP_JUMPI;     // jump to jumpdest_offset if counter != 0
+}
+
+bytes_view generate_code(CodeParams params)
+{
+    static std::map<CodeParams, bytecode> cache;
+
+    auto& code = cache[params];
+    if (!code.empty())
+        return code;
+
+    code = generate_loop_v2(generate_loop_inner_code(params));  // Cache it.
+    return code;
+}
+}  // namespace
+
+void register_synthetic_benchmarks()
+{
+    std::vector<CodeParams> params_list;
+
+    // Nops & unops.
+    for (const auto opcode : {OP_JUMPDEST, OP_ISZERO, OP_NOT})
+        params_list.push_back({opcode, Mode::min_stack});
+
+    // Binops.
+    for (const auto opcode : {OP_ADD, OP_MUL, OP_SUB, OP_SIGNEXTEND, OP_LT, OP_GT, OP_SLT, OP_SGT,
+             OP_EQ, OP_AND, OP_OR, OP_XOR, OP_BYTE, OP_SHL, OP_SHR, OP_SAR})
+        params_list.insert(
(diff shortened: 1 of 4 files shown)


test/bench/helpers.hpp
```diff
@@ -0,0 +1,77 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2019 The evmone Authors.
```
gumb0, line 2:
```suggestion
// Copyright 2021 The evmone Authors.
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+{
+namespace
+{
+/// Stack limit inside the EVM benchmarking loop (one stack item is used for the loop count).
+constexpr auto stack_limit = 1023;
+
+enum class Mode
+{
+    min_stack = 0,   ///< The code uses as minimal stack as possible.
+    full_stack = 1,  ///< The code fill the stuck up to its limit.
```
gumb0, line 23:
```suggestion
    full_stack = 1,  ///< The code fills the stack up to its limit.
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+bytecode generate_loop_inner_code(evmc_opcode opcode, Mode mode)
+{
+    const auto kind = get_instruction_kind(opcode);
+    switch (mode)
+    {
+    case Mode::min_stack:
+        switch (kind)
+        {
+        case InstructionKind::push:
+            return stack_limit * (push(opcode, {}) + OP_POP);
```
gumb0, line 60: What's the idea behind using `stack_limit` as a limit in `min_stack` mode? It could as well fill up the code size limit.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+        case InstructionKind::producer:
+            return stack_limit * (bytecode{opcode} + OP_POP);
+        case InstructionKind::unop:
+            return OP_DUP1 + stack_limit * 2 * bytecode{opcode} + OP_POP;
```
gumb0, line 64: I think adding here in comments examples like in hackmd note  would be helpful
```
// DUP1 NOT NOT NOT NOT NOT NOT NOT ...
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+    default:
+        INTX_UNREACHABLE();
+    }
+
+    return {};  // Make old compilers happy.
+}
+
+
+const auto loop_prefix = push(255) + OP_JUMPDEST;
+const auto loop_suffix = push("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") +
```
gumb0, line 91: Why is it not `push(1)  + OP_SUB` ?

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+                         OP_ADD + OP_DUP1 + push(2) + OP_JUMPI;
+
+bytes_view generate_code(evmc_opcode opcode, Mode mode) noexcept
```
gumb0, line 94: `generate_loop_inner_code` is not `noexcept`
```suggestion
bytes_view generate_code(evmc_opcode opcode, Mode mode)
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+        RegisterBenchmark((std::string{vm_name} + "/execute/synth/loop").c_str(),
+            [&vm = vm](State& state) {
+                const auto code = loop_prefix + loop_suffix;
+                execute(state, vm, code);
+            })
+            ->Unit(kMicrosecond);
+    }
+
+    for (auto opcode : opcodes)
```
gumb0, line 133:
```suggestion
    for (const auto opcode : opcodes)
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+        const auto kind = get_instruction_kind(opcode);
+
+        for (auto mode : {Mode::min_stack, Mode::full_stack})
```
gumb0, line 137:
```suggestion
        for (const auto mode : {Mode::min_stack, Mode::full_stack})
```

test/bench/helpers.hpp
chfast: I moved it from other place only.

test/bench/synthetic_benchmarks.cpp
chfast: You can notice that both `min_stack` and `full_stack` have exactly the same instructions just with different order.

test/bench/synthetic_benchmarks.cpp
chfast: It would be `push(1) + OP_SWAP1 + OP_SUB`.

chfast: The loop can also be 2^256-256 and the +1 until 0. We can experiment with this and measure overhead later.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+{
+    const auto [opcode, mode] = params;
+    const auto kind = get_instruction_kind(opcode);
+    switch (mode)
+    {
+    case Mode::min_stack:
+        switch (kind)
+        {
+        case InstructionKind::push:
+            return stack_limit * (push(opcode, {}) + OP_POP);
```
gumb0, line 88:
```suggestion
            // PUSH1 POP PUSH1 POP ...
            return stack_limit * (push(opcode, {}) + OP_POP);
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+        case InstructionKind::swap:
+        {
+            const auto n = opcode - OP_SWAP1 + 1;
+            return n * OP_DUP1 + stack_limit * 2 * bytecode{opcode} + n * OP_POP;
```
gumb0, line 92:
```suggestion
            // DUP1 DUP1 ... SWAP1 SWAP1 ... POP POP ...
            return n * OP_DUP1 + stack_limit * 2 * bytecode{opcode} + n * OP_POP;
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+        case InstructionKind::dup:
+        {
+            const auto n = opcode - OP_DUP1;
+            return n * OP_DUP1 + stack_limit * (bytecode{opcode} + OP_POP) + n * OP_POP;
```
gumb0, line 97:
```suggestion
            // DUP1 DUP1 ... DUP1 POP DUP1 POP ... POP POP ...
            return n * OP_DUP1 + stack_limit * (bytecode{opcode} + OP_POP) + n * OP_POP;
```

test/bench/synthetic_benchmarks.cpp
gumb0:
```suggestion
            // DUP1 NOT NOT ... POP
            return OP_DUP1 + stack_limit * 2 * bytecode{opcode} + OP_POP;
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+        case InstructionKind::producer:
+            return stack_limit * (bytecode{opcode} + OP_POP);
+        case InstructionKind::nullop:
+            return stack_limit * 2 * bytecode{opcode};
+        case InstructionKind::unop:
+            return OP_DUP1 + stack_limit * 2 * bytecode{opcode} + OP_POP;
+        case InstructionKind::binop:
+            return OP_DUP1 + (stack_limit - 1) * (OP_DUP1 + bytecode{opcode}) + OP_POP;
```
gumb0, line 106:
```suggestion
            // DUP1 DUP1 ADD DUP1 ADD DUP1 ADD ... POP
            return OP_DUP1 + (stack_limit - 1) * (OP_DUP1 + bytecode{opcode}) + OP_POP;
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+        default:
+            break;
+        }
+        break;
+    case Mode::full_stack:
+        switch (kind)
+        {
+        case InstructionKind::push:
+            return stack_limit * push(opcode, {}) + stack_limit * OP_POP;
```
gumb0, line 115:
```suggestion
            // PUSH1 PUSH1 PUSH1 ... POP POP POP ...
            return stack_limit * push(opcode, {}) + stack_limit * OP_POP;
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+        case InstructionKind::producer:
+            return stack_limit * opcode + stack_limit * OP_POP;
```
gumb0, line 117:
```suggestion
            // CALLER CALLER ... POP POP ...
            return stack_limit * opcode + stack_limit * OP_POP;
```

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,202 @@
...
+        case InstructionKind::binop:
+            return stack_limit * OP_DUP1 + (stack_limit - 1) * opcode + OP_POP;
```
gumb0, line 119:
```suggestion
            // DUP1 DUP1 DUP1 ... ADD ADD ADD ... POP
            return stack_limit * OP_DUP1 + (stack_limit - 1) * opcode + OP_POP;
```

test/bench/synthetic_benchmarks.cpp
gumb0, line 92: If other cases have all `stack_limit` tested opcodes, why does this one have `2 * stack_limit`?

test/bench/synthetic_benchmarks.cpp
chfast: I tried the other loop and it seems better. v2 is being used now, both are benchmarked.

test/bench/synthetic_benchmarks.cpp
chfast: To have similar number of instructions in the inner code. This not strictly needed (any number here works), but should allows easier eyeball results comparison (and loop overhead should have similar impact).

I was also considering changing everything to `4*stack_limit` to also use full stack for unops: `DUP1 NOT NOT DUP1 NOT NOT ... POP POP POP ...`.

But it looks like speed difference between "full stack" and "min stack" is due too branch misprediction than to memory cache miss.

test/utils/bytecode.hpp
```diff
@@ -76,6 +77,20 @@ inline bytecode push(std::string_view hex_data)
     return push(from_hex(hex_data));
 }
 
+bytecode push(evmc_opcode opcode) = delete;
+
+inline bytecode push(evmc_opcode opcode, const bytecode& data)
```
axic, line 82: I think if you want to reduce the scope of this PR, this helper with its tests could be merged outright.

chfast: https://github.com/ethereum/evmone/pull/286

test/bench/helpers.hpp
```diff
@@ -0,0 +1,77 @@
...
+    state.counters["rate"] = Counter(static_cast<double>(bytes_analysed), Counter::kIsRate);
+}
+
+inline evmc::result execute(evmc::VM& vm, bytes_view code, bytes_view input) noexcept
+{
+    auto msg = evmc_message{};
+    msg.gas = gas_limit;
+    msg.input_data = input.data();
+    msg.input_size = input.size();
+    return vm.execute(EVMC_ISTANBUL, msg, code.data(), code.size());
```
axic, line 38: Since `EVMC_ISTANBUL` is duplicated twice in this file, perhaps a `constexpr evmc_version` on the top might be a good idea to have a single place for the setting? We don't have `EVMC_LATEST`, right? I think we removed that.

test/bench/helpers.hpp
```diff
@@ -0,0 +1,77 @@
...
+        benchmark::DoNotOptimize(r);
+        bytes_analysed += code.size();
+    }
+
+    using benchmark::Counter;
+    state.counters["size"] = Counter(static_cast<double>(code.size()));
+    state.counters["rate"] = Counter(static_cast<double>(bytes_analysed), Counter::kIsRate);
+}
+
+inline evmc::result execute(evmc::VM& vm, bytes_view code, bytes_view input) noexcept
```
axic, line 32: I think it would make the loops in the caller more intuitive if the gas_limit is passed as an argument.

test/bench/helpers.hpp
```diff
@@ -0,0 +1,77 @@
...
+            state.SkipWithError(("failure: " + std::to_string(r.status_code)).c_str());
+            return;
+        }
+
+        if (!expected_output.empty())
+        {
+            const auto output = bytes_view{r.output_data, r.output_size};
+            if (output != expected_output)
+            {
+                auto error = "got: " + hex(output) + "  expected: " + hex(expected_output);
```
axic, line 57: Nit: Why is this split into two lines with a temp variable, while it is merged above?

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,240 @@
...
+        }
+        break;
+    }
+
+    return {};
+}
+
+/// Generates a benchmarking loop with given inner code.
+///
+/// This generates do-while loop with 256 iterations and is starts with PUSH1 of 255 as the loop
```
axic, line 148:
```suggestion
/// This generates do-while loop with 256 iterations and it starts with PUSH1 of 255 as the loop
```

chfast: Ha. I actually miscalculated the loop iterations. It should be 255. Someone please double check.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,241 @@
...
+bytecode generate_loop_inner_code(CodeParams params)
+{
+    const auto [opcode, mode] = params;
+    const auto kind = get_instruction_kind(opcode);
+    switch (mode)
+    {
+    case Mode::min_stack:
+        switch (kind)
+        {
+        case InstructionKind::push:
```
chfast, line 89: I will order the switch cases by the enum definition.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,241 @@
...
+///
+/// This generates do-while loop with 255 iterations and it starts with PUSH1 of 255 as the loop
+/// counter. The while check is done as `(counter += -1) != 0`. The SUB is avoided because it
+/// consumes arguments in unnatural order and additional SWAP would be required.
+///
+/// The loop counter stays on the stack top. The inner code is allowed to duplicate it, but must not
+/// modify it.
+bytecode generate_loop_v1(const bytecode& inner_code)
+{
+    return push(255) + OP_JUMPDEST + inner_code +
```
axic, line 157: Perhaps you want to make the first push explicit: `push(uint8_t{255})` so that the jump offset `push(2)` remains always valid.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,241 @@
...
+           push("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") + OP_ADD +
+           OP_DUP1 + push(2) + OP_JUMPI;
+}
```
axic, line 160: Perhaps you could use the fact that `0` should remain on top of the stack at the end to check that the loop actually completed.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+/// Generates a benchmark loop with given inner code.
+///
+/// This is improved variant of v1. It has exactly the same instructions and consumes the same
+/// amount of gas, but according to performed benchmarks (see "loop_v1" and "loop_v2") it runs
+/// faster. And we want the lowest possible loop overhead.
+/// The change is to set the loop counter to -255 and check `(counter += 1) != 0`.
+bytecode generate_loop_v2(const bytecode& inner_code)
+{
+    return push("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff01") + OP_JUMPDEST +
```
axic, line 170: Following the comment from above, since here we can not use `uint256_t{}`, maybe a better options is to:
```
auto counter = push("ff...");
return counter + inner_code ... + push(counter.size()) + OP_JUMPI;
```

In both loop generators.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,241 @@
...
+        case InstructionKind::swap:
+        {
+            const auto n = opcode - OP_SWAP1 + 1;
+            // DUP1 SWAP1 SWAP1 ... POP
+            return n * OP_DUP1 + stack_limit * 2 * bytecode{opcode} + n * OP_POP;
+        }
+        case InstructionKind::dup:
+        {
+            const auto n = opcode - OP_DUP1;
+            // DUP1 DUP1 POP DUP1 POP ... POP
```
axic, line 101:
```suggestion
            // DUP1 DUP1 ... DUPn POP DUPn POP ... POP
```

test/bench/synthetic_benchmarks.cpp
axic: I too don't really get why some of these have 2*stack_limit.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,241 @@
...
+        switch (kind)
+        {
+        case InstructionKind::push:
+            // PUSH1 PUSH1 PUSH1 ... POP POP POP ...
+            return stack_limit * push(opcode, {}) + stack_limit * OP_POP;
+        case InstructionKind::dup:
+        {
+            const auto n = opcode - OP_DUP1;
+            // DUP1 DUP1 POP DUP1 POP ... POP
+            return n * OP_DUP1 + (stack_limit - n) * bytecode{opcode} + stack_limit * OP_POP;
```
axic, line 130: Perhaps splitting all these tests into multiple lines could show the grouping better, but then clang-tidy will complain:
```suggestion
            return n * OP_DUP1 +
              (stack_limit - n) * bytecode{opcode} +
              stack_limit * OP_POP;
```

test/bench/synthetic_benchmarks.cpp
chfast: The `2*stack_limit` is your arbitrary number.

gumb0: I see now that it's `stack_limit` for instructions that require POP after each instruction, and `2*stack_limit` for those that don't, so that total count is always `2*stack_limit`.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,269 @@
...
+            const auto n = opcode - OP_SWAP1 + 1;
+            // DUP1 ...  SWAPn SWAPn ...  POP ...
+            // \  n   /                   \  n  /
+            return n * OP_DUP1 +                         // Required stack height for SWAPn.
+                   stack_limit * 2 * bytecode{opcode} +  // Multiple SWAPns.
+                   n * OP_POP;                           // Pop initially duplicated values.
+        }
+        case InstructionCategory::dup:
+        {
+            const auto k = opcode - OP_DUP1;  // For DUPn, k = n-1.
```
gumb0, line 114: As I understand this, DUPs, unlike SWAPs, can use the loop counter that is already on stack, and we want to minimize the overhead of all preparatory stack filling, that's why we use `n - 1`.
I think it's worth to explain in comment.

test/bench/synthetic_benchmarks.cpp
gumb0: 255 looks right.

test/bench/synthetic_benchmarks.cpp
chfast: Not exactly. SWAPn has `n` stack requirement while DUPn only `n-1`.

test/bench/synthetic_benchmarks.cpp
chfast: Because I need to re-run the benchmark anyway, should I change the value of this?

test/bench/synthetic_benchmarks.cpp
chfast: I should get some selfchecking from histogram. Here I prefer to keep it minimal.

test/bench/synthetic_benchmarks.cpp
```diff
@@ -0,0 +1,277 @@
...
+            [&vm = vm](State& state) { execute(state, vm, generate_loop_v2({})); });
+    }
+
+    for (const auto params : params_list)
+    {
+        for (auto& [vm_name, vm] : registered_vms)
+        {
+            RegisterBenchmark(
+                (std::string{vm_name} + "/execute/synth/" + to_string(params)).c_str(),
+                [&vm = vm, params](State& state) { execute(state, vm, generate_code(params)); })
```
gumb0, line 272: This should be possible?
```suggestion
                [&vm = vm, code = generate_code(params)](State& state) { execute(state, vm, code); })
```

chfast: This way we will generate all bytecodes on process startup. Not sure this is good thing, but I will measure this.

gumb0: Why do you think including code generation into the benchmark of executing code can be a good thing?

chfast: This code generation is outside of the benchmark loop so its time is not measured.

With the change the `evmone-bench` startup time in Debug increases from `0.01s` to `0.34s`, so I don't think this is good change.

chfast: The benchmark loop is here: https://github.com/ethereum/evmone/blob/master/test/bench/helpers.hpp#L70-L75.
Everything outside does not contribute to the timings.

gumb0 (approved):

PR #281, opened 2020-12-22
fuzzer: Add baseline interpreter

chfast:


test/fuzzer/fuzzer.cpp
```diff
@@ -309,11 +309,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t data_size) noe
     if (!in)
         return 0;
 
-    auto& ref_host = in.host;
+    auto ref_host = in.host;  // Copy Host.
```
chfast, line 312: More copies because we test 3 interpreters now.

test/fuzzer/fuzzer.cpp
```diff
@@ -371,6 +370,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t data_size) noe
             ASSERT(std::equal(ref_host.recorded_logs.begin(), ref_host.recorded_logs.end(),
                 host.recorded_logs.begin(), host.recorded_logs.end()));
 
+            ASSERT_EQ(ref_host.recorded_blockhashes.size(), host.recorded_blockhashes.size());
```
gumb0, line 373: This is checked by `std::equal` is it just for simpler failure messsage?

gumb0 (approved):

test/fuzzer/fuzzer.cpp
chfast: This is redundant check for easier debugging.

PR #289, opened 2021-02-15
EIP-2929: Gas cost increases for state access opcodes

yperbasis:
Keeping as draft until https://github.com/ethereum/evmc/pull/571 is merged.

### DIFF
--- a/lib/evmone/instruction_traits.hpp
+++ b/lib/evmone/instruction_traits.hpp
@@ -1,12 +1,18 @@
 // evmone: Fast Ethereum Virtual Machine implementation
 // Copyright 2020 The evmone Authors.
 // SPDX-License-Identifier: Apache-2.0
+#pragma once
 
 #include <evmc/instructions.h>
 #include <array>
 
 namespace evmone::instr
 {
+/// EIP-2929 constants (https://eips.ethereum.org/EIPS/eip-2929).
+inline constexpr auto cold_sload_cost = 2100;
+inline constexpr auto cold_account_access_cost = 2600;
+inline constexpr auto warm_storage_read_cost = 100;
+
 /// The EVM instruction traits.
 struct Traits
 {
@@ -333,5 +339,18 @@ constexpr inline std::array<int16_t, 256> gas_costs<EVMC_ISTANBUL> = []() noexce
 }();
 
 template <>
-constexpr inline auto gas_costs<EVMC_BERLIN> = gas_costs<EVMC_ISTANBUL>;
+constexpr inline std::array<int16_t, 256> gas_costs<EVMC_BERLIN> = []() noexcept {
+    auto table = gas_costs<EVMC_ISTANBUL>;
+    table[OP_EXTCODESIZE] = warm_storage_read_cost;
+    table[OP_EXTCODECOPY] = warm_storage_read_cost;
+    table[OP_EXTCODEHASH] = warm_storage_read_cost;
+    table[OP_BALANCE] = warm_storage_read_cost;
+    table[OP_CALL] = warm_storage_read_cost;
+    table[OP_CALLCODE] = warm_storage_read_cost;
+    table[OP_DELEGATECALL] = warm_storage_read_cost;
+    table[OP_STATICCALL] = warm_storage_read_cost;
+    table[OP_SLOAD] = warm_storage_read_cost;
+    return table;
+}();
+
 }  // namespace evmone::instr
--- a/lib/evmone/instruction_traits.hpp
+++ b/lib/evmone/instruction_traits.hpp
@@ -9,9 +9,18 @@
 namespace evmone::instr
 {
 /// EIP-2929 constants (https://eips.ethereum.org/EIPS/eip-2929).
+/// @{
 inline constexpr auto cold_sload_cost = 2100;
 inline constexpr auto cold_account_access_cost = 2600;
 inline constexpr auto warm_storage_read_cost = 100;
+
+/// Additional cold account access cost.
+///
+/// The warm access cost is unconditionally applied for every account access instruction.
+/// If the access turns out to be cold, this cost must be applied additionally.
+inline constexpr auto additional_cold_account_access_cost =
+    cold_account_access_cost - warm_storage_read_cost;
+/// @}
 
 /// The EVM instruction traits.
 struct Traits
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -4,8 +4,8 @@
 #pragma once
 
 #include "execution_state.hpp"
+#include "instruction_traits.hpp"
 #include <ethash/keccak.hpp>
-#include <evmc/instructions.h>
 
 namespace evmone
 {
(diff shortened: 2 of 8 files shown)


lib/evmone/instructions_calls.cpp
```diff
@@ -20,6 +20,13 @@ evmc_status_code call(ExecutionState& state) noexcept
 
     state.stack.push(0);  // Assume failure.
 
+    if (state.rev >= EVMC_BERLIN && state.host.access_account(dst) == EVMC_COLD_ACCESS)
```
chfast, line 23: Minimal improvement here would be to merge `access_account()` and `account_exists()`. I believe the account can have 3 statuses then: hot, cold, non-existing.

BTW, can a non-existing account become hot?

lib/evmone/instructions.hpp
```diff
@@ -279,10 +279,20 @@ inline void address(ExecutionState& state) noexcept
...
+    const auto addr = intx::be::trunc<evmc::address>(x);
+
+    if (state.rev >= EVMC_BERLIN && state.host.access_account(addr) == EVMC_COLD_ACCESS)
+    {
+        // EIP-2929: COLD_ACCOUNT_ACCESS_COST - WARM_STORAGE_READ_COST = 2500
+        if ((state.gas_left -= 2500) < 0)
+            return EVMC_OUT_OF_GAS;
+    }
+
+    x = intx::be::load<uint256>(state.host.get_balance(addr));
```
chfast, line 294: Here and with other methods accessing information about accounts the `get_balance()` can be extended to return both account status and the balance. The gas check should be done after the call, although I'm sure @holiman would be happy with such implementation. But my reasoning that even "access account" makes it hot so it can also return the balance in the same time.

yperbasis: That was my thinking at first with sload, to return both the value and the access status. But then I realised that it complicates the API. I believe that having a separate method specifically to record access in the sense of EIP-2929 leads to a simpler API compared to reporting access status in various methods. The performance overhead shouldn't be too big, just an extra function call.

lib/evmone/instructions_calls.cpp
yperbasis: Not sure that merging `access_account` with `account_exists` is worth it. It might save a function call, but the two are quite different: `account_exists` ([EIP-161](https://eips.ethereum.org/EIPS/eip-161)) is a constant operation that doesn't change anything, while `access_account` modifies `accessed_addresses` substate ([EIP-2929](https://eips.ethereum.org/EIPS/eip-2929)).

lib/evmone/instructions.hpp
```diff
@@ -680,6 +736,13 @@ inline evmc_status_code selfdestruct(ExecutionState& state) noexcept
...
+    if (state.rev >= EVMC_BERLIN && state.host.access_account(beneficiary) == EVMC_COLD_ACCESS)
+    {
+        // EIP-2929: COLD_ACCOUNT_ACCESS_COST = 2600
+        if ((state.gas_left -= 2600) < 0)
+            return EVMC_OUT_OF_GAS;
+    }
+
     if (state.rev >= EVMC_TANGERINE_WHISTLE)
     {
         if (state.rev == EVMC_TANGERINE_WHISTLE || state.host.get_balance(state.msg->destination))
```
holiman, line 748: This looks odd.. "If TangerineWhistle, OR we're sending some money". Why not `&&` ?

gumb0: In Tangerine Whistle, additional cost is applied regardless of value being sent to beneficiary.
Starting with Spurious Dragon it is applied only for non-zero-value selfdestructs.

lib/evmone/instructions_calls.cpp
gumb0: `SELFDESTRUCT` puts non-existing beneficiary into a warm set, the same I thihk with calls, so warm/cold and existing/non-existing seem to be orthogonal.

One could also think that `account_exists` puts non-existing account into a warm set, it's used only in `SELFDESTRUCT` and calls which do it, and it would reflect reality of updating non-existence caches in clients.

chfast:
Consensus tests are going to be updated later in https://github.com/ethereum/evmone/pull/298.

lib/evmone/instruction_traits.hpp
```diff
@@ -1,12 +1,20 @@
...
 // SPDX-License-Identifier: Apache-2.0
+#pragma once
 
 #include <evmc/instructions.h>
 #include <array>
 
 namespace evmone::instr
 {
+/// EIP-2929 constants (https://eips.ethereum.org/EIPS/eip-2929).
+inline constexpr auto cold_sload_cost = 2100;
```
gumb0, line 12: Not sure what we're going to do with these constants if they change in future forks... Mabye just prepend with `berlin_`. But I guess it's ok to leave like this now.

chfast: True. But it would be easy to move the to constexpr array. E.g. `instr::constants[EVMC_BERLIN].cold_sload_cost`.

lib/evmone/instruction_traits.hpp
```diff
@@ -1,12 +1,20 @@
...
+inline constexpr auto cold_account_access_cost = 2600;
+inline constexpr auto warm_storage_read_cost = 100;
+inline constexpr auto additional_cold_account_access_cost =
```
gumb0, line 15: Maybe introduce also `additional_sload_cost` and use it in `sload` for consistency.

test/unittests/evm_eip2929_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+    const auto code =
+        "0x60013f5060023b506003315060f13f5060f23b5060f3315060f23f5060f33b5060f1315032315030315000";
+
+    execute(code);
+    EXPECT_GAS_USED(EVMC_SUCCESS, 8653);
+    EXPECT_EQ(result.output_size, 0);
+
+    const auto& r = host.recorded_account_accesses;
+    ASSERT_EQ(r.size(), 24);
+    EXPECT_EQ(r[0], msg.sender);
```
gumb0, line 28: Maybe could be better-looking with some fancy assert
```
EXPECT_THAT(r, ElementsAre(msg.sender, msg.destination, ...));
```

chfast: This is fine so I don't need to count these in my head. And I don't need to import Google Mock.

lib/evmone/instruction_traits.hpp
chfast: I used the name, but only locally where it is used. Also explained the global definition.

lib/evmone/instructions.hpp
```diff
@@ -561,11 +561,25 @@ inline evmc_status_code mstore8(ExecutionState& state) noexcept
...
 
-inline void sload(ExecutionState& state) noexcept
+inline evmc_status_code sload(ExecutionState& state) noexcept
 {
     auto& x = state.stack.top();
-    x = intx::be::load<uint256>(
-        state.host.get_storage(state.msg->destination, intx::be::store<evmc::bytes32>(x)));
+    const auto key = intx::be::store<evmc::bytes32>(x);
+
+    if (state.rev >= EVMC_BERLIN &&
```
axic, line 602: Is there any measureable performance diff for these revision checks? Should we templetize based on version at some point?

chfast: There are not benchmarks for state access. I could add some, but these are may not be "portable" because depending on the pre-state initialization.

I don't think it is worth duplicating the code with templates for these. However, it may be good to priorities the most recent revision in the code.

lib/evmone/instructions.cpp
```diff
@@ -316,7 +316,7 @@ constexpr std::array<instruction_exec_fn, 256> instruction_implementations = [](
     table[OP_SHL] = op<shl>;
     table[OP_SHR] = op<shr>;
     table[OP_SAR] = op<sar>;
-    table[OP_EXTCODEHASH] = op<extcodehash>;
```
axic, line 319: Why is this removed here?

chfast: Oh yeah, I was wandering the same @yperbasis.

yperbasis: Oh, I simply removed duplicated code; see line 323.

chfast: Ah ok. I will sort these in separate PR then.

axic:
This was merged under #301.

PR #290, opened 2021-02-19
ci: Add Silkworm-driven consensus tests

chfast:


yperbasis (approved):

circle.yml
```diff
@@ -173,6 +231,24 @@ jobs:
...
+    environment:
+      TESTS_REV: v7.0.0
+    executor: linux-gcc-latest
+    steps:
+      - build_silkworm
+      - attach_workspace:
+          at: ~/package
+      - run:
+          name: "Unpack evmone"
+          background: true
```
axic, line 244: So this runs parallel? Does it really take that much time to risk any potential wrong order execution here?

circle.yml
```diff
@@ -15,6 +15,64 @@ executors:
...
+          command: |
+            [ "$SILKWORM_BUILD" = true ] || exit 0
+            cmake --build build --parallel --target consensus
+      - save_cache:
+          name: "Save Silkworm cache"
+          key: *silkworm-cache-key
+          paths:
+            - ~/silkworm/consensus
+
+  download_consensus_tests:
```
axic, line 64: Why not chace this too?

chfast: Cache means you have to download and unpack it, just from different server.

axic: Ah ok, I thought it is some better integration with circleci.

axic (approved):

PR #297, opened 2021-04-07
Upgrade EVMC to 8.0.0-alpha.1

chfast:


chfast:
Silkworm consensus tests are expected to fail here.

PR #303, opened 2021-04-19
EOF validation

chfast:


test/experimental/eof_validation.cpp
```diff
@@ -0,0 +1,420 @@
...
+                if (section_sizes[CODE_SECTION] == 0)
+                    return error_code::code_section_missing;
+                state = State::terminated;
+                break;
+            case DATA_SECTION:
+                if (section_sizes[CODE_SECTION] == 0)
+                    return error_code::code_section_missing;
+                [[fallthrough]];
+            case CODE_SECTION:
+                state = State::section_size;
```
gumb0, line 123: Invalid case of multiple code/data sections is not handled here. Fixed in https://github.com/ethereum/evmone/pull/334/commits/0995f07141bc82204f4f5fd97c5fc84da0ca75f4

PR #305, opened 2021-04-22
Export and benchmark build_jumpdest_map()

chfast:


yperbasis (approved):

lib/evmone/baseline.cpp
```diff
@@ -23,11 +19,13 @@ JumpdestMap build_jumpdest_map(const uint8_t* code, size_t code_size)
         if (op == OP_JUMPDEST)
             m[i] = true;
         else if (op >= OP_PUSH1 && op <= OP_PUSH32)
-            i += static_cast<size_t>(op - OP_PUSH1 + 1);
+            i += op - size_t{OP_PUSH1 - 1};
```
axic, line 22: Wonder why don't we just have some kind of `push_size()` helper if this is used in other places too?

chfast: Is it?

axic: Not sure, is it?

chfast: I have not found other case in lib/evmone.

PR #313, opened 2021-05-04
Provide better execute() C++ API

chfast:
Change the execute API for Advanced interpreter to map the Baseline API.
The new API requires to provide the execution state object (user can
make better decision how to allocate it) and the code analysis (it is
bound to the code itself so user can cache them together).

### DIFF
--- a/lib/evmone/analysis.hpp
+++ b/lib/evmone/analysis.hpp
@@ -55,14 +55,7 @@ struct AdvancedExecutionState : ExecutionState
     /// Pointer to code analysis.
     const AdvancedCodeAnalysis* analysis = nullptr;
 
-    AdvancedExecutionState() = default;
-
-    AdvancedExecutionState(const evmc_message& message, evmc_revision revision,
-        const evmc_host_interface& host_interface, evmc_host_context* host_ctx,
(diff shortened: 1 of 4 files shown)


yperbasis (approved):

lib/evmone/analysis.hpp
```diff
@@ -74,11 +67,11 @@ struct AdvancedExecutionState : ExecutionState
     /// Resets the contents of the execution_state so that it could be reused.
     void reset(const evmc_message& message, evmc_revision revision,
         const evmc_host_interface& host_interface, evmc_host_context* host_ctx,
-        const uint8_t* code_ptr, size_t code_size, const AdvancedCodeAnalysis& a) noexcept
+        const uint8_t* code_ptr, size_t code_size) noexcept
```
gumb0, line 70: Is `reset` needed after this?

chfast: Reset seems better than `state = {}` because in the second case it will also copy/clear the stack data. The stack data can stay dirty.

chfast: Silkworm is using reset() because they maintain a execution state pool.

gumb0 (approved):

PR #314, opened 2021-05-04
Baseline API

chfast:


lib/evmone/evmone.cpp
```diff
@@ -31,7 +31,7 @@ evmc_set_option_result set_option(evmc_vm* vm, char const* name, char const* val
     {
         if (value[0] == '0' && value[1] == '\0')  // O=0
         {
-            vm->execute = evmone::baseline_execute;
+            vm->execute = evmone::baseline::execute;
```
axic, line 34: Nice! Will you also move the other one to the `evmone::advanced` namesapce?

chfast: It was easy esthetic change, but I don't plan to do it more right now.

PR #321, opened 2021-05-12
Add tracing infrastructure to Baseline interpreter

chfast:
With the templated implementation there is no change in the performance. Without a template, the overhead can be up to 10%.

The `Tracer` interface is:
```cpp
class Tracer
{ 
    virtual void on_execution_start(
        evmc_revision rev, const evmc_message& msg, bytes_view code) noexcept = 0;
    virtual void on_instruction_start(uint32_t pc) noexcept = 0;
    virtual void on_execution_end(const evmc_result& result) noexcept = 0;
};
```

It can be extended/modified easily depending on needs of the practical tracer implementations.

### DIFF
--- a/lib/evmone/tracing.hpp
+++ b/lib/evmone/tracing.hpp
@@ -1,0 +1,51 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2021 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <evmc/instructions.h>
+#include <memory>
+#include <string_view>
+
+namespace evmone
+{
+using bytes_view = std::basic_string_view<uint8_t>;
+
+class Tracer
+{
+    friend class VM;  // Has access the the m_next_tracer to traverse the list forward.
+    std::unique_ptr<Tracer> m_next_tracer;
+
+public:
+    virtual ~Tracer() = default;
+
+    void notify_execution_start(
+        evmc_revision rev, const evmc_message& msg, bytes_view code) noexcept
+    {
+        on_execution_start(rev, msg, code);
+        if (m_next_tracer)
+            m_next_tracer->notify_execution_start(rev, msg, code);
+    }
+
+    void notify_execution_end(const evmc_result& result) noexcept  // NOLINT(misc-no-recursion)
+    {
+        on_execution_end(result);
+        if (m_next_tracer)
+            m_next_tracer->notify_execution_end(result);
+    }
+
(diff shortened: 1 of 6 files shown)


lib/evmone/tracing.hpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    void notify_execution_end() noexcept  // NOLINT(misc-no-recursion)
+    {
+        on_execution_end();
+        if (m_next_tracer)
+            m_next_tracer->notify_execution_end();
+    }
+
+    void notify_instruction_start(uint32_t pc) noexcept  // NOLINT(misc-no-recursion)
```
jwasinger, line 36: Ah, so it seems that with `notify_instruction_start` and `on_execution_end`, there wouldn't be a need for `notify_instruction_end`.  I'm curious if there are any awkward edge-cases for an EIP3155-compatible tracer because of this.

lib/evmone/tracing.hpp
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <evmc/instructions.h>
+#include <memory>
+#include <string_view>
+
+namespace evmone
+{
+using bytes_view = std::basic_string_view<uint8_t>;
```
axic, line 12: Don't we have this defined somewhere already?

test/unittests/tracing_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+      : m_baseline_vm{evmc_create_evmone(), {{"O", "0"}}},
+        vm{*static_cast<evmone::VM*>(m_baseline_vm.get_raw_pointer())}
+    {}
+
+    std::string trace(bytes_view code)
+    {
+        evmc::MockedHost host;
+        evmc_message msg{};
+        msg.gas = 1000000;
+        m_baseline_vm.execute(host, EVMC_LONDON, msg, code.data(), code.size());
```
axic, line 35: I think using berlin would be better, since that is well defined.

axic (approved):
Looks good otherwise to me, though I did not do a super strict line-by-line review.

lib/evmone/tracing.hpp
chfast: We can also add `notify_instruction_end()` if needed. The biggest issue seems to be "instruction gas cost".

lib/evmone/tracing.hpp
chfast: Yes, but in some completely different place. Apparently you can alias the same type multiple times.

PR #323, opened 2021-05-13
Histogram tracer

chfast:


test/unittests/tracing_test.cpp
```diff
@@ -95,3 +95,16 @@ TEST_F(tracing, three_tracers)
...
     EXPECT_EQ(trace(dup1(0)), "A0:PUSH1 B0:PUSH1 C0:PUSH1 A2:DUP1 B2:DUP1 C2:DUP1 ");
 }
+
+TEST_F(tracing, histogram)
+{
+    vm.add_tracer(evmone::create_histogram_tracer(trace_stream));
+
+    trace_stream << '\n';
+    EXPECT_EQ(trace(add(0, 0)), R"(
+--- # HISTOGRAM depth=0
```
axic, line 105: Is this valid in CSV?

lib/evmone/tracing.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+        m_code = nullptr;
+    }
+
+public:
+    explicit HistogramTracer(std::ostream& out) noexcept : m_out{out} {}
+};
+}  // namespace
+
+std::unique_ptr<Tracer> create_histogram_tracer(std::ostream& out)
```
axic, line 47: Would add a comment (or on the class) explaining what this does, i.e. outputs a histogram of opcodes in a CSV format.

lib/evmone/tracing.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    void on_instruction_start(uint32_t pc) noexcept override { ++m_counts[m_code[pc]]; }
+
+    void on_execution_end(const evmc_result& /*result*/) noexcept override
+    {
+        const auto names = evmc_get_instruction_names_table(m_rev);
+        m_out << "--- # HISTOGRAM depth=" << m_depth << "\nopcode,count\n";
+        for (size_t i = 0; i < std::size(m_counts); ++i)
+        {
+            if (m_counts[i] != 0)
```
axic, line 35: Should we have thre options?
- do not print zeroes
- one which also prints all the zero opcodes
- only those zero opcodes which are defined for the evm version

axic:
There's missing coverage on the `histogram` evmc option.

test/unittests/tracing_test.cpp
chfast: No. But this is yaml document start mark.

chfast: The CSV can only have optional header line with column names. You cannot have any title line.

The fancier YAML wrapper would be
```yaml
--- # HISTOGRAM
depth: 0
csv: |
  opcode,count
  ADD,1
  PUSH1,2
```

lib/evmone/tracing.cpp
chfast: Some of these options make sense, but I don't think it is worth doing these right now.

lib/evmone/tracing.cpp
```diff
@@ -0,0 +1,59 @@
...
+            {
+                if (const auto name = names[i]; name != nullptr)
+                    m_out << name;
+                else
+                    m_out << "0x" << evmc::hex(static_cast<uint8_t>(i));
+                m_out << ',' << m_counts[i] << '\n';
+            }
+        }
+
+        m_code = nullptr;
```
gumb0, line 47: Maybe zero out `m_counts` and `m_depth` then, too?

chfast: True.

chfast: Well... because we have only single table in tracer we cannot measure internal calls separately. Therefore we should output only at depth=0 (and only then zeroing counters make sense). Moreover, reporting depth is useless.

test/unittests/tracing_test.cpp
axic: So why not output YAML and then whoever wants CSV can do the conversion themselves (or in worst-case by skipping the top 3 lines) ?

lib/evmone/tracing.cpp
```diff
@@ -0,0 +1,64 @@
...
+    uint32_t m_counts[256]{};
+    std::ostream& m_out;
+
+    void on_execution_start(
+        evmc_revision rev, const evmc_message& msg, bytes_view code) noexcept override
+    {
+        m_code = code.data();
+        m_depth = msg.depth;
+        if (m_depth == 0)
+            m_opcode_names = evmc_get_instruction_names_table(rev);
```
gumb0, line 27: Why save it in the class, and not just call `evmc_get_instruction_names_table` in `on_execution_end` ?

chfast: This needs revision. So I need save one of these.

lib/evmone/tracing.cpp
gumb0: Wait, is `on_execution_start` called for each subcall? It changes `m_code`, how would it work after return from subcall then?

chfast: Oh shit, it was suppose to be easy tracer example...

test/unittests/tracing_test.cpp
chfast: Because nobody wants YAML. I used `---` because of no better ideas.

PR #327, opened 2021-05-20
State types extensions for tracing

chfast:
This adds some simple inspection methods to `Stack` and `Memory` types needed for tracing.

### DIFF
--- a/lib/evmone/execution_state.hpp
+++ b/lib/evmone/execution_state.hpp
@@ -44,6 +44,12 @@ struct Stack
     /// Returns the reference to the stack item on given position from the stack top.
     [[nodiscard]] intx::uint256& operator[](int index) noexcept { return *(top_item - index); }
 
+    /// Returns the const reference to the stack item on given position from the stack top.
+    [[nodiscard]] const intx::uint256& operator[](int index) const noexcept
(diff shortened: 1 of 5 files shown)


lib/evmone/execution_state.hpp
```diff
@@ -83,6 +83,7 @@ class Memory
 
     uint8_t& operator[](size_t index) noexcept { return m_memory[index]; }
 
+    [[nodiscard]] const uint8_t* data() const noexcept { return m_memory.data(); }
```
axic, line 86: I wonder if in fact you don't actually want something like `bytes_view substr(size_t start, size_t size) const` here?

axic (approved):

lib/evmone/execution_state.hpp
chfast: There might be uses of such pattern, but tracing is not one of these.

axic: Well your test was doing this, what will be the use case in tracing?

chfast: Dump all memory, e.g. `hex({memory.data(), memory.size()})`.

PR #330, opened 2021-05-21
Upgrade EVMC

chfast:


axic:
The tgz artifact in the release-linux job still only contains `bin/evmone-bench` and nothing else.

chfast:
> The tgz artifact in the release-linux job still only contains `bin/evmone-bench` and nothing else.

This is controlled by CMake `install()`. I believe we don't install EVMC therefore binaries are not included.

chfast:
Included in #333.

PR #331, opened 2021-05-26
cmake: Disable stack protection (-fstack-protector)

chfast:
Disable runtime stack protection (-fstack-protector). It does not
compile well to WebAssembly. ASan can detect the same issues.


axic (approved):

chfast:
I don't see any performance changes.

PR #334, opened 2021-05-28
EOF execution and validation (EIP-3540)

gumb0:
[Make EOF header validation more generic (supporting any sections)](https://github.com/ethereum/evmone/pull/334/commits/b7ef61abaee316c716f805467584c48841fb2a99) commit is an optional refactoring, than will make adding new sections easier in the future.

### DIFF
--- a/lib/evmone/baseline.cpp
+++ b/lib/evmone/baseline.cpp
@@ -4,6 +4,7 @@
 
 #include "baseline.hpp"
 #include "baseline_instruction_table.hpp"
+#include "eof.hpp"
 #include "execution_state.hpp"
 #include "instructions.hpp"
 #include "vm.hpp"
@@ -25,7 +26,9 @@
 
 namespace evmone::baseline
 {
-CodeAnalysis analyze(bytes_view code)
+namespace
+{
+CodeAnalysis analyze_jumpdests(bytes_view code)
 {
     // We need at most 33 bytes of code padding: 32 for possible missing all data bytes of PUSH32
     // at the very end of the code; and one more byte for STOP to guarantee there is a terminating
@@ -54,6 +57,27 @@ CodeAnalysis analyze(bytes_view code)
 
     // TODO: The padded code buffer and jumpdest bitmap can be created with single allocation.
     return CodeAnalysis{std::move(padded_code), std::move(map)};
+}
+
+
+CodeAnalysis analyze_legacy(bytes_view code)
+{
+    return analyze_jumpdests(code);
+}
+
+CodeAnalysis analyze_eof1(bytes_view::const_iterator code, const EOF1Header& header)
+{
+    return analyze_jumpdests({&code[header.code_begin()], header.code_size});
+}
+}  // namespace
+
+CodeAnalysis analyze(evmc_revision rev, bytes_view code)
+{
+    if (rev < EVMC_SHANGHAI || !is_eof_code(code))
+        return analyze_legacy(code);
+
+    const auto eof1_header = read_valid_eof1_header(code.begin());
+    return analyze_eof1(code.begin(), eof1_header);
 }
 
 namespace
@@ -274,7 +298,7 @@ evmc_result execute(evmc_vm* c_vm, const evmc_host_interface* host, evmc_host_co
     evmc_revision rev, const evmc_message* msg, const uint8_t* code, size_t code_size) noexcept
 {
     auto vm = static_cast<VM*>(c_vm);
-    const auto jumpdest_map = analyze({code, code_size});
+    const auto jumpdest_map = analyze(rev, {code, code_size});
     auto state =
         std::make_unique<ExecutionState>(*msg, rev, *host, ctx, bytes_view{code, code_size});
     return execute(*vm, *state, jumpdest_map);
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -13,6 +13,89 @@ namespace
 {
 constexpr uint8_t FORMAT = 0xef;
 constexpr uint8_t MAGIC = 0x00;
+constexpr uint8_t TERMINATOR = 0x00;
+constexpr uint8_t CODE_SECTION = 0x01;
+constexpr uint8_t DATA_SECTION = 0x02;
+
+std::pair<EOF1Header, EOFValidationError> validate_eof_headers(bytes_view container) noexcept
+{
+    enum class State
+    {
+        section_id,
+        section_size,
+        terminated
+    };
+
+    auto state = State::section_id;
+    uint8_t section_id = 0;
+    uint16_t section_sizes[3] = {0, 0, 0};
+    const auto container_end = container.end();
+    auto it = container.begin() + 1 + sizeof(MAGIC) + 1;  // FORMAT + MAGIC + VERSION
+    while (it != container_end && state != State::terminated)
+    {
+        switch (state)
+        {
+        case State::section_id:
+        {
+            section_id = *it++;
+            switch (section_id)
+            {
+            case TERMINATOR:
+                if (section_sizes[CODE_SECTION] == 0)
+                    return {{}, EOFValidationError::code_section_missing};
+                state = State::terminated;
+                break;
+            case DATA_SECTION:
+                if (section_sizes[CODE_SECTION] == 0)
+                    return {{}, EOFValidationError::code_section_missing};
+                if (section_sizes[DATA_SECTION] != 0)
+                    return {{}, EOFValidationError::multiple_data_sections};
+                state = State::section_size;
+                break;
+            case CODE_SECTION:
+                if (section_sizes[CODE_SECTION] != 0)
+                    return {{}, EOFValidationError::multiple_code_sections};
+                state = State::section_size;
+                break;
+            default:
+                return {{}, EOFValidationError::unknown_section_id};
+            }
+            break;
+        }
+        case State::section_size:
+        {
+            const auto size_hi = *it++;
+            if (it == container_end)
+                return {{}, EOFValidationError::incomplete_section_size};
+            const auto size_lo = *it++;
+            const auto section_size = static_cast<uint16_t>((size_hi << 8) | size_lo);
+            if (section_size == 0)
+                return {{}, EOFValidationError::zero_section_size};
+
+            section_sizes[section_id] = section_size;
+            state = State::section_id;
+            break;
+        }
+        case State::terminated:
+            return {{}, EOFValidationError::impossible};
+        }
+    }
+
+    if (state != State::terminated)
+        return {{}, EOFValidationError::section_headers_not_terminated};
+
+    const auto section_bodies_size = section_sizes[CODE_SECTION] + section_sizes[DATA_SECTION];
+    const auto remaining_container_size = container_end - it;
+    if (section_bodies_size != remaining_container_size)
+        return {{}, EOFValidationError::invalid_section_bodies_size};
+
+    return {{section_sizes[0], section_sizes[1]}, EOFValidationError::success};
+}
+
+std::pair<EOF1Header, EOFValidationError> validate_eof1(bytes_view container) noexcept
+{
+    return validate_eof_headers(container);
+}
 }  // namespace
 
 size_t EOF1Header::code_begin() const noexcept
@@ -43,5 +126,29 @@ EOF1Header read_valid_eof1_header(bytes_view::const_iterator code) noexcept
             static_cast<uint16_t>((code[data_size_offset] << 8) | code[data_size_offset + 1]);
     }
     return header;
+}
+
+uint8_t get_eof_version(bytes_view container) noexcept
+{
+    return (container.size() >= 3 && container[0] == FORMAT && container[1] == MAGIC) ?
+               container[2] :
+               0;
+}
+
+EOFValidationError validate_eof(evmc_revision rev, bytes_view container) noexcept
+{
+    if (!is_eof_code(container))
+        return EOFValidationError::invalid_prefix;
+
+    const auto version = get_eof_version(container);
+
+    if (version == 1)
+    {
+        if (rev < EVMC_SHANGHAI)
+            return EOFValidationError::eof_version_unknown;
+        return validate_eof1(container).second;
+    }
+    else
+        return EOFValidationError::eof_version_unknown;
 }
 }  // namespace evmone
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -1,0 +1,47 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2021 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "eof.hpp"
+
+#include <array>
+#include <cassert>
+
+namespace evmone
+{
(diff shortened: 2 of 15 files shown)


lib/evmone/eof.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "eof.hpp"
+
+#include <array>
+
+namespace evmone
+{
+namespace
+{
+constexpr uint8_t FORMAT = 0xef;
+constexpr uint8_t MAGIC[] = {0xa6, 0x1c};
```
axic, line 14: Why not use the same as in geth/tests?

gumb0: I took this from #303 

Yeah we should better change it to have the same thing everywhere.

chfast (commented):
Looks good so far.

test/utils/bytecode.hpp
```diff
@@ -62,6 +62,25 @@ inline bytecode operator*(int n, evmc_opcode op)
     return n * bytecode{op};
 }
 
+inline bytecode eof_header(uint16_t code_size, uint16_t data_size = 0)
```
axic, line 65: Nice how non-complex this is.

lib/evmone/eof.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <stddef.h>
+#include <cstdint>
+
+namespace evmone
+{
+struct EOF1Header
+{
+    int code_size = 0;
+    int data_size = 0;
```
axic, line 14: Do we need to expose `data_size` now? It is not used for anything in the VM. Perhaps just leave a comment that this only cares about code?

gumb0: Currently it's used to find code section offset (i.e. total header size).

lib/evmone/baseline.cpp
```diff
@@ -109,25 +112,25 @@ inline evmc_status_code check_requirements(
...
 
     const auto& instruction_table = get_baseline_instruction_table(state.rev);
 
-    const auto* const code = state.code.data();
-    auto pc = code;
-    while (true)  // Guaranteed to terminate because padded code ends with STOP.
+    // Use padded code for pc.
+    // CODECOPY/CODESIZE will still use original code from state.code
+    const auto* const padded_code = analysis.padded_code.get();
+    const auto* pc = padded_code + header.code_begin();
```
gumb0, line 127: Somewhat hacky we're setting here PC to padded code without changing `state.code`. But `CODECOPY` needs to see original code with data section.

lib/evmone/baseline.cpp
```diff
@@ -398,12 +401,16 @@ evmc_result execute(const VM& vm, ExecutionState& state, const CodeAnalysis& ana
         }
 
         case OP_JUMP:
-            pc = op_jump(state, analysis.jumpdest_map);
+            pc = op_jump(state, analysis.jumpdest_map, analysis.padded_code.get());
+            if (state.status == EVMC_BAD_JUMP_DESTINATION)
```
gumb0, line 405: This is quite ugly, without it if `op_jump` jumps to final `OP_INVALID`, returned status will be overwritten witn `EVMC_INVALID_INSTRUCTION`.

gumb0: It could be instead a check for `pc` value:
```suggestion
            if (pc ==  analysis.padded_code.get() + header.code_end(state.code.size()))
```

test/utils/bytecode.hpp
```diff
@@ -63,6 +63,25 @@ inline bytecode operator*(int n, evmc_opcode op)
...
+{
+    bytecode out{"efcafe0101"};
+    out += bytecode{static_cast<evmc_opcode>(code_size >> 8)} +
```
gumb0, line 69:
```suggestion
    out += bytecode{bytes{code_size >> 8} +
```

lib/evmone/baseline.cpp
gumb0: This was refactored on master.

lib/evmone/execution_state.hpp
```diff
@@ -121,6 +121,7 @@ class Memory
 
 
 /// Generic execution state for generic instructions implementations.
+// NOLINTNEXTLINE(clang-analyzer-optin.performance.Padding)
```
gumb0, line 124:
```
/home/builder/project/lib/evmone/execution_state.hpp:124:8: error: Excessive padding in 'struct evmone::ExecutionState' (32 padding bytes, where 0 is optimal). 
Optimal fields order: 
stack_space, 
gas_left, 
msg, 
output_offset, 
output_size, 
analysis, 
code, 
original_code, 
memory, 
return_data, 
host, 
rev, 
status, 
consider reordering the fields or adding explicit padding members [clang-analyzer-optin.performance.Padding,-warnings-as-errors]
```

lib/evmone/eof.hpp
```diff
@@ -0,0 +1,29 @@
...
+namespace evmone
+{
+using bytes_view = std::basic_string_view<uint8_t>;
+
+struct EOF1Header
+{
+    size_t code_size = 0;
+    size_t data_size = 0;
+
+    [[nodiscard]] EVMC_EXPORT size_t code_begin() const noexcept;
```
gumb0, line 21: @chfast Does this look ok to mark functions in this file `EVMC_EXPORT` in order to call them from unit tests?

chfast: Yes, although you may reconsider you really need to test it if it's trivial function.

gumb0:
Rebased, please review @chfast @axic

lib/evmone/execution_state.hpp
```diff
@@ -165,7 +172,8 @@ class ExecutionState
         msg{&message},
         host{host_interface, host_ctx},
         rev{revision},
-        code{_code}
+        code{_code},
```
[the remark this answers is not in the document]
gumb0: It needs at least the size of it, at least to report correct code to tracing...

But also the field itself is set in baseline to refenence the padded code, `JUMP`s and `PC` use it.

lib/evmone/eof.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <cstdint>
+#include <string>
+
+namespace evmone
+{
+using bytes_view = std::basic_string_view<uint8_t>;
+
+struct EOF1Header
+{
+    size_t code_size = 0;
```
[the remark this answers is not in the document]
gumb0: Me neither, is `uint16_t` better?

test/utils/bytecode.hpp
```diff
@@ -64,6 +69,35 @@ inline bytecode operator*(int n, evmc_opcode op)
...
 
+inline bytes big_endian(uint16_t value)
+{
+    return {static_cast<uint8_t>(value >> 8), static_cast<uint8_t>(value & 0xff)};
+}
+
+inline bytecode eof_header(uint8_t version, uint16_t code_size, uint16_t data_size)
+{
+    bytecode out{"ef00"};
+    out += bytes{version};
```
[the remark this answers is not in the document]
gumb0: This works as `bytecode{bytes{0xEF, 0x00, version}}`, because `bytecode` doesn't have `initializer_list` constructor. And adding such constructor would break `bytecode{int_const}` expressions, which currently translates to PUSH.

test/utils/bytecode.hpp
```diff
@@ -64,6 +69,35 @@ inline bytecode operator*(int n, evmc_opcode op)
...
+}
+
+inline bytecode eof1_header(uint16_t code_size, uint16_t data_size = 0)
+{
+    return eof_header(1, code_size, data_size);
+}
+
+inline bytecode eof1_bytecode(bytecode code, bytecode data = {})
+{
+    return eof1_header(static_cast<uint16_t>(code.size()), static_cast<uint16_t>(data.size())) +
```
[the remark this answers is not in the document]
gumb0: Added.

test/utils/bytecode.hpp
```diff
@@ -64,6 +69,35 @@ inline bytecode operator*(int n, evmc_opcode op)
     return n * bytecode{op};
 }
 
+inline bytes big_endian(uint16_t value)
+{
+    return {static_cast<uint8_t>(value >> 8), static_cast<uint8_t>(value & 0xff)};
```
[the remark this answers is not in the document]
gumb0: Done.

lib/evmone/eof.hpp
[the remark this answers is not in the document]
chfast: Yes, I think `uint16_t` is better.

lib/evmone/eof.cpp
```diff
@@ -0,0 +1,165 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2020 The evmone Authors.
```
axic, line 2: Nit: is this 2020? :)

gumb0: It was created in 2021, so can be changed to that :)

lib/evmone/eof.cpp
```diff
@@ -0,0 +1,47 @@
...
+    return code.size() > 1 && code[0] == FORMAT && code[1] == MAGIC;
+}
+
+EOF1Header read_valid_eof1_header(bytes_view::const_iterator code) noexcept
+{
+    EOF1Header header;
+    const auto code_size_offset = 4;  // FORMAT + MAGIC + VERSION + CODE_SECTION_ID
+    header.code_size =
+        static_cast<uint16_t>((code[code_size_offset] << 8) | code[code_size_offset + 1]);
+    if (code[code_size_offset + 2] == 2)  // is data section present
```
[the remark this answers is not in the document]
gumb0: Yes, this is a function to be used for validated deployed code.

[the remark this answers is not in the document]
axic: Can you add this as a comment on the function?

lib/evmone/eof.cpp
```diff
@@ -13,6 +13,92 @@ namespace
...
+    enum class State
+    {
+        section_id,
+        section_size,
+        terminated
+    };
+
+    auto state = State::section_id;
+    uint8_t section_id = 0;
+    size_t section_sizes[3] = {0, 0, 0};
```
axic: What are the 3 sizes?

lib/evmone/eof.cpp
```diff
@@ -13,6 +13,92 @@ namespace
...
+    const auto code_end = code.end();
+    auto it = code.begin() + sizeof(MAGIC) + 2;  // FORMAT + MAGIC + VERSION
```
axic, line 36: Just for clear mapping between the numbers and comment:

```suggestion
    auto it = code.begin() + 1 + sizeof(MAGIC) + 1; // FORMAT + MAGIC + VERSION
```

lib/evmone/eof.cpp
```diff
@@ -44,4 +130,28 @@ EOF1Header read_valid_eof1_header(bytes_view::const_iterator code) noexcept
     }
     return header;
 }
+
+uint8_t get_eof_version(bytes_view code) noexcept
+{
+    return (code.size() >= 3 && code[0] == FORMAT && code[1] == MAGIC) ? code[2] : 0;
+}
+
+EOFValidationErrror validate_eof(evmc_revision rev, bytes_view code) noexcept
```
axic, line 147: EOFValidationE**rrr**or, how many r's we need? :)

lib/evmone/eof.cpp
```diff
@@ -13,6 +13,92 @@ namespace
...
+        section_size,
+        terminated
+    };
+
+    auto state = State::section_id;
+    uint8_t section_id = 0;
+    size_t section_sizes[3] = {0, 0, 0};
+    const auto code_end = code.end();
+    auto it = code.begin() + sizeof(MAGIC) + 2;  // FORMAT + MAGIC + VERSION
+    while (it != code_end && state != State::terminated)
```
axic, line 37: Why did you decide to implement the generic loop as opposed the short two checks we have in the EIP? Did we ever measure any speed difference?

lib/evmone/eof.cpp
```diff
@@ -13,6 +13,92 @@ namespace
...
+
+            section_sizes[section_id] = section_size;
+            state = State::section_id;
+            break;
+        }
+        case State::terminated:
+            return {{}, EOFValidationErrror::impossible};
+        }
+
+        ++it;
```
axic, line 87: I think this may be better moved to the cases, ie. `size_hi = *it++` ?

lib/evmone/eof.cpp
```diff
@@ -13,6 +13,92 @@ namespace
 {
 constexpr uint8_t FORMAT = 0xef;
 constexpr uint8_t MAGIC = 0x00;
+constexpr uint8_t TERMINATOR = 0x00;
+constexpr uint8_t CODE_SECTION = 0x01;
+constexpr uint8_t DATA_SECTION = 0x02;
+
+std::pair<EOF1Header, EOFValidationErrror> validate_eof_headers(bytes_view code) noexcept
```
axic: Same here, how about calling this container?

lib/evmone/eof.cpp
gumb0: It became handy at one point when implementing RJUMPTABLE with additional jump table section.

That idea is abandoned now, but it can become handy again when we add type section for EOF Functions.

But I don't mind dropping this commit for now.

lib/evmone/eof.cpp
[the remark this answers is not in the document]
gumb0: https://github.com/ethereum/evmone/blob/52912ab9f959e50a96e713025ef7b29191948431/lib/evmone/eof.hpp#L28-L29

lib/evmone/eof.cpp
gumb0: Sizes for each possible value of section_id, 0th one is always 0 for terminator.

In later commit this is replaces with `std::array<uint16_t, MAX_SECTION + 1>`

lib/evmone/eof.cpp
gumb0: Actually it was written originaly by @chfast in https://github.com/ethereum/evmone/pull/303, I decided not to write another one.

lib/evmone/eof.cpp
gumb0: Moved.

lib/evmone/eof.cpp
gumb0: Renamed.

lib/evmone/eof.cpp
```diff
@@ -0,0 +1,160 @@
...
+    if (code[code_size_offset + 2] == 2)  // is data section present
+    {
+        const auto data_size_offset = code_size_offset + 3;
+        header.data_size =
+            static_cast<uint16_t>((code[data_size_offset] << 8) | code[data_size_offset + 1]);
+    }
+    return header;
+}
+
+uint8_t get_eof_version(bytes_view cont) noexcept
```
chfast, line 139:
```suggestion
uint8_t get_eof_version(bytes_view container) noexcept
```

chfast (approved):

lib/evmone/eof.cpp
gumb0: Renamed. (I originally wanted to avoid line wrapping.)

PR #339, opened 2021-06-01
advanced: Prevent invalid memory subscript index

yperbasis:
triggered in MSVC debug build (Visual Studio 2019 version 16.10)

### DIFF
--- a/lib/evmone/execution.cpp
+++ b/lib/evmone/execution.cpp
@@ -20,7 +20,7 @@ evmc_result execute(AdvancedExecutionState& state, const AdvancedCodeAnalysis& a
         (state.status == EVMC_SUCCESS || state.status == EVMC_REVERT) ? state.gas_left : 0;
 
     return evmc::make_result(
-        state.status, gas_left, &state.memory[state.output_offset], state.output_size);
+        state.status, gas_left, state.memory.data() + state.output_offset, state.output_size);
 }
 
 evmc_result execute(evmc_vm* /*unused*/, const evmc_host_interface* host, evmc_host_context* ctx,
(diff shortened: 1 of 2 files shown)


chfast:
> triggered in MSVC debug build

What version? Do you have CI log?

axic:
Do you have a unit test for this?

lib/evmone/execution.cpp
```diff
@@ -19,8 +19,8 @@ evmc_result execute(AdvancedExecutionState& state, const AdvancedCodeAnalysis& a
     const auto gas_left =
         (state.status == EVMC_SUCCESS || state.status == EVMC_REVERT) ? state.gas_left : 0;
 
-    return evmc::make_result(
-        state.status, gas_left, &state.memory[state.output_offset], state.output_size);
+    return evmc::make_result(state.status, gas_left,
+        state.output_size != 0 ? &state.memory[state.output_offset] : nullptr, state.output_size);
```
chfast, line 23: This can be fixed with `state.memory.data() + state.output_offset` because `evmc::make_result()` will not use the pointer if `output_size` is 0.

chfast (changes requested):

yperbasis:
> Do you have a unit test for this?

It was triggered  by running Silkworm's core_test in Debug on Windows. Unfortunately, it didn't show in our Windows CI because it only has Release.

lib/evmone/execution.cpp
yperbasis: Sure, done.

axic:
I suppose something like `revert(100, 0)` would work to trigger, assuming this is the beginning of the code, so memory was not expanded yet.

chfast (approved):

PR #347, opened 2021-06-30
Convert instr::gas_costs to multi-dimentional array

chfast:
Instead of using template specialization `instr::gas_costs<Revision>[op]` convert the instr::gas_costs to multi-dimensional array `instr::gas_costs[Revision][op]`.

The immediate benefit is ability to loop over all revision at compile-time - lower instruction table generation maintenance.


gumb0 (approved):
Looks good.

PR #351, opened 2021-06-30
EOF2 with static relative jumps

gumb0:


lib/evmone/baseline.cpp
```diff
@@ -12,16 +13,21 @@
...
     // it can be noticed that OP_PUSH32 is INT8_MAX (0x7f) therefore
     // static_cast<int8_t>(op) <= OP_PUSH32 is always true and can be skipped.
     static_assert(OP_PUSH32 == std::numeric_limits<int8_t>::max());
 
-    CodeAnalysis::JumpdestMap map(code_size);  // Allocate and init bitmap with zeros.
-    size_t i = 0;
-    while (i < code_size)
+    const auto code_begin = header.code_begin();
+    const auto code_end = header.code_end(code_size);
+    CodeAnalysis::JumpdestMap map(code_end);  // Allocate and init bitmap with zeros.
```
axic, line 28: Why `code_end` not keep `code_size`?

gumb0: A small optimization - there're no jumpdests after code section end, so no point in storing those 0 bits in the map.

gcolvin:
Nice!

gcolvin:
Small suggestion: name these static jumps RJMP and RJMPI to distinguish them from the dynamic JUMP* opcodes.  (Dynamic RJUMP* opcodes are possible, even if undesirable.)

Possible optimization: at validation time you can replace PUSH...RJMP* sequences with equivalent RJMP*... in the bytecode.  These sequences are very common, I think probably the most common way that dynamic jumps are used.  (This will also let you test performance impact of static jumps.)

gumb0:
> Small suggestion: name these static jumps RJMP and RJMPI to distinguish them from the dynamic JUMP* opcodes. (Dynamic RJUMP* opcodes are possible, even if undesirable. And static, absolute JMP* are possible, even if unnecessary.)

Well it wouldn't be really distinguishable when pronouncing it. If there would be desire for such distinction, I would call them something like "RSJUMP*" maybe.

> Possible optimization: at validation time you can replace PUSH...RJMP* sequences with equivalent RJMP*... in the bytecode. These sequences are very common, I think probably the most common way that dynamic jumps are used. (This will also let you test the performance impact of static jumps.)

I guess you mean to replace PUSH... JUMP, i.e. old dynamic jumps? We're thinking rather to make dynamic jumps invalid right away in the same EOF version (it's not part of this PR yet).

gcolvin:
> ... If there would be desire for such distinction, I would call them something like "RSJUMP*" maybe.

It's not that important, especially if JUMP is going away.  In the proposal I've been working on I have JMP and JMPI with static relative addresses, and SUB and RET with static absolute addresses.  And I'm using a variable-length encoding with the high bit always set to avoid the jumpdest-analysis problem in a backwards-compatible way.

> ... I guess you mean to replace PUSH... JUMP, i.e. old dynamic jumps? We're thinking rather to make dynamic jumps invalid right away in the same EOF version (it's not part of this PR yet).

Yes, that's what I meant, thanks.  Short term it's a way get performance numbers.  Longer term it's a way to optimize legacy code.

I don't think you can eliminate dynamic jumps without another mechanism for returning from subroutines.

I've been thinking about ways to constrain dynamic jumps for things like switches and function pointers, along the lines of Wasm tables, but nothing really clear yet.

gumb0:
> I don't think we can eliminate dynamic jumps without another mechanism for returning from subroutines.
> 
> I've been thinking about ways to constrain dynamic jumps for things like switches and function references, along the lines of Wasm tables, but nothing really clear yet.

That's similar to my approach here. `RJUMPTABLE` taking an immediate argument referencing one of the table sections of EOF, and taking a stack argument for the index in the table. Table itself contains relative offsets currently, but maybe better to change to absolute.

We haven't tried to use this on Solidity side yet, but we're hoping this table jump would be enough to implement returns from functions, too.

gcolvin:
> We haven't tried to use this on Solidity side yet, but we're hoping this table jump would be enough to implement returns from functions, too.

I don't quite understand the semantics here, but it sounds to be more complex and less efficient than a traditional return stack with `SUB<immediate offset>`  and `RET`.

gcolvin:
> ... We're thinking rather to make dynamic jumps invalid right away in the same EOF version (it's not part of this PR yet).

I could be wrong, but I still cannot think of any way to make this work without some sort of instruction for returning, which would have to maintain some sort of stack of currently valid PCs to return to.

gcolvin:
Anyway, these look good to me.  Will there be an EIP?

gcolvin:
> Check out also jumps-related [EIP-3690](https://eips.ethereum.org/EIPS/eip-3690) if you haven't seen it yet.

I have, thanks @gumb0.  I'd prefer this proposal plus static subroutines and deprecated (or validated) JUMP and JUMPI.

gumb0:
Current thinking about this PR: `RJUMPTABLE` might be not that useful idea, it's basicallly very similar to what we have now with dynamic jumps constrained by `JUMPDEST` opcodes as destination. `RJUMPTABLE` changes this to store jump destinations in a dedicated EOF section instead of in the code, which is also similar to the idea of [EIP-3690: EOF - JUMPDEST Table](https://eips.ethereum.org/EIPS/eip-3690). But also such a table could be simply calculated and cached by a client at deploy time without storing it in a EOF section.

`RJUMP` and `RJUMPI` still look useful and are being specced out in [EIP-4200](https://eips.ethereum.org/EIPS/eip-4200).

I will probably make a new PR having only `RJUMP` and `RJUMPI` and activating them in EOF1 (in this PR they were in EOF2).

gumb0:
Closing in favor of https://github.com/ethereum/evmone/pull/389

PR #353, opened 2021-07-01
Use common function signature for instruction implementations

chfast:


gumb0:
Shouldn't `OP_GAS` in Baseline now use `gas()` function?

chfast:
> Shouldn't `OP_GAS` in Baseline now use `gas()` function?

Fixed.

lib/evmone/instructions.cpp
```diff
@@ -10,24 +10,10 @@ namespace evmone
 {
 namespace
 {
-template <void InstrFn(Stack&)>
+template <InstrFn Fn>
```
gumb0, line 13: Can't you leave the signature here?
```suggestion
template <template <evmc_status_code Fn(ExecutionState&)>>
```

chfast: I can. This however anticipates another change where the table of instruction implementations is introduced: https://github.com/ethereum/evmone/commit/1fade9cca2024398a9be8bf66016bfc578c72fd4. The table itself is not needed yet, so I don't wanted to commit it early.

Do you think it is better to leave the `InstrFn` for later?

gumb0: Ok to do it later.

gumb0 (approved):

PR #357, opened 2021-07-12
test: Add more tests for terminating instructions

chfast:


test/unittests/evm_test.cpp
```diff
@@ -630,6 +630,35 @@ TEST_P(evm, invalid)
...
+{
+    rev = EVMC_FRONTIER;
+    const auto code = push(0) + OP_SELFDESTRUCT + push(0);
+    execute(3, code);
+    EXPECT_GAS_USED(EVMC_SUCCESS, 3);
+}
+
 TEST_P(evm, keccak256)
 {
     execute("6108006103ff2060005260206000f3");
```
axic, line 664: Could use the `keccak256` helper now?

test/unittests/evm_test.cpp
```diff
@@ -630,6 +630,35 @@ TEST_P(evm, invalid)
...
+{
+    const auto code = push(0) + OP_STOP + OP_POP;
+    execute(3, code);
+    EXPECT_GAS_USED(EVMC_SUCCESS, 3);
+}
+
+TEST_P(evm, inner_return)
+{
+    const auto code = ret(0, 0) + push(0);
+    execute(6, code);
```
axic, line 643: Random question, why don't we make this also to be in the form of:
```
const auto code = ret(0, 0) + push(0;
assert(code == "60006000fd6000");
execute(code;
```

And have some script populate the assertion haha?

chfast:
A lot of unit tests can be improved individually including adding all opcodes to bytecode.hpp. The question remains how much to we want to rehash these.

test/unittests/evm_test.cpp
chfast: In C++20 there is `constexpr std::string` so you can generate the code at compile time... Not sure how to easily script it inject hex strings to the source files.

PR #360, opened 2021-07-13
Use code_address member from evmc_message (EVMC 10.0.0-alpha.1)

yperbasis:
See https://github.com/torquem-ch/silkworm/pull/286 for the original discussion.

Consensus tests pass with https://github.com/torquem-ch/silkworm/pull/288.

### DIFF
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -36,8 +36,8 @@ evmc_status_code call(ExecutionState& state) noexcept
     msg.kind = Kind;
     msg.flags = Static ? uint32_t{EVMC_STATIC} : state.msg->flags;
     msg.depth = state.msg->depth + 1;
-    msg.destination = dst;
-    msg.sender = (Kind == EVMC_DELEGATECALL) ? state.msg->sender : state.msg->destination;
+    msg.recipient = dst;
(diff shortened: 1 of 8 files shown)


lib/evmone/instructions_calls.cpp
```diff
@@ -36,7 +36,8 @@ evmc_status_code call(ExecutionState& state) noexcept
     msg.kind = Kind;
     msg.flags = Static ? uint32_t{EVMC_STATIC} : state.msg->flags;
     msg.depth = state.msg->depth + 1;
-    msg.destination = dst;
+    msg.destination = (Kind == EVMC_CALL) ? dst : state.msg->destination;
```
chfast, line 39: Wait. Shouldn't `CALLCODE` also use `dst`? That was one of the broken part about `CALLCODE`?

yperbasis: Message recipient in case of `CALLCODE` is the same as in case of `DELEGATECALL`. They differ in message sender.

chfast: Ok, this is the address where VALUE goes and where storage is modified... This is still confusing after all these years.

chfast (approved):

yperbasis:
@chfast Looks good to me.

PR #362, opened 2021-07-13
baseline: Implement instructions with a macro

chfast:


gumb0:
What's the benefit of all this?

lib/evmone/instruction_traits.hpp
```diff
@@ -176,13 +176,19 @@ struct Traits
 
     /// The stack height change caused by the instruction execution. Can be negative.
     int8_t stack_height_change = 0;
+
+    /// Does the instruction terminate execution?
+    bool terminator = false;
```
gumb0, line 181:
```suggestion
    bool is_terminator = false;
```

chfast:
Replaced by #382

PR #364, opened 2021-07-15
Optimize BYTE and SAR for code size

chfast:
Rework implementations of `BYTE` and `SAR` to make its code size smaller. The `BYTE` now has branchless implementation. The `SAR` is still not inlined by is 33% smaller by assembly lines.

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -269,20 +269,15 @@ inline evmc_status_code shr(ExecutionState& state) noexcept
 
 inline evmc_status_code sar(ExecutionState& state) noexcept
 {
-    if ((state.stack[1] & (uint256{1} << 255)) == 0)
-        return shr(state);
+    const auto y = state.stack.pop();
+    auto& x = state.stack.top();
 
-    constexpr auto allones = ~uint256{};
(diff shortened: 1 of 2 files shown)


lib/evmone/instructions.hpp
```diff
@@ -269,20 +269,15 @@ inline evmc_status_code shr(ExecutionState& state) noexcept
 
 inline evmc_status_code sar(ExecutionState& state) noexcept
 {
-    if ((state.stack[1] & (uint256{1} << 255)) == 0)
-        return shr(state);
+    const auto y = state.stack.pop();
+    auto& x = state.stack.top();
 
-    constexpr auto allones = ~uint256{};
+    const bool is_neg = static_cast<int64_t>(x[3]) < 0;
```
axic, line 275: Why is this looking at the lowest 64-bit (BE)?

chfast: This is the highest. Words are in LE.

axic: Can you leave a comment in the code?

chfast: Added comment.

PR #369, opened 2021-07-30
bench: Add internal jumpdests section load benchmarks

chfast:


test/internal_benchmarks/eof_jumpdest_section_bench.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        result |= static_cast<T>((static_cast<T>(*pos) & 0x7F) << result_shift);
+        if ((*pos & 0x80) == 0)
+            break;
+    }
+
+    return {result, pos + 1};
+}
+
+static std::vector<bool> load_jumpdests_safe(
```
axic, line 52: Can't you templetize this since the only difference is using `leb128u_decode_safe` vs` leb128u_decode_fast`?

chfast: Can be, though they use different number of arguments.

kmk142789 (approved):

PR #370, opened 2021-08-02
baseline: Fix incorrect exit after invalid jump

chfast:
To handle invalid jump the implementation targets the byte just after
the official code length. For padded code this byte may be uninitialized
push data causing unpredictable behavior. The fix is to also init this
byte to OP_STOP.

The bug was detected in https://github.com/torquem-ch/silkworm/issues/305.

Here two simple unit tests were added which are able to detect the core issue in MSVC/ASan/memcheck builds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
 
+## [0.8.1] — unreleased
+
+### Fixed
+
+- baseline: Fix incorrect exit after invalid jump.
+  [#370](https://github.com/ethereum/evmone/pull/370)
+
+
 ## [0.8.0] — 2021-07-01
 
 ## Added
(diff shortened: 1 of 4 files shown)


yperbasis (approved):
LGTM, but I would still do something like
```
std::fill_n(padded_code.get() + code_size, i + 1 - code_size, uint8_t{OP_STOP});
```
instead.
The overhead is probably low, while it's a guarantee against inconsistent behaviour across different platforms.

chfast:
> LGTM, but I would still do something like
> 
> ```
> std::fill_n(padded_code.get() + code_size, i + 1 - code_size, uint8_t{OP_STOP});
> ```
> 
> instead.
> The overhead is probably low, while it's a guarantee against inconsistent behaviour across different platforms.

Sounds reasonable, but I have slight preference not to be too defensive. The other bytes may only be used by a PUSH instruction and the very end and the values are not observable. If I'm wrong some fuzzing my discover this what is also good.

I also had some issues with `std::fill` because the compiler is not aware that the max length is 33 and it produces big unrolled loop or calls memset() or do both.

We also plan to use fixed-size padding with zeros because it allows single allocation instead of 2 (see TODO comment).

So I'm not sure the `std::fill` is worth right now, but I will follow your recommendation.

yperbasis:
> > LGTM, but I would still do something like
> > ```
> > std::fill_n(padded_code.get() + code_size, i + 1 - code_size, uint8_t{OP_STOP});
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead.
> > The overhead is probably low, while it's a guarantee against inconsistent behaviour across different platforms.
> 
> Sounds reasonable, but I have slight preference not to be too defensive. The other bytes may only be used by a PUSH instruction and the very end and the values are not observable. If I'm wrong some fuzzing my discover this what is also good.
> 
> I also had some issues with `std::fill` because the compiler is not aware that the max length is 33 and it produces big unrolled loop or calls memset() or do both.
> 
> We also plan to use fixed-size padding with zeros because it allows single allocation instead of 2 (see TODO comment).
> 
> So I'm not sure the `std::fill` is worth right now, but I will follow your recommendation.

Sure, fair enough, it's good as is now, especially given that you're going to revisit this code.

PR #371, opened 2021-08-11
Unit test that smashes evmone stack.

bshastry:


bshastry:
Closing because it is not reproducible. May need more work until then.

PR #386, opened 2021-09-17
Upgrade EVMC to 10.0.0-alpha.1

chfast:


chfast:
Should be done together with #360.

PR #390, opened 2021-10-23
Optimize SIGNEXTEND implementation

chfast:
Implement `SIGNEXTEND` by handling individual words of the uint256 value. There only single word must be sign-extended so the core operation must be performed on 64-bit value instead of 256-bit one. Then only the extended words must be reset to the sign bits and the lower words may stay unmodified.

The performance of `SIGNEXTEND` is not very important because not frequently and costly in the same time (gas cost 5). But the implementation is one of two outliers because it is long and not inlined in the interpreter. This PR changes this situation.

```
advanced/execute/micro/signextend/zero_mean                  -0.3513         -0.3513            85            55            85            55
baseline/execute/micro/signextend/zero_mean                  -0.2752         -0.2752           107            77           107            77
advanced/execute/micro/signextend/one_mean                   -0.3792         -0.3792            89            55            89            55
baseline/execute/micro/signextend/one_mean                   -0.2961         -0.2961           109            77           109            77
```


lib/evmone/instructions.hpp
```diff
@@ -173,13 +173,33 @@ inline void signextend(ExecutionState& state) noexcept
...
+
+        const auto sign_byte_offset = sign_byte_index * 8;
+        const auto sign_byte = sign_word >> sign_byte_offset;  // Move sign byte to position 0.
+
+        // Sign-extend the "sign" byte and move it to the right position. Value bits are zeros.
+        const auto sext_byte = static_cast<uint64_t>(int64_t{static_cast<int8_t>(sign_byte)});
+        const auto sext = sext_byte << sign_byte_offset;
+
+        const auto sign_mask = ~uint64_t{0} << sign_byte_offset;
+        const auto value = sign_word & ~sign_mask;  // Reset extended bytes.
```
gumb0, line 192: maybe this small simplification possible?
```suggestion
        const auto sign_mask = (uint64_t{1} << sign_byte_offset) - 1;
        const auto value = sign_word & sign_mask;  // Reset extended bytes.
```

lib/evmone/instructions.hpp
```diff
@@ -173,13 +173,33 @@ inline void signextend(ExecutionState& state) noexcept
...
-        auto sign_bit = static_cast<int>(ext) * 8 + 7;
-        auto sign_mask = uint256{1} << sign_bit;
-        auto value_mask = sign_mask - 1;
-        auto is_neg = (x & sign_mask) != 0;
-        x = is_neg ? x | ~value_mask : x & value_mask;
+        const auto e = ext[0];  // uint256 -> uint64.
+        const auto sign_word_index =
+            static_cast<size_t>(e / sizeof(e));      // Index of the word with the sign bit.
+        const auto sign_byte_index = e % sizeof(e);  // Index of the sign byte in the sign word.
+        const auto sign_word = x[sign_word_index];
```
gumb0, line 182: also random thought, no idea how compilers would react:
```suggestion
        auto& sign_word = x[sign_word_index];
```
then remove `sext_word` variable and directly assign
```
sign_word = sext | value;
```

lib/evmone/instructions.hpp
chfast: This was the original form, but I changed this because the compiler is converting back to `x & ~(-1 << s)` (see pure -O3 version).

Things get fancy with BMI and BMI2 extensions. In BMI there is `andn` instruction (x & ~y) (GCC uses it) and `bextr` (clang).
In BMI2 these two lines are single `bzhi` instruction.

https://godbolt.org/z/K8GE9b5bW

lib/evmone/instructions.hpp
chfast: I applied this change, thanks. The output assembly is the same just the order of instructions changes a bit.

test/unittests/evm_test.cpp
```diff
@@ -478,32 +478,61 @@ TEST_P(evm, signextend)
...
+
+    const auto code = bytecode{} + calldataload(0) + calldataload(32) + OP_SIGNEXTEND + ret_top();
+
+    for (int b = 0; b <= 0xff; ++b)
+    {
+        uint8_t input[64]{};
+
+        auto g = b;
+        for (size_t i = 0; i < 32; ++i)
+            input[i] = static_cast<uint8_t>(g++);  // Generate SIGNEXTEND base argument.
```
gumb0, line 524: As I understand this generates inputs like (in first 32 bytes)
```
000102...1f
010203...20
...
ff0001...1e
```
Is there some idea behind this or is it kind of random?

gumb0 (approved):

test/unittests/evm_test.cpp
chfast: Kind of random. I wanted the value bytes to be all different so there is less chance something is shifted unnoticed.

chfast (commented):
Thanks for the review.

PR #395, opened 2021-11-15
Change benchmark suite organization and loading rules

chfast:
The files with the runtime code have been renamed `*.evm` to `*.bin-runtime` (extension used by solidity). Only files with this extensions are considered when loading benchmark cases. Previously all files were considered what prevented having README, contract sources or any other files inside the suite directory.

Furthermore, now we only execute code which has matching inputs file (`*.inputs`). Otherwise, only code analysis is benchmarked. So we also added 3 new inputs files with "empty" case to execute cases we care about in this context.


gumb0:
> Furthermore, now we only execute code which has matching inputs file (`*.inputs`). Otherwise, only code analysis is benchmarked.

Hm this seems a bit implicit convention, maybe better to have some special suffix for analysis-only benchmars, like `*.analysis-inputs` ? (and still allow to have no inputs, doing full execution benchmark without them)

chfast:
> > Furthermore, now we only execute code which has matching inputs file (`*.inputs`). Otherwise, only code analysis is benchmarked.
> 
> Hm this seems a bit implicit convention, maybe better to have some special suffix for analysis-only benchmars, like `*.analysis-inputs` ? (and still allow to have no inputs, doing full execution benchmark without them)

No sure we understand each other. The code analysis is always benchmarked for every code file. No change here. Just some micro benchmark cases were added only to check analysis performance and it does not make sense to execute them because they fail on first instruction anyway. The change here is that `*.inputs` is required now for execution (must contain at least one input case).

gumb0:
> No sure we understand each other. The code analysis is always benchmarked for every code file. No change here. Just some micro benchmark cases were added only to check analysis performance and it does not make sense to execute them because they fail on first instruction anyway. The change here is that `*.inputs` is required now for execution (must contain at least one input case).

Ok I was a bit confused, but still the suggestion is: having `*.inputs` executes, having empty (or whatever) `*.skip-execution` does only analysis, having none of these still executes.

But I don't insist it's much better, fine to leave it like now.

gumb0 (approved):
Fine to leave it only analysing in case there's no inputs file.

PR #396, opened 2021-11-15
evm-bench.py script

chfast:
Adds `evm-bench.py` script to help benchmarking external EVM implementations.


fallingicarus (approved):

fallingicarus (commented):
py script

PR #401, opened 2022-01-03
baseline: Better opcode to implementation map

chfast:
Replace the op2fn map based on name matching (requires macro)
with the `template <> constexpr` pointing to the functions with
instruction implementations.


lib/evmone/baseline.cpp
```diff
@@ -72,10 +72,10 @@ inline evmc_status_code check_requirements(
 
 
 /// Implementation of a generic instruction "case".
-#define DISPATCH_CASE(OPCODE)                                          \
-    case OPCODE:                                                       \
-        if (code_it = invoke(op2fn::OPCODE, state, code_it); !code_it) \
-            goto exit;                                                 \
+#define DISPATCH_CASE(OPCODE)                                   \
+    case OPCODE:                                                \
+        if (code_it = invoke<OPCODE>(state, code_it); !code_it) \
```
gumb0, line 77: Why is helper required, can't it be
```suggestion
        if (code_it = invoke(instr::impl<OPCODE>, state, code_it); !code_it) \
```

chfast: It can be. This comes from the way I will use it later.

gumb0 (approved):

PR #407, opened 2022-01-11
Rename structs / unions to CamelCase

chfast:


test/.clang-tidy
```diff
@@ -0,0 +1,4 @@
+InheritParentConfig: true
+CheckOptions:
+  - key: readability-identifier-naming.ClassCase
+    value: aNy_CasE
```
gumb0, line 4: Why any case here?

gumb0 (approved):

test/.clang-tidy
chfast: There are some test records which I don't want to touch right now.
```cpp
struct memory_access_opcode
{
    evmc_opcode opcode;
    int memory_index_arg;
    int memory_size_arg;
};
```

PR #408, opened 2022-01-12
Separate and clean-up Advanced implementation

chfast:


yperbasis (approved):

gumb0:
Is max_code_limit removed because it is not applied to initcode? Is it addressed in some later PRs or should `static_assert`s be replaced with some `assert`s for overflow of `BlockInfo` fields?

gumb0:
Another suggestion not yet looking at later PRs: renaming
`AdvancedExecutionState` -> `advanced::ExecutionState`
`AdvancedCodeAnalysis` - > `advanced::CodeAnalysis`

chfast:
> Is max_code_limit removed because it is not applied to initcode? Is it addressed in some later PRs or should `static_assert`s be replaced with some `assert`s for overflow of `BlockInfo` fields?

This is already handled by analysis by saturating to the max type value. https://github.com/ethereum/evmone/blob/master/lib/evmone/analysis.cpp#L33-L40.

Here all these assumptions are wrong (and we had related bugs and unit tests) by assuming 0x6000 is the max possible code size (initcode may be much bigger).

chfast:
> Another suggestion not yet looking at later PRs: renaming `AdvancedExecutionState` -> `advanced::ExecutionState` `AdvancedCodeAnalysis` - > `advanced::CodeAnalysis`

I was thinking about it, but not sure this is very good pattern to have the same type names differ only by the namespace.

gumb0:
> > Another suggestion not yet looking at later PRs: renaming `AdvancedExecutionState` -> `advanced::ExecutionState` `AdvancedCodeAnalysis` - > `advanced::CodeAnalysis`
> 
> I was thinking about it, but not sure this is very good pattern to have the same type names differ only by the namespace.

I think this is the exact purpose of  namespaces - to disambiguate the same names in different components. But up to you.

chfast:
I noticed the whole `limits.hpp` can be removed. The commit now has explanation for the change.

gumb0 (approved):

chfast:
> > > Another suggestion not yet looking at later PRs: renaming `AdvancedExecutionState` -> `advanced::ExecutionState` `AdvancedCodeAnalysis` - > `advanced::CodeAnalysis`
> > 
> > 
> > I was thinking about it, but not sure this is very good pattern to have the same type names differ only by the namespace.
> 
> I think this is the exact purpose of namespaces - to disambiguate the same names in different components. But up to you.

We can try this is separate PR. I partly did it and this causes some confusion because `evmone::advanced::ExecutionState` inherit `evmone::ExecutionState` and in many places in `advanced` you must explicitly use `evmone::ExecutionState` is not obvious for the reader.

chfast:
> > 
> 
> Ok, I haven't seen that I think. Why would it silently clamp instead of e.g. `assert` or other failure?

Saturation is fine for stack checks. It does not matter if a basic block increases stack by 32767 or 42000.

For the gas check I don't remember exactly. I think the `uint32_t` can be broken but this is not practical. One line of protection is to pass over 4GB of gas to execution. If you are able to construct a code that breaks it we can reconsider.

PR #410, opened 2022-01-12
advanced: Reuse more generic instructions

chfast:
Advanced can easily wrap also generic instruction implementations with `StopToken`.

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -841,8 +841,9 @@ inline StopToken return_impl(ExecutionState& state) noexcept
     if (!check_memory(state, offset, size))
         return {EVMC_OUT_OF_GAS};
 
-    state.output_offset = static_cast<size_t>(offset);  // Can be garbage if size is 0.
     state.output_size = static_cast<size_t>(size);
+    if (state.output_size != 0)
+        state.output_offset = static_cast<size_t>(offset);
     return {StatusCode};
 }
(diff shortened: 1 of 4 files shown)


chfast:
Here we have a bug to be fixed and maybe more unit tests are needed.

lib/evmone/instructions.hpp
```diff
@@ -841,8 +841,9 @@ inline StopToken return_impl(ExecutionState& state) noexcept
     if (!check_memory(state, offset, size))
         return {EVMC_OUT_OF_GAS};
 
-    state.output_offset = static_cast<size_t>(offset);  // Can be garbage if size is 0.
     state.output_size = static_cast<size_t>(size);
+    if (state.output_size != 0)
+        state.output_offset = static_cast<size_t>(offset);  // Can be garbage if size is 0.
```
gumb0, line 846:
```suggestion
        state.output_offset = static_cast<size_t>(offset);
```

gumb0:
> Here we have a bug to be fixed and maybe more unit tests are needed.

what bug?

chfast:
> > Here we have a bug to be fixed and maybe more unit tests are needed.
> 
> what bug?

Already fixed by the change to `output_offset`.

lib/evmone/instructions.hpp
chfast: The comment is still true though :)

gumb0: But it's not executed if size is 0

gumb0 (approved):

PR #411, opened 2022-01-12
Use opcode for call/create instruction templates

chfast:
This makes better interface for VM implementations (the evmc_call_kind
is hidden) and the instruction implementations are not more complex.

Rename and alias {call,create,return} -> {*}_impl

To simplify opcode name mapping, rename templates implementing groups of
instruction with _impl suffix and provide aliases matching opcode names
(e.g. call, create2, revert).


yperbasis (approved):

lib/evmone/instructions_calls.cpp
```diff
@@ -50,7 +55,7 @@ evmc_status_code call(ExecutionState& state) noexcept
 
     auto cost = has_value ? 9000 : 0;
 
-    if constexpr (Kind == EVMC_CALL)
+    if constexpr (Op == OP_CALL)
```
gumb0, line 58: This looks suspicious, because previously this was CALL or STATICCALL.

I guess for static call conditions inside are always false: `has_value` is false and STATICALL was added after Spurious Dragon...

Maybe makes sense to add some assertions here for these assumptions?

chfast: Yes. Previously I missed this and there was a bug related to the other check like this. But here this seems correct. I will add asserts. I don't want to change the structure of the implementation but also redundantly adding `Op == OP_STATICCALL` felt wrong.

chfast: I can add even static asserts.
```patch
Index: lib/evmone/instructions_calls.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/evmone/instructions_calls.cpp b/lib/evmone/instructions_calls.cpp
--- a/lib/evmone/instructions_calls.cpp	(revision 078fe563dffa80e2c298080a8202f4e87434d1a0)
+++ b/lib/evmone/instructions_calls.cpp	(date 1642011381658)
@@ -12,9 +12,11 @@
     static_assert(
         Op == OP_CALL || Op == OP_CALLCODE || Op == OP_DELEGATECALL || Op == OP_STATICCALL);
 
+    constexpr auto has_value_arg = (Op == OP_CALL || Op == OP_CALLCODE);
+
     const auto gas = state.stack.pop();
     const auto dst = intx::be::trunc<evmc::address>(state.stack.pop());
-    const auto value = (Op == OP_STATICCALL || Op == OP_DELEGATECALL) ? 0 : state.stack.pop();
+    const auto value = has_value_arg ? state.stack.pop() : 0;
     const auto has_value = value != 0;
     const auto input_offset = state.stack.pop();
     const auto input_size = state.stack.pop();
@@ -55,6 +57,9 @@
 
     auto cost = has_value ? 9000 : 0;
 
+    // STATICCALL was introduces after Spurious Dragon and never has value.
+    static_assert(instr::traits[OP_STATICCALL].since >= EVMC_SPURIOUS_DRAGON);
+    static_assert(Op != OP_STATICCALL || !has_value_arg);
     if constexpr (Op == OP_CALL)
     {
         if (has_value && state.msg->flags & EVMC_STATIC)
```

But I think these suggest some kind of spec optimization. The core of the problem is that the previous check was too broad — `STATICCALL` was never meant to be included.

First check — `STATICCALL` does not violate static mode.

Second check — this refers to for `CALL`
image
but for `STATICCALL`
image

gumb0: Yeah makes sense :+1:

test/unittests/evm_calls_test.cpp
```diff
@@ -587,6 +587,63 @@ TEST_P(evm, call_oog_after_depth_check)
...
+    for (auto op : {OP_CALL, OP_CALLCODE, OP_DELEGATECALL, OP_STATICCALL})
+    {
+        const auto has_value_arg = (op == OP_CALL || op == OP_CALLCODE);
+        const uint64_t value = has_value_arg ? passed_value : 0;
+        const auto value_cost = has_value_arg ? 9000 : 0;
+        const auto expected_value = (op == OP_CALL || op == OP_CALLCODE) ? passed_value :
+                                    (op == OP_DELEGATECALL)              ? origin_value :
+                                                                           0;
+
+        const auto code = 4 * push(0) + push(value) + push(recipient) + push(0) + op;
```
gumb0, line 637: Hm this was a bit confusing, STATICCALL doesn't have `value` argument, right?
This just pushes more zeroes than STATICCALL needs, as I understand, and `value` is 0, too...

I guess it's fine for a test, but maybe some clarifying comment would be nice.

gumb0: Maybe I would for clarity do it with some `if`s constructing codes and expectations differently instead of `has_value_arg`, `value_cost`...

test/unittests/evm_calls_test.cpp
```diff
@@ -587,6 +587,63 @@ TEST_P(evm, call_oog_after_depth_check)
...
+    constexpr auto executor = 0xee000000000000000000000000000000000000ee_address;
+    constexpr auto recipient = 0x4400000000000000000000000000000000000044_address;
+
+    constexpr auto passed_value = 3;
+    constexpr auto origin_value = 8;
+
+    msg.sender = origin;
+    msg.recipient = executor;
+    msg.value.bytes[31] = origin_value;
+    host.accounts[executor].set_balance(3);
```
gumb0, line 625:
```suggestion
    host.accounts[executor].set_balance(passed_value);
```

gumb0 (approved):
Overall looks good, I like the simplification of call implementation.

test/unittests/evm_calls_test.cpp
chfast: I did small changes which make it better although may still not be perfect.

PR #412, opened 2022-01-12
Introduce opcode->identifier mapping X-Macro

chfast:
Use [X-Macro](https://en.wikipedia.org/wiki/X_Macro) (here having a pair of values) where a list of all opcodes is needed.


lib/evmone/advanced_instructions.cpp
```diff
@@ -161,66 +170,12 @@ const Instruction* opx_beginblock(const Instruction* instr, AdvancedExecutionSta
...
-    table[OP_MLOAD] = op<mload>;
-    table[OP_MSTORE] = op<mstore>;
-    table[OP_MSTORE8] = op<mstore8>;
-    table[OP_SLOAD] = op<sload>;
+    // Init table with wrapped generic implementations.
+#define X(OPCODE, IDENTIFIER) table[OPCODE] = op<IDENTIFIER>;
+    MAP_OPCODE_TO_IDENTIFIER
+#undef X
+
+    // Overwrite with Advanced-specific implementations.
```
yperbasis, line 178: `table[OP_MSIZE] = op<msize>` can be removed as well

lib/evmone/advanced_instructions.cpp
```diff
@@ -161,66 +170,12 @@ const Instruction* opx_beginblock(const Instruction* instr, AdvancedExecutionSta
...
     table[OP_SSTORE] = op_sstore;
```
yperbasis, line 179: Why is sstore overridden for the advanced interpreter but not sload?

chfast: All instructions that need access to gas counter are replaced. This information is not directly available in the Advanced so we need to produce it for them. In particular `SSTORE` needs it if `gas < stipend`.

lib/evmone/advanced_instructions.cpp
chfast: Yes, done.

yperbasis (approved):

lib/evmone/advanced_instructions.cpp
chfast: No. I.e. many instruction are doing additional gas checks, but they don't  need to know exact value.

lib/evmone/advanced_instructions.cpp
```diff
@@ -8,8 +8,16 @@
 
 namespace evmone::advanced
 {
+/// Fake wrap for generic instruction implementations accessing current code location.
+/// This is to make any op<...> compile, but pointers must be replaced with Advanced-specific
+/// implementation. Definition not provided.
+template <code_iterator InstrFn(ExecutionState&, code_iterator)>
+const Instruction* op(const Instruction* /*instr*/, AdvancedExecutionState& state) noexcept;
```
gumb0, line 15: Can this be `= 0` maybe?

gumb0 (approved):
Well some of the C++ feel is lost to this :cry: 

But it's nice that in advanced it's now more clearly visible which implementations are advanced-specific.

lib/evmone/advanced_instructions.cpp
chfast: Do you mean `= delete`? I don't think so. The trick here is that this is actually used during compilation but later overwritten by Advanced's variants so linker does not complain. So this has to be external function definition (I even needed to move it out of anonymous namespace).

chfast:
> Well some of the C++ feel is lost to this

This is really helpful in prototyping and also removes long blocks of 100+ lines of uninteresting code. This can be removed in future if instruction implementation signature stabilizes.

PR #416, opened 2022-01-14
Introduce instruction "core" implementations

chfast:
Introduce instruction "core" implementations which differ by not
adjusting EVM stack height. Also provide wrappers (instr::impl<Op>)
to support previous instruction functions.

Use "core" instructions in Baseline.

### DIFF
--- a/lib/evmone/baseline.cpp
+++ b/lib/evmone/baseline.cpp
@@ -116,17 +116,24 @@ inline evmc_status_code check_requirements(
 /// Helpers for invoking instruction implementations of different signatures.
 /// @{
 [[gnu::always_inline]] inline code_iterator invoke(
-    void (*instr_fn)(ExecutionState&) noexcept, ExecutionState& state, code_iterator pos) noexcept
+    void (*instr_fn)(StackTop) noexcept, ExecutionState& state, code_iterator pos) noexcept
 {
-    instr_fn(state);
+    instr_fn(state.stack.top_item);
     return pos + 1;
 }
 
 [[gnu::always_inline]] inline code_iterator invoke(
-    evmc_status_code (*instr_fn)(ExecutionState&) noexcept, ExecutionState& state,
+    StopToken (*instr_fn)() noexcept, ExecutionState& state, code_iterator /*pos*/) noexcept
+{
+    state.status = instr_fn().status;
+    return nullptr;
+}
+
+[[gnu::always_inline]] inline code_iterator invoke(
+    evmc_status_code (*instr_fn)(StackTop, ExecutionState&) noexcept, ExecutionState& state,
     code_iterator pos) noexcept
 {
-    if (const auto status = instr_fn(state); status != EVMC_SUCCESS)
+    if (const auto status = instr_fn(state.stack.top_item, state); status != EVMC_SUCCESS)
     {
         state.status = status;
         return nullptr;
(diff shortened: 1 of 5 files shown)


lib/evmone/instructions.hpp
```diff
@@ -91,75 +91,82 @@ inline bool check_memory(ExecutionState& state, const uint256& offset, const uin
...
+    stack[-1] = stack[0] * stack[-1];
 }
 
-inline void div(ExecutionState& state) noexcept
+inline void div(intx::uint256* stack) noexcept
 {
-    auto& v = state.stack[1];
-    v = v != 0 ? state.stack[0] / v : 0;
-    state.stack.pop();
+    stack[-1] = stack[-1] != 0 ? stack[0] / stack[-1] : 0;
```
gumb0, line 133: Maybe better to leave it with `auto& v` for consistency with functions just below.

lib/evmone/instructions.hpp
```diff
@@ -205,65 +212,60 @@ inline void signextend(ExecutionState& state) noexcept
...
-    state.stack[0] = x < state.stack[0];
+    stack[-1] = stack[0] < stack[-1];
 }
 
-inline void gt(ExecutionState& state) noexcept
+inline void gt(intx::uint256* stack) noexcept
 {
-    const auto x = state.stack.pop();
-    state.stack[0] = state.stack[0] < x;  // Arguments are swapped and < is used.
+    stack[-1] = stack[-1] < stack[0];  // Arguments are swapped and < is used.
```
gumb0, line 222: Why wouldn't this use `lt(stack[-1], stack[0])` like `sgt` below?

lib/evmone/instructions.hpp
```diff
@@ -91,75 +91,82 @@ inline bool check_memory(ExecutionState& state, const uint256& offset, const uin
...
+template <evmc_status_code Status>
+inline StopToken stop_impl() noexcept
+{
+    return {Status};
 }
+inline constexpr auto stop = stop_impl<EVMC_SUCCESS>;
+inline constexpr auto invalid = stop_impl<EVMC_INVALID_INSTRUCTION>;
 
-inline void add(ExecutionState& state) noexcept
+inline void add(intx::uint256* stack) noexcept
```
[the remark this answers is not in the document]
chfast: I previously had `StackController` which kept stack pointer and stack height and had the same API as `Stack`. Maybe it will be back in some place, but I'm not sure yet.

The `StackView` seems good idea, but mostly for keeping old code unchanged. It would work similar to `reverse_iterator`, but I'm not sure if keeping it simple is not better.

lib/evmone/instructions.hpp
```diff
@@ -815,8 +803,9 @@ inline evmc_status_code log(ExecutionState& state) noexcept
         return EVMC_OUT_OF_GAS;
 
     std::array<evmc::bytes32, NumTopics> topics;  // NOLINT(cppcoreguidelines-pro-type-member-init)
+    auto topic_arg = &stack[-2];
```
gumb0, line 806:
```suggestion
    auto* topic_arg = &stack[-2];
```

lib/evmone/baseline.cpp
```diff
@@ -116,36 +116,50 @@ inline evmc_status_code check_requirements(
 /// Helpers for invoking instruction implementations of different signatures.
 /// @{
 [[gnu::always_inline]] inline code_iterator invoke(
-    void (*instr_fn)(ExecutionState&) noexcept, ExecutionState& state, code_iterator pos) noexcept
+    void (*instr_fn)(uint256*), ExecutionState& state, code_iterator pos) noexcept
```
gumb0, line 119: Why does this become not `noexcept`?

chfast: Looks like a mistake.

chfast:
> Is there any need in `Stack` class after this?

Not really, but Advanced is still using it. I plan to remove it.

lib/evmone/instructions.hpp
```diff
@@ -91,75 +108,87 @@ inline bool check_memory(ExecutionState& state, const uint256& offset, const uin
...
     m = m != 0 ? intx::addmod(x, y, m) : 0;
 }
 
-inline void mulmod(ExecutionState& state) noexcept
+inline void mulmod(StackTop stack) noexcept
 {
-    const auto x = state.stack.pop();
-    const auto y = state.stack.pop();
-    auto& m = state.stack.top();
+    const auto& x = stack[0];
```
chfast, line 182: I changed only this implementation because this gives slightly better assembly. All other changes make no difference except to symbol names.

gumb0:
Looks good. There's still some UBSan error and MSVC warning.

PR #426, opened 2022-01-27
[WIP] Do not init stack space to zero

chfast:
This proposes not to init stack space to zero. In theory, EVM should never read a stack item which has not been initialized previously with push or any other instruction. In practice, this change breaks at least Advanced. Problems are reported by valgrind and ASan e.g. in "dup_stack_overflow" test.


lib/evmone/execution_state.hpp
```diff
@@ -40,16 +45,19 @@ struct Stack
...
+    [[nodiscard]] const uint256* storage() const noexcept { return reinterpret_cast<const uint256*>(m_storage); }
+    uint256* storage() noexcept { return reinterpret_cast<uint256*>(m_storage); }
 
     /// Returns the pointer to below the stack storage.
-    [[nodiscard, clang::no_sanitize("bounds")]] uint256* bottom() noexcept { return storage - 1; }
+    [[nodiscard, clang::no_sanitize("bounds")]] uint256* bottom() noexcept { return storage() - 1; }
 
     /// Default constructor. Stack is empty.
-    Stack() noexcept : top_item{bottom()} {}
+    Stack() noexcept : top_item{bottom()}, m_storage() {}
```
chfast, line 57: I believe this will perform value-initialization and will init aligned_storage with zeros. And this is what we want to avoid.
You need to disable the warning with `// NOLINT(...)`.

lib/evmone/execution_state.hpp
```diff
@@ -40,16 +45,19 @@ struct Stack
 
     /// The storage allocated for maximum possible number of items.
     /// Items are aligned to 256 bits for better packing in cache lines.
-    alignas(sizeof(uint256)) uint256 storage[limit];
+    std::aligned_storage_t<sizeof(uint256), sizeof(uint256)> m_storage[limit];
+
+    [[nodiscard]] const uint256* storage() const noexcept { return reinterpret_cast<const uint256*>(m_storage); }
+    uint256* storage() noexcept { return reinterpret_cast<uint256*>(m_storage); }
```
chfast, line 51: Mark this `[[nodiscard]]` too.

PR #430, opened 2022-02-07
baseline: Drop [[always_inline]] from invoke()

chfast:
The [[gnu::always_inline]] attribute on the invoke() does not affect
optimized builds (the function is still inlined). But it saves stack
space in debug builds.
In particular the main baseline::execute() function in debug build:
- use 13872 bytes of stack space with [[gnu::always_inline]],
- 2432 without [[gnu::always_inline]],
- 608 with additional -Og optimization level
  (not enabled in this change).

### DIFF
--- a/lib/evmone/CMakeLists.txt
+++ b/lib/evmone/CMakeLists.txt
@@ -33,7 +33,11 @@ target_include_directories(evmone PUBLIC
     $<BUILD_INTERFACE:${include_dir}>$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 if(CABLE_COMPILER_GNULIKE)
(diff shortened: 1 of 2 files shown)


chfast:
I used `-fstack-usage` and `-Wstack-usage=` GCC flags to track these values.
There is also an option to record stack sizes in separate ELF section (can be later read with tools like `readelf`) but I have not tried this in practice.

yperbasis (approved):
This fixes the stack overflow in Silkworm in debug, thank you!

chfast:
> This fixes the stack overflow in Silkworm in debug, thank you!

Yes. Now the execute() function requires 2.5M of stack space in the worst case. Also the call instruction require additional 0.8M (also in release build). We can try to improve the later by moving call instruction checks to separate function.

PR #436, opened 2022-02-14
Refactor execution state

chfast:
This mostly moved Advanced-only parts of execution state and instruction to Advanced's files.

### DIFF
--- a/lib/evmone/advanced_analysis.hpp
+++ b/lib/evmone/advanced_analysis.hpp
@@ -35,7 +35,7 @@ class Stack
 {
 public:
     /// The pointer to the top item.
-    /// Once stack space is provided with reset() this is never null.
+    /// This is never null.
     uint256* top_item = nullptr;
 
 private:
@@ -43,6 +43,9 @@ private:
     uint256* m_bottom = nullptr;
 
 public:
+    /// Init with the provided stack space.
+    explicit Stack(uint256* stack_space_bottom) noexcept { reset(stack_space_bottom); }
+
     /// The current number of items on the stack.
     [[nodiscard]] int size() const noexcept { return static_cast<int>(top_item - m_bottom); }
 
@@ -84,15 +87,14 @@ struct AdvancedExecutionState : ExecutionState
     /// This is only needed to correctly calculate the "current gas left" value.
     uint32_t current_block_cost = 0;
 
-    AdvancedExecutionState() noexcept { stack.reset(stack_space.bottom()); }
+    AdvancedExecutionState() noexcept : stack{stack_space.bottom()} {}
 
     AdvancedExecutionState(const evmc_message& message, evmc_revision revision,
         const evmc_host_interface& host_interface, evmc_host_context* host_ctx,
         bytes_view _code) noexcept
-      : ExecutionState{message, revision, host_interface, host_ctx, _code}
-    {
-        stack.reset(stack_space.bottom());
-    }
+      : ExecutionState{message, revision, host_interface, host_ctx, _code},
+        stack{stack_space.bottom()}
+    {}
 
     /// Terminates the execution with the given status code.
     const Instruction* exit(evmc_status_code status_code) noexcept
--- a/lib/evmone/advanced_analysis.hpp
+++ b/lib/evmone/advanced_analysis.hpp
@@ -30,15 +30,69 @@ struct BlockInfo
 };
 static_assert(sizeof(BlockInfo) == 8);
 
+/// The view/controller for EVM stack.
+class Stack
+{
+public:
+    /// The pointer to the top item.
+    /// Once stack space is provided with reset() this is never null.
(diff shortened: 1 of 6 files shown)


lib/evmone/advanced_analysis.hpp
```diff
@@ -53,6 +108,7 @@ struct AdvancedExecutionState : ExecutionState
         const uint8_t* code_ptr, size_t code_size) noexcept
```
yperbasis, line 108: `bytes_view` for code?

lib/evmone/advanced_analysis.hpp
```diff
@@ -30,15 +30,70 @@ struct BlockInfo
...
     ///
     /// This is only needed to correctly calculate the "current gas left" value.
     uint32_t current_block_cost = 0;
 
-    using ExecutionState::ExecutionState;
+    AdvancedExecutionState() noexcept { stack.reset(stack_space.bottom()); }
+
+    AdvancedExecutionState(const evmc_message& message, evmc_revision revision,
+        const evmc_host_interface& host_interface, evmc_host_context* host_ctx,
+        const uint8_t* code_ptr, size_t code_size) noexcept
```
yperbasis, line 92: `bytes_view` for code?

chfast: Yes, I think it make sense.

chfast (commented):
Did I manage to break anything in Silkworm?

yperbasis:
> Did I manage to break anything in Silkworm?

Nope, Silkworm works fine with this change: https://github.com/torquem-ch/silkworm/pull/547/commits/7f21280dd5fa49e26d180b5e4783237d72effff9

yperbasis (approved):

lib/evmone/execution_state.hpp
```diff
@@ -177,9 +192,9 @@ struct ExecutionState
     } analysis{};
 
     Stack stack;
+    StackSpace stack_space;
```
gumb0, line 149: Is this one kept last member for alignment reasons? Perhaps add a comment about it.

lib/evmone/execution_state.hpp
```diff
@@ -24,32 +24,22 @@ using bytes = std::basic_string<uint8_t>;
...
-struct Stack
+/// The view/controller for EVM stack.
+class Stack
 {
-    /// The maximum number of stack items.
-    static constexpr auto limit = 1024;
-
+public:
     /// The pointer to the top item.
     /// This is never nullptr and when stack is empty it points to bottom(),
```
gumb0: It's not quite true now, as it's nullptr until `reset()` is called for the first time... Maybe comment somewhere that this class assumes that `reset()` is called after constructor before any other method?

test/unittests/execution_state_test.cpp
```diff
@@ -138,30 +138,29 @@ TEST(execution_state, reset_advanced)
...
-TEST(execution_state, stack_clear)
+TEST(execution_state, stack_reset)
 {
     evmone::Stack stack;
+    intx::uint256 stack_space[2]{};
+    stack.reset(stack_space);
 
-    stack.clear();
     EXPECT_EQ(stack.size(), 0);
-    EXPECT_EQ(stack.top_item + 1, stack.storage);
```
gumb0, line 147: Why not use `StackSpace` in these teste? Then you could keep checking `stack.top_item` against `stack_space.bottom()`

gumb0: I mean can you keep here
```
EXPECT_EQ(stack.top_item + 1, stack_space.bottom());
```

lib/evmone/advanced_analysis.hpp
```diff
@@ -30,15 +30,69 @@ struct BlockInfo
...
 
-    using ExecutionState::ExecutionState;
+    AdvancedExecutionState() noexcept { stack.reset(stack_space.bottom()); }
+
+    AdvancedExecutionState(const evmc_message& message, evmc_revision revision,
+        const evmc_host_interface& host_interface, evmc_host_context* host_ctx,
+        bytes_view _code) noexcept
+      : ExecutionState{message, revision, host_interface, host_ctx, _code}
+    {
+        stack.reset(stack_space.bottom());
```
gumb0, line 94: I got to the last commit, and now it seems that `Stack` could have a constructor doing this, then it really could be said that `Stack::top_item` is "never null".

(from the first commit I didn't see it because they `Stack` and `StackSpace` were constructed in different order.

chfast: Done.

test/unittests/execution_state_test.cpp
chfast: GCC complains that "array subscript -1 is outside array bounds". I gave up.

gumb0 (approved):

PR #425, opened 2022-01-27
baseline: Optimize PUSH instructions

chfast:
With 32+ code padding it is aways safe to up to 32 bytes when loading push data. The partial word can always be read as 8 bytes + bswap + shift.

TODO:
- [x] Clean-up the change which pads code with fixed 33 bytes. #440 
- [x] Implement optimization for all PUSH variants.

PR #440, opened 2022-02-15
baseline: Always pad code with 33 bytes

chfast:
#425 will use it to optimize PUSH instructions a bit.


chfast:
> Wasn't this already added? Or only for advanced?

Padding was added but not 33 bytes but the exact amount needed.

PR #442, opened 2022-02-22
test: Add more memory access cases

chfast:


test/unittests/evm_memory_test.cpp
```diff
@@ -186,28 +178,29 @@ TEST_P(evm, memory_access)
                 while (--h != t.memory_size_arg)
                     code += push(0);
 
-                code += push_size;
```
gumb0: this will use PUSH32 as opposed to PUSH5 before? (I guess it doesn't matter)

test/unittests/evm_memory_test.cpp
```diff
@@ -186,28 +178,29 @@ TEST_P(evm, memory_access)
...
                 code += push(0);
 
             code += bytecode{t.opcode};
 
-            auto const gas = 8796294610952;
+            const auto gas = 8796294610952;
             execute(gas, code);
 
             auto case_descr_str = std::ostringstream{};
-            case_descr_str << "offset = 0x" << std::hex << p.index << " size = 0x" << std::hex
```
gumb0: Does `to_string` output it in hexadecimal? then `std::hex` is not needed?

test/unittests/evm_memory_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+            while (h-- != 0)
+                code += push(0);
+
+            code += bytecode{t.opcode};
+
+            const auto gas = 8796294610952;
+            execute(gas, code);
+
+            auto case_descr_str = std::ostringstream{};
+            case_descr_str << "offset = 0x" << std::hex << to_string(p.index) << " size = 0x"
```
gumb0, line 218: Does `to_string` output it in hexadecimal? then `std::hex` is not needed?

gumb0 (approved):

test/unittests/evm_memory_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+            const auto num_args = int{instr::traits[t.opcode].stack_height_required};
+            auto h = std::max(num_args, t.memory_size_arg + 1);
+            auto code = bytecode{};
+
+            if (t.memory_size_arg >= 0)
+            {
+                while (--h != t.memory_size_arg)
+                    code += push(0);
+
+                code += push(p.size);
```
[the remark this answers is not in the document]
chfast: I'm guessing PUSH5 was to fit previous values. This probably was before we had bytecode utils.

test/unittests/evm_memory_test.cpp
chfast: Converted to proper hex.

PR #448, opened 2022-03-17
EVMC upgrade to 10.0.0-alpha.4

chfast:


lib/evmone/execution_state.hpp
```diff
@@ -138,6 +138,10 @@ struct ExecutionState
     size_t output_offset = 0;
     size_t output_size = 0;
 
+private:
+    evmc_tx_context m_tx = {};
```
chfast, line 142: Maybe better to name it `m_tx_context`.

lib/evmone/execution_state.hpp
```diff
@@ -181,5 +185,12 @@ struct ExecutionState
     }
 
     bool in_static_mode() const { return (msg->flags & EVMC_STATIC) != 0; }
```
gumb0, line 187: Sanitizers want [[nodiscard]] on this ontouched line.

chfast: This is because we added public/private specifiers and clang-tidy do not consider it as a record type any more. Fixed.

gumb0 (approved):

PR #389, opened 2021-09-30
EOF static relative jumps (EIP-4200)

gumb0:

PR #452, opened 2022-03-29
RJUMP for legacy code

chfast:
The `RJUMP` and `RJUMPI` implementation ported from #389.

```sh
bin/evmc run --vm ./lib/libevmone.so,O=0,histogram --rev 12 5cfffd
Config: ./lib/libevmone.so,O=0,histogram
Executing on Cancun with 1000000 gas limit

--- # HISTOGRAM depth=0
opcode,count
RJUMP,200001
Result:   out of gas
Gas used: 1000000
```


axic:
I guess we can close this now?

PR #455, opened 2022-03-31
Add immediate_size and is_terminating to instr::Traits

chfast:


lib/evmone/instructions_traits.hpp
```diff
@@ -177,6 +177,12 @@ struct Traits
     /// The instruction name;
     const char* name = nullptr;
 
+    /// Size of the immediate argument in bytes.
+    uint8_t immediate_size = 0;
+
+    /// Whether the instruction terminates execution.
+    bool is_terminating = false;
```
chfast, line 184: This is not currently the case, but all undefined instructions may be considered "terminating".

chfast: It does not matter for unimplemented in EOF context, because unimplemented are handled separately. Not used in other places.

test/unittests/instructions_test.cpp
```diff
@@ -16,20 +20,72 @@ constexpr int get_revision_defined_in(size_t op) noexcept
...
+        return false;
+    }
+}
+
+constexpr bool validate_traits()
+{
+    for (size_t op = 0x00; op <= 0xff; ++op)
+    {
+        const auto& tr = instr::traits[op];
+        if (tr.is_terminating != test::is_terminating(static_cast<evmc_opcode>(op)))
```
gumb0, line 57:
```suggestion
        if (tr.is_terminating != is_terminating(static_cast<evmc_opcode>(op)))
```

test/unittests/instructions_test.cpp
```diff
@@ -16,20 +20,72 @@ constexpr int get_revision_defined_in(size_t op) noexcept
...
+            throw std::logic_error{"invalid terminating instruction"};
+
+        if (tr.is_terminating && tr.immediate_size != 0)
+            throw std::logic_error{"terminating instruction has immediate bytes"};
+
+        if (tr.immediate_size != 0)
```
gumb0, line 63: maybe a bit simpler
```
if (op >= OP_PUSH1 && op <= OP_PUAH32)
{
 ...
}
else if (tr.immediate_size != 0)
  throw std::logic_error{"unexpected instruction with immediate bytes"};
```

gumb0 (approved):

test/unittests/instructions_test.cpp
```diff
@@ -16,36 +20,78 @@ constexpr int get_revision_defined_in(size_t op) noexcept
...
+
+    // since
+    constexpr auto expected_rev = get_revision_defined_in(Op);
+    static_assert(tr.since.has_value() ? *tr.since == expected_rev : expected_rev == unspecified);
+}
+
+template <std::size_t... Ops>
+constexpr auto validate_traits(std::index_sequence<Ops...>)
+{
+    return (validate_traits_of<static_cast<evmc_opcode>(Ops)>(), ..., true);
```
gumb0, line 67: this is confusing... what type does `validate_traits` return?

test/unittests/instructions_test.cpp
```diff
@@ -16,36 +20,78 @@ constexpr int get_revision_defined_in(size_t op) noexcept
...
+
+    // immediate_size
+    if constexpr (Op >= OP_PUSH1 && Op <= OP_PUSH32)
+        static_assert(tr.immediate_size == Op - OP_PUSH1 + 1);
+    else
+        static_assert(tr.immediate_size == 0);
+
+    // is_terminating
+    static_assert(tr.is_terminating == is_terminating(Op));
+    static_assert(!tr.is_terminating || tr.immediate_size == 0);
```
gumb0, line 57: Add comment why this is a requirement

test/unittests/instructions_test.cpp
gumb0: ah, I think I got, it always returns `true` as result of repeated `,` operator, but calls 256 `validate_traits_of` as "side-effect" of expression.

gumb0: maybe possible to split to two lines
```
validate_traits_of<static_cast<evmc_opcode>(Ops)>(), ...;
return true;
```
?

gumb0 (approved):

PR #463, opened 2022-05-06
Add RLP encoding implementation to evmone::state library

chfast:


test/state/rlp.hpp
```diff
@@ -0,0 +1,117 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2021 The evmone Authors.
```
axic, line 2: :eyes:

test/state/rlp.hpp
```diff
@@ -0,0 +1,117 @@
...
+#include <cassert>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace evmone::rlp
+{
+using bytes = std::basic_string<uint8_t>;
+using bytes_view = std::basic_string_view<uint8_t>;
```
axic, line 17: Weren't these defined somewhere? Or only in Fizzy?

test/state/rlp.hpp
chfast: I started this last year :(

test/state/rlp.hpp
chfast: This is already defined in multiple places (2x in EVMC). But as long as the definition is the same this is fine. Notice this RLP implementation does not include any evmone/evmc headers. Also intx include can be easily dropped.

chfast:
> Should consider adding/reviewing test cases from

I think I can easily manually port "official" and geth tests.

test/state/rlp.hpp
```diff
@@ -0,0 +1,117 @@
...
+inline bytes_view trim(bytes_view b) noexcept
+{
+    b.remove_prefix(std::min(b.find_first_not_of(uint8_t{0x00}), b.size()));
+    return b;
+}
+
+template <typename T>
+inline decltype(rlp_encode(std::declval<T>())) encode(const T& v)
+{
+    return rlp_encode(v);
```
gumb0, line 56: confused about where is this defined

chfast: You need to provide an overloading for any custom type.

test/state/rlp.hpp
```diff
@@ -0,0 +1,117 @@
...
+        return {LongBase + 1, l0};
+    else if (const auto l1 = static_cast<uint8_t>(l >> 8); l <= 0xffff)
+        return {LongBase + 2, l1, l0};
+    else
+        return {LongBase + 3, static_cast<uint8_t>(l >> 16), l1, l0};
+}
+
+inline bytes wrap_list(const bytes& content)
+{
+    return internal::encode_length<0xc0, 0xf7>(content.size()) + content;
```
gumb0, line 40: perhaps use decimals because YP defines them as decimals, and also `short_cutoff` on line 24
```suggestion
    return internal::encode_length<192, 247>(content.size()) + content;
```

test/state/rlp.hpp
gumb0: Doesn it not work to require to define `encode(const T& v)` overload directly?

chfast: Probably, but there are 2 reasons to use different name:
1. These are ADL-aware functions defined outside of RLP namespace, so `encode()` is ambiguous.
2. Having single overloaded name for everything already caused issues, that's why I needed `tuple()` separated.

test/state/rlp.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+inline bytes encode(bytes_view data)
+{
+    static constexpr uint8_t short_base = 0x80;
+    if (data.size() == 1 && data[0] < short_base)
+        return {data[0]};
+
+    auto r = internal::encode_length<short_base, 0xb7>(data.size());
+    r += data;
+    return r;
```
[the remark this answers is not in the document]
chfast: Unfortunately, there is no `string + string_view` operator 🤷

test/unittests/state_rlp_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+{
+    const auto value = 0x00000000000000000000000000000000000000000000000000000000000001ff_bytes32;
+    const auto xvalue = rlp::encode(rlp::trim(value));
+    EXPECT_EQ(xvalue, "8201ff"_hex);
+}
+
+TEST(state_rlp, encode_mpt_node)
+{
+    const bytes path{0x20, 0x41};
+    const bytes value{'v', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_',
```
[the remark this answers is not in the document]
chfast: This is just a test case evolved...

test/state/rlp.hpp
```diff
@@ -0,0 +1,117 @@
...
+    return rlp_encode(v);
+}
+
+inline bytes encode(bytes_view data)
+{
+    static constexpr uint8_t short_base = 0x80;
+    if (data.size() == 1 && data[0] < short_base)
+        return {data[0]};
+
+    auto r = internal::encode_length<short_base, 0xb7>(data.size());
```
gumb0, line 65:
```suggestion
    auto r = internal::encode_length<short_base, 183>(data.size());
```

gumb0 (approved):
just one more place to use decimal value

test/state/rlp.hpp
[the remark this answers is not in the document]
chfast: Actually, I changed this to `internal::encode_length<short_base, 0xb7>(data.size()) += data`.

PR #467, opened 2022-05-25
CI: Upgrade compilers to GCC 12 and clang 14

chfast:


axic (approved):

circle.yml
```diff
@@ -3,18 +3,18 @@ version: 2.1
 executors:
   lint:
     docker:
-      - image: ethereum/cpp-build-env:17-lint
+      - image: ethereum/cpp-build-env:18-lint
```
axic, line 6: These wren't pushed to https://github.com/ethereum/cpp-build-env

axic (approved):

PR #469, opened 2022-06-02
test: Introduce _hex literal to build bytes

chfast:


test/unittests/eof_validation_test.cpp
```diff
@@ -180,12 +180,12 @@ TEST(eof_validation, EOF1_undefined_opcodes)
         EXPECT_EQ(validate_eof(cont), expected) << hex(cont);
     }
 
-    EXPECT_EQ(validate_eof(from_hex("EF0001 010001 00 FE")), EOFValidationError::success);
+    EXPECT_EQ(validate_eof("EF0001 010001 00 FE"), EOFValidationError::success);
```
gumb0, line 183: `_hex` missing?

PR #470, opened 2022-06-02
Upgrade EVMC hex utilities

chfast:


test/utils/utils.hpp
```diff
@@ -8,11 +8,13 @@
 using evmc::bytes;
 using evmc::bytes_view;
 using evmc::from_hex;
+using evmc::from_spaced_hex;
 using evmc::hex;
 
+
 inline bytes operator""_hex(const char* s, size_t size)
```
axic, line 15: Can't this just include the operator from evmc?

chfast: There is no such operator in EVMC.

gumb0 (approved):

PR #477, opened 2022-06-15
state: Add insert-only MPT implementation

chfast:
Add insert-only Merkle Patricia Trie implementation for computing MTP root hashes of (key, value) maps.

### DIFF
--- a/test/state/mpt.cpp
+++ b/test/state/mpt.cpp
@@ -1,0 +1,283 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2022 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "mpt.hpp"
+#include "rlp.hpp"
+#include <algorithm>
+#include <cassert>
+
+namespace evmone::state
+{
+namespace
+{
+/// The collection of nibbles (4-bit values) representing a path in a MPT.
+struct Path
+{
+    size_t length = 0;  // TODO: Can be converted to uint8_t.
+    uint8_t nibbles[64]{};
+
+    Path() = default;
+
+    explicit Path(bytes_view key) noexcept : length{2 * key.size()}
+    {
+        assert(length <= std::size(nibbles));
+        size_t i = 0;
+        for (const auto b : key)
+        {
+            // static_cast is only needed in GCC <= 8.
+            nibbles[i++] = static_cast<uint8_t>(b >> 4);
+            nibbles[i++] = static_cast<uint8_t>(b & 0x0f);
+        }
+    }
+
+    [[nodiscard]] Path tail(size_t pos) const noexcept
+    {
+        assert(pos > 0 && pos <= length);  // MPT never requests whole path copy (pos == 0).
+        Path p;
+        p.length = length - pos;
+        std::copy_n(&nibbles[pos], p.length, p.nibbles);
+        return p;
+    }
+
+    [[nodiscard]] Path head(size_t size) const noexcept
+    {
+        assert(size < length);  // MPT never requests whole path copy (size == length).
+        Path p;
+        p.length = size;
+        std::copy_n(nibbles, size, p.nibbles);
+        return p;
+    }
+
+    [[nodiscard]] bytes encode(bool extended) const
+    {
+        bytes bs;
+        const auto is_even = length % 2 == 0;
+        if (is_even)
+            bs.push_back(0x00);
+        else
+            bs.push_back(0x10 | nibbles[0]);
+        for (size_t i = is_even ? 0 : 1; i < length; ++i)
+        {
+            const auto h = nibbles[i++];
+            const auto l = nibbles[i];
+            assert(h <= 0x0f);
+            assert(l <= 0x0f);
+            bs.push_back(static_cast<uint8_t>((h << 4) | l));
+        }
+        if (!extended)
+            bs[0] |= 0x20;
+        return bs;
+    }
+};
+}  // namespace
+
+/// The MPT Node.
+///
+/// The implementation is based on StackTrie from go-ethereum.
+// clang-tidy bug: https://github.com/llvm/llvm-project/issues/50006
+// NOLINTNEXTLINE(bugprone-reserved-identifier)
+class MPTNode
+{
+    enum class Kind : uint8_t
+    {
+        leaf,
+        ext,
+        branch
+    };
+
+    static constexpr size_t num_children = 16;
+
+    Kind m_kind = Kind::leaf;
+    Path m_path;
+    bytes m_value;
+    std::unique_ptr<MPTNode> m_children[num_children];
+
+    explicit MPTNode(Kind kind, const Path& path = {}, bytes&& value = {}) noexcept
+      : m_kind{kind}, m_path{path}, m_value{std::move(value)}
+    {}
+
+    /// Creates an extended node.
+    static MPTNode ext(const Path& path, std::unique_ptr<MPTNode> child) noexcept
+    {
+        assert(child->m_kind == Kind::branch);
+        MPTNode node{Kind::ext, path};
+        node.m_children[0] = std::move(child);
+        return node;
+    }
+
+    /// Optionally wraps the child node with newly created extended node in case
+    /// the provided path is not empty.
+    static std::unique_ptr<MPTNode> optional_ext(
+        const Path& path, std::unique_ptr<MPTNode> child) noexcept
+    {
+        return (path.length != 0) ? std::make_unique<MPTNode>(ext(path, std::move(child))) :
+                                    std::move(child);
+    }
+
+    /// Creates a branch node out of two children and optionally extends it with an extended
+    /// node in case the path is not empty.
+    static MPTNode ext_branch(const Path& path, size_t idx1, std::unique_ptr<MPTNode> child1,
+        size_t idx2, std::unique_ptr<MPTNode> child2) noexcept
+    {
+        assert(idx1 != idx2);
+        assert(idx1 < num_children);
+        assert(idx2 < num_children);
+
+        MPTNode br{Kind::branch};
+        br.m_children[idx1] = std::move(child1);
+        br.m_children[idx2] = std::move(child2);
+
+        return (path.length != 0) ? ext(path, std::make_unique<MPTNode>(std::move(br))) :
+                                    std::move(br);
+    }
+
+    /// Finds the position at witch two paths differ.
+    static size_t mismatch(const Path& p1, const Path& p2) noexcept
+    {
+        assert(p1.length <= p2.length);
+        return static_cast<size_t>(
+            std::mismatch(p1.nibbles, p1.nibbles + p1.length, p2.nibbles).first - p1.nibbles);
+    }
+
+public:
+    MPTNode() = default;
+
+    /// Creates new leaf node.
+    static std::unique_ptr<MPTNode> leaf(const Path& path, bytes&& value) noexcept
+    {
+        return std::make_unique<MPTNode>(MPTNode{Kind::leaf, path, std::move(value)});
+    }
+
+    void insert(const Path& path, bytes&& value);
+
+    [[nodiscard]] hash256 hash() const;
+};
+
+void MPTNode::insert(const Path& path, bytes&& value)  // NOLINT(misc-no-recursion)
+{
+    // The insertion is all about branch nodes. In happy case we will find an empty slot
+    // in an existing branch node. Otherwise, we need to create new branch node
+    // (possibly with an adjusted extended node) and transform existing nodes around it.
+
+    switch (m_kind)
+    {
+    case Kind::branch:
+    {
+        assert(m_path.length == 0);  // Branch has no path.
+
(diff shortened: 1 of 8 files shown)


gumb0:
typo in commit message: MTP => MPT

test/state/mpt.cpp
```diff
@@ -0,0 +1,282 @@
...
+        size_t i = 0;
+        for (const auto b : k)
+        {
+            // static_cast is only needed in GCC <= 8.
+            nibbles[i++] = static_cast<uint8_t>(b >> 4);
+            nibbles[i++] = static_cast<uint8_t>(b & 0x0f);
+        }
+    }
+
+    [[nodiscard]] Path tail(size_t index) const
```
gumb0, line 33: noexcept

test/state/mpt.cpp
```diff
@@ -0,0 +1,282 @@
...
+    Path m_path{{}};
+    bytes m_value;
+    std::unique_ptr<MPTNode> children[num_children];
+
+    MPTNode(Kind kind, const Path& path, bytes&& value = {}) noexcept
+      : m_kind{kind}, m_path{path}, m_value{std::move(value)}
+    {}
+
+    /// Named constructor for an extended node.
+    static MPTNode ext(const Path& k, std::unique_ptr<MPTNode> child) noexcept
```
gumb0, line 98: Why `Path` variable is called `k` ? (I'd call it `path` as in constructor above)

test/state/mpt.cpp
```diff
@@ -0,0 +1,282 @@
...
+    {
+        MPTNode node{Kind::ext, k};
+        node.children[0] = std::move(child);
+        return node;
+    }
+
+    static size_t diff_index(const Path& p1, const Path& p2) noexcept
```
gumb0, line 105: Add a comment, it's not immediately clear what it does

test/state/mpt.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    /// Named constructor for a leaf node.
+    static MPTNode leaf(const Path& k, bytes&& v) noexcept { return {Kind::leaf, k, std::move(v)}; }
+
+    void insert(const Path& k, bytes&& v);
+
+    [[nodiscard]] hash256 hash() const;
+};
+
+MPT::MPT() noexcept = default;
```
gumb0, line 123: Cosmetic: maybe move `MPT` method definitions after `class MPTNode` and all `MPTNode` method definitions.

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+        assert(pos <= length);
+        Path p{{}};
+        p.length = length - pos;
+        std::copy_n(&nibbles[pos], p.length, p.nibbles);
+        return p;
+    }
+
+    [[nodiscard]] Path head(size_t size) const noexcept
+    {
+        assert(size < length);
```
gumb0, line 43: `tail(0)` can return entire original path, but `head(length)` cannot?

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+            bs.push_back(0x00);
+        else
+            bs.push_back(0x10 | nibbles[0]);
+        for (size_t i = is_even ? 0 : 1; i < length; ++i)
+        {
+            const auto h = nibbles[i++];
+            const auto l = nibbles[i];
+            assert(h <= 0x0f);
+            assert(l <= 0x0f);
+            bs.push_back(uint8_t((h << 4) | l));
```
gumb0, line 64: Why C cast?
```suggestion
            bs.push_back(static_cast<uint8_t>((h << 4) | l));
```

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+        if (!child)
+            child = std::make_unique<MPTNode>(leaf(path.tail(1), std::move(value)));
+        else
+            child->insert(path.tail(1), std::move(value));
+        break;
+    }
+
+    case Kind::ext:
+    {
+        const auto m = mismatch(m_path, path);
```
gumb0, line 144: too long and complicated code to use single-letter variables imho
```suggestion
        const auto mismatch_pos = mismatch(m_path, path);
```

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+        if (m == m_path.length)
+        {
+            // Go into child.
+            return children[0]->insert(path.tail(m), std::move(value));
+        }
+
+        std::unique_ptr<MPTNode> n;
```
gumb0, line 152:
```suggestion
        std::unique_ptr<MPTNode> node;
```

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+        if (m < m_path.length - 1)
+            n = std::make_unique<MPTNode>(ext(m_path.tail(m + 1), std::move(children[0])));
+        else
+            n = std::move(children[0]);
+
+        MPTNode* branch = nullptr;
```
gumb0, line 158: Add some comment(s) what's happening in this part - converting to a branch etc...

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+        if (m == 0)
+        {
+            branch = this;
+            branch->m_kind = Kind::branch;
+        }
+        else
+        {
+            branch = (children[0] = std::make_unique<MPTNode>()).get();
```
gumb0, line 166: I think it would be good to have `MPTNode::branch()` named constructor for consistency with `leaf()` and `ext()`, and also it would be clear that branch is initialized with `m_path` empty.

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+    };
+
+    static constexpr size_t num_children = 16;
+
+    Kind m_kind = Kind::leaf;
+    Path m_path{{}};
+    bytes m_value;
+    std::unique_ptr<MPTNode> children[num_children];
+
+    MPTNode(Kind kind, const Path& path, bytes&& value = {}) noexcept
```
gumb0, line 92: I think it's better to have `bytes value` parameter (pass by value). As I remember, `&&` is a universal reference, supposed to handle both r-value references and values, useful mostly in templates. Here you intend to always move into constructor.

Also in `leaf()` and `insert()`.

test/state/mpt.cpp
```diff
@@ -0,0 +1,286 @@
...
+        const auto m = mismatch(m_path, path);
+
+        if (m == m_path.length)
+        {
+            // Go into child.
+            return children[0]->insert(path.tail(m), std::move(value));
+        }
+
+        std::unique_ptr<MPTNode> n;
+        if (m < m_path.length - 1)
```
gumb0, line 153: perhaps add `assert(m_path.length > 0)`

chfast:
Inspired by the review, I changed the "insert" implementation in a way the nodes are built bottom-up. Therefore, they can be built completely in a functional way.

test/state/mpt.cpp
```diff
@@ -0,0 +1,283 @@
...
+            (mismatch_pos < m_path.length - 1) ?
+                std::make_unique<MPTNode>(
+                    ext(m_path.tail(mismatch_pos + 1), std::move(m_children[0]))) :
+                std::move(m_children[0]),
+            new_idx,
+            std::make_unique<MPTNode>(leaf(path.tail(mismatch_pos + 1), std::move(value))));
+
+        *this = (mismatch_pos == 0) ?
+                    std::move(br) :
+                    ext(m_path.head(mismatch_pos), std::make_unique<MPTNode>(std::move(br)));
```
gumb0, line 183: isn't this enough
```suggestion
                    ext(m_path.head(mismatch_pos), std::move(br));
```

test/state/mpt.cpp
```diff
@@ -0,0 +1,283 @@
...
+        {
+            // Paths match: go into the child.
+            return m_children[0]->insert(path.tail(mismatch_pos), std::move(value));
+        }
+
+        const auto orig_idx = m_path.nibbles[mismatch_pos];
+        const auto new_idx = path.nibbles[mismatch_pos];
+
+        auto br = branch(orig_idx,
+            (mismatch_pos < m_path.length - 1) ?
```
gumb0, line 174: this is quite complicated, I suggest to define `orig_child` and `new_child` temporaries.

test/state/mpt.cpp
```diff
@@ -0,0 +1,283 @@
...
+                std::make_unique<MPTNode>(
+                    ext(m_path.tail(mismatch_pos + 1), std::move(m_children[0]))) :
+                std::move(m_children[0]),
+            new_idx,
+            std::make_unique<MPTNode>(leaf(path.tail(mismatch_pos + 1), std::move(value))));
+
+        *this = (mismatch_pos == 0) ?
```
gumb0, line 181: What if `mismatch_pos == 1`, it will create new extension node with 1-nibble path, is this correct?

test/state/mpt.cpp
```diff
@@ -0,0 +1,283 @@
...
+    case Kind::branch:
+    {
+        assert(m_path.length == 0);
+
+        // Temporary storage for children hashes.
+        // The `bytes` type could be used instead, but this way dynamic allocation is avoided.
+        hash256 children_hashes[num_children];
+
+        // Views of children hash bytes. Additional item for hash list
+        // terminator (always empty). Does not seem needed for correctness,
```
gumb0, line 233: weird how it could not influence correctness, hash would be different without it?

test/unittests/state_mpt_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+TEST(state_mpt, branch_node_example1)
+{
+    // A trie of single branch node and two leaf nodes with paths of length 2.
+    // The branch node has leaf nodes at positions [4] and [5].
+    // {4:1, 5:a}
+
+    auto value1 = "v___________________________1"_b;
+    const auto key1 = "A"_b;
+    const uint8_t path1[]{0x4, 0x1};
+    const bytes encoded_path1{uint8_t(0x30 | path1[1])};
```
gumb0, line 71: static_cast maybe

test/unittests/state_mpt_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+    const auto leaf_node1 = rlp::encode_tuple(encoded_path1, value1);
+    EXPECT_EQ(hex(leaf_node1), "df319d765f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f5f31");
```
gumb0, line 73: well this tests just how `rlp::encode_tuple` works and doesn't influence the check for `trie.hash()` in the end...

maybe either make a separate test with these two `leaf_node` checks, or just delete them?

test/unittests/state_mpt_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+    trie.insert(key1, std::move(value1));
+    trie.insert(key2, std::move(value2));
+    EXPECT_EQ(hex(trie.hash()), "3eefc183db443d44810b7d925684eb07256e691d5c9cb13215660107121454f9");
+}
+
+TEST(state_mpt, extension_node_example2)
+{
+    // A trie of an extension node followed by a branch node with
+    // two leafs with longer paths.
+    // 585:{8:4a, 9:5a}
```
gumb0, line 122:
```suggestion
    // 585:{8:41, 9:5a}
```

test/unittests/state_mpt_test.cpp
chfast: This is true, they just try to reveal the internal of the trie, but these are not directly accessible by API any more. I left these as documentation for new implementations.

test/unittests/state_mpt_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+    auto value1 = "v___________________________1"_b;
+    const auto key1 = "XXA"_b;
+    const uint8_t path1[]{0x5, 0x8, 0x5, 0x8, 0x4, 0x1};
+
+    auto value2 = "v___________________________2"_b;
+    const auto key2 = "XYZ"_b;
+    const uint8_t path2[]{0x5, 0x8, 0x5, 0x9, 0x5, 0xa};
+
+    const uint8_t common_path[]{0x5, 0x8, 0x5};
+    const bytes encoded_path1{0x20, uint8_t((path1[4] << 4) | path1[5])};
```
gumb0, line 133: static_cast maybe
(several places in this test)

gumb0 (approved):

test/state/mpt.cpp
```diff
@@ -0,0 +1,279 @@
...
+        assert(pos <= length);
+        Path p;
+        p.length = length - pos;
+        std::copy_n(&nibbles[pos], p.length, p.nibbles);
+        return p;
+    }
+
+    [[nodiscard]] Path head(size_t size) const noexcept
+    {
+        assert(size < length);
```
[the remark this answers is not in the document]
chfast: The assert comes from how this is used in MPT: using `head()` or `tail()` it may produce empty path but never produces a full copy of a path. I added separate asserts for this scope narrowing.

PR #478, opened 2022-06-17
state: Define MPT root hash procedure for accounts

chfast:


test/state/mpt_hash.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace evmone::state
+{
+namespace
+{
+hash256 mpt_hash(const std::unordered_map<hash256, StorageValue>& storage)
+{
+    MPT trie;
+    for (const auto& [key, value] : storage)
+    {
+        if (!is_zero(value.current))  // Skip "deleted" values.
```
gumb0, line 19: Maybe this is rather logic of how MPT data structure is used in Ethereum... so not logic of hashing itself, may be better kept outside.

But I don't mind either way.

test/state/account.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <unordered_map>
+
+namespace evmone::state
+{
+using evmc::address;
+using evmc::bytes;
+using evmc::bytes32;
+
+/// The representation of the account storage value.
+struct StorageValue
```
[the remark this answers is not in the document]
chfast: I stripped field that are not needed currently but left the struct. Later I will add fields like "original value" and "accessed flag". The raw `bytes32` can also be used for now.

test/state/mpt_hash.cpp
chfast: Both "hash functions" are use-case specific because at least you need to decide how to serialize a structure. As long as we will keep original and current value together we cannot erase instead zeroing.

In more generic fashion in this case `std::copy_if` could be used if we support `std::insterter()`. In the other case we would need something like "transform if" so ranges should be more suited. Interesting exercise, but decided is too big detour.

gumb0 (approved):

PR #479, opened 2022-06-17
evmone-statetest tool with JSON test loading

chfast:


test/statetest/statetest.hpp
```diff
@@ -0,0 +1,63 @@
...
+        tx.data = inputs.at(indexes.input);
+        tx.gas_limit = gas_limits.at(indexes.gas_limit);
+        tx.value = values.at(indexes.value);
+        return tx;
+    }
+};
+
+struct TestCase
+{
+    struct Case
```
gumb0, line 41: Seems a bit inconsistent that `TestCase` has a vector of nested `TestCase::Case`, but `StateTransitionTest` has a vector of not-nested `TestCase`.

So maybe this shouln't be nested? 
And renamin suggestion; `Case` => `Expectation` or `CaseExpectation`

test/statetest/statetest_loader.cpp
```diff
@@ -0,0 +1,198 @@
...
+    return intx::from_string<intx::uint256>(s);
+}
+
+template <>
+state::AccessList from_json<state::AccessList>(const json::json& j)
+{
+    state::AccessList o;
+    for (const auto& a : j)
+    {
+        o.push_back({from_json<address>(a.at("address")), {}});
```
gumb0, line 65: emplace_back?

gumb0 (approved):
Looks good overall.

Suprisingly not that many weird hacks required to parse a test.

test/statetest/statetest.hpp
chfast: I actually moved structs into `StateTransitionTest`. Usually I don't do this but here this looks nice.

test/statetest/statetest_loader.cpp
chfast: Done.

gumb0 (approved):

test/statetest/statetest_loader.cpp
```diff
@@ -0,0 +1,199 @@
...
+            slot.current = value;
+        }
+    }
+
+    o.multi_tx = j_t.at("transaction").get<TestMultiTransaction>();
+
+    o.block = from_json<state::BlockInfo>(j_t.at("env"));
+
+    for (const auto& [rev_name, expectations] : j_t.at("post").items())
+        o.cases.push_back({to_rev(rev_name),
```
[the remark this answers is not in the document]
chfast: I will try, but this may not work because this class does not have explicit constructor. I believe this is fixed in C++20 or C++23 where `S(a, b)` is valid initialization for an aggregate `S`.

PR #484, opened 2022-07-15
Add implementation of evmc::Host and state transition

chfast:
This implements state transition as application of a single transaction to a predefined state. Most of the logic is in the `evmc::Host` implementation plus the transaction specific handling in `evmone::state::transition()`.

These are used to execute all tests by the statetest runner.

The precompiles are not implemented.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "host.hpp"
+#include "precompiles.hpp"
 #include "rlp.hpp"
 
 namespace evmone::state
@@ -246,6 +247,9 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
         m_state.get(msg.sender).balance -= value;
         dst_acc->balance += value;
     }
+
+    if (auto precompiled_result = call_precompile(m_rev, msg); precompiled_result.has_value())
+        return std::move(*precompiled_result);
 
     // Copy of the code. Revert will invalidate the account.
     const auto code = dst_acc != nullptr ? dst_acc->code : bytes{};
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -1,0 +1,337 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2022 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "host.hpp"
+#include "rlp.hpp"
+
+namespace evmone::state
+{
+bool Host::account_exists(const address& addr) const noexcept
+{
+    const auto* const acc = m_state.find(addr);
+    return acc != nullptr && (m_rev < EVMC_SPURIOUS_DRAGON || !acc->is_empty());
+}
+
+bytes32 Host::get_storage(const address& addr, const bytes32& key) const noexcept
+{
+    const auto& acc = m_state.get(addr);
+    if (const auto it = acc.storage.find(key); it != acc.storage.end())
+        return it->second.current;
+    return {};
+}
+
+evmc_storage_status Host::set_storage(
+    const address& addr, const bytes32& key, const bytes32& value) noexcept
+{
+    // Follow EVMC documentation https://evmc.ethereum.org/storagestatus.html#autotoc_md3
+    // and EIP-2200 specification https://eips.ethereum.org/EIPS/eip-2200.
+
+    auto& storage_slot = m_state.get(addr).storage[key];
+    const auto& [current, original, _] = storage_slot;
+
+    const auto dirty = original != current;
+    const auto restored = original == value;
+    const auto current_is_zero = is_zero(current);
+    const auto value_is_zero = is_zero(value);
+
+    auto status = EVMC_STORAGE_ASSIGNED;  // All other cases.
+    if (!dirty && !restored)
+    {
+        if (current_is_zero)
+            status = EVMC_STORAGE_ADDED;  // 0 → 0 → Z
+        else if (value_is_zero)
+            status = EVMC_STORAGE_DELETED;  // X → X → 0
+        else
+            status = EVMC_STORAGE_MODIFIED;  // X → X → Z
+    }
+    else if (dirty && !restored)
+    {
+        if (current_is_zero && !value_is_zero)
+            status = EVMC_STORAGE_DELETED_ADDED;  // X → 0 → Z
+        else if (!current_is_zero && value_is_zero)
+            status = EVMC_STORAGE_MODIFIED_DELETED;  // X → Y → 0
+    }
+    else if (dirty && restored)
+    {
+        if (current_is_zero)
+            status = EVMC_STORAGE_DELETED_RESTORED;  // X → 0 → X
+        else if (value_is_zero)
+            status = EVMC_STORAGE_ADDED_DELETED;  // 0 → Y → 0
+        else
+            status = EVMC_STORAGE_MODIFIED_RESTORED;  // X → Y → X
+    }
+
+    storage_slot.current = value;  // Update current value.
+    return status;
+}
+
+uint256be Host::get_balance(const address& addr) const noexcept
+{
+    const auto* const acc = m_state.find(addr);
+    return (acc != nullptr) ? intx::be::store<uint256be>(acc->balance) : uint256be{};
+}
+
+size_t Host::get_code_size(const address& addr) const noexcept
+{
+    const auto* const acc = m_state.find(addr);
+    return (acc != nullptr) ? acc->code.size() : 0;
+}
+
+bytes32 Host::get_code_hash(const address& addr) const noexcept
+{
+    // TODO: Cache code hash. It will be needed also to compute the MPT hash.
+    const auto* const acc = m_state.find(addr);
+    return (acc != nullptr && !acc->is_empty()) ? keccak256(acc->code) : bytes32{};
+}
+
+size_t Host::copy_code(const address& addr, size_t code_offset, uint8_t* buffer_data,
+    size_t buffer_size) const noexcept
+{
+    const auto* const acc = m_state.find(addr);
+    const auto code = (acc != nullptr) ? bytes_view{acc->code} : bytes_view{};
+    const auto code_slice = code.substr(std::min(code_offset, code.size()));
+    const auto num_bytes = std::min(buffer_size, code_slice.size());
+    std::copy_n(code_slice.begin(), num_bytes, buffer_data);
+    return num_bytes;
+}
+
+bool Host::selfdestruct(const address& addr, const address& beneficiary) noexcept
+{
+    // Touch beneficiary and transfer all balance to it.
+    // This may happen multiple times per single account as account's balance
+    // can be increased with a call following previous selfdestruct.
+    auto& acc = m_state.get(addr);
+    m_state.touch(beneficiary).balance += acc.balance;
+    acc.balance = 0;  // Zero balance (this can be the beneficiary).
+
+    // Mark the destruction if not done already.
+    return !std::exchange(acc.destructed, true);
+}
+
+static address compute_new_address(const evmc_message& msg, uint64_t sender_nonce) noexcept
+{
+    hash256 addr_base_hash;
+    if (msg.kind == EVMC_CREATE)
+    {
+        // TODO: Compute CREATE address without using RLP library.
+        const auto rlp_list = rlp::encode_tuple(address{msg.sender}, sender_nonce);
+        addr_base_hash = keccak256(rlp_list);
+    }
+    else
+    {
+        assert(msg.kind == EVMC_CREATE2);
+        const auto init_code_hash = keccak256({msg.input_data, msg.input_size});
+        uint8_t buffer[1 + sizeof(msg.sender) + sizeof(msg.create2_salt) + sizeof(init_code_hash)];
+        static_assert(std::size(buffer) == 85);
+        buffer[0] = 0xff;
+        std::copy_n(msg.sender.bytes, sizeof(msg.sender), &buffer[1]);
+        std::copy_n(
+            msg.create2_salt.bytes, sizeof(msg.create2_salt), &buffer[1 + sizeof(msg.sender)]);
+        std::copy_n(init_code_hash.bytes, sizeof(init_code_hash),
+            &buffer[1 + sizeof(msg.sender) + sizeof(msg.create2_salt)]);
+        addr_base_hash = keccak256({buffer, std::size(buffer)});
+    }
+    evmc_address new_addr{};
+    std::copy_n(&addr_base_hash.bytes[12], sizeof(new_addr), new_addr.bytes);
+    return new_addr;
+}
+
+std::optional<evmc_message> Host::prepare_message(evmc_message msg)
+{
+    auto& sender_acc = m_state.get(msg.sender);
+    const auto sender_nonce = sender_acc.nonce;
+
+    // Bump sender nonce.
+    if (msg.depth == 0 || msg.kind == EVMC_CREATE || msg.kind == EVMC_CREATE2)
+    {
+        if (sender_nonce == Account::NonceMax)
+            return {};  // Light early exception, cannot happen for depth == 0.
+        ++sender_acc.nonce;
+    }
+
+    if (msg.kind == EVMC_CREATE || msg.kind == EVMC_CREATE2)
+    {
+        // Compute and fill create address.
+        assert(msg.recipient == address{});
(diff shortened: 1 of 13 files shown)


test/state/host.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+public:
+    Host(evmc_revision rev, evmc::VM& vm, State& state, const BlockInfo& block,
+        const Transaction& tx) noexcept
+      : m_rev{rev}, m_vm{vm}, m_state{state}, m_block{block}, m_tx{tx}
+    {}
+
+    [[nodiscard]] const auto& get_destructs() const noexcept { return m_destructs; }
+
+    [[nodiscard]] std::vector<Log>&& take_logs() noexcept { return std::move(m_logs); }
```
axic, line 34: Why is this moving while the others are just references?

test/state/host.cpp
```diff
@@ -0,0 +1,368 @@
...
+{
+bool Host::account_exists(const address& addr) const noexcept
+{
+    const auto* const acc = m_state.get_or_null(addr);
+    return acc != nullptr && (m_rev < EVMC_SPURIOUS_DRAGON || !acc->is_empty());
+}
+
+bytes32 Host::get_storage(const address& addr, const bytes32& key) const noexcept
+{
+    const auto& acc = m_state.get(addr);
```
axic, line 19: Wouldn't this put it into the hot list?

test/state/host.cpp
```diff
@@ -0,0 +1,368 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "host.hpp"
+#include "rlp.hpp"
+
+namespace evmone::state
+{
+bool Host::account_exists(const address& addr) const noexcept
+{
+    const auto* const acc = m_state.get_or_null(addr);
```
axic, line 13: Wouldn't this also put it into the hot list?

test/state/state.hpp
```diff
@@ -24,6 +24,26 @@ class State
...
+
+    Account* get_or_null(const address& addr)
+    {
+        const auto it = m_accounts.find(addr);
+        if (it != m_accounts.end())
+            return &it->second;
+        return nullptr;
+    }
+
+    Account& get_or_create(const address& addr) { return m_accounts[addr]; }
```
axic, line 42: Hm, so will `get()` fail if the account doesn't exists? Or why do you need `get_or_create`?

test/state/precompiles.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "precompiles.hpp"
+
+namespace evmone::state
+{
+using namespace evmc::literals;
+
+std::optional<evmc::Result> call_precompile(evmc_revision rev, const evmc_message& msg) noexcept
+{
+    if (evmc::is_zero(msg.code_address) || msg.code_address > 0x09_address)
```
axic, line 13: Why have `is_zero`? Wouldn't `> 0x09_address` be enough?

test/statetest/statetest_runner.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace evmone::test
+{
+void run_state_test(const StateTransitionTest& test, evmc::VM& vm)
+{
+    for (const auto& [rev, cases] : test.cases)
+    {
+        for (size_t case_index = 0; case_index != cases.size(); ++case_index)
+        {
+            SCOPED_TRACE(std::string{evmc::to_string(rev)} + '/' + std::to_string(case_index));
+            // if (rev != EVMC_FRONTIER)
```
axic, line 28: What are these comments?

test/state/host.cpp
chfast: This will not (it will not create an account if it doesn't exist). And I believe it should not - this is related to old call cost penalty check. But we should have a test which check if this does not affect account access list.

test/state/host.cpp
chfast: This assumes the account must exist already. And it is hot already.

test/state/host.hpp
chfast: We return the logs in the end so we can move them out of the Host. The other structures are discarded.

test/state/precompiles.cpp
chfast: There is no precompile at address 0.

test/statetest/statetest_runner.cpp
chfast: There are for debugging. You can narrow a test to single revision and/or case.

test/state/precompiles.cpp
axic: Oh right.

test/state/state.hpp
```diff
@@ -24,6 +24,24 @@ class State
         assert(r.second);
         return r.first->second;
     }
+
+    Account& get(const address& addr)
```
gumb0, line 28: should there be `const` version?

test/state/state.hpp
```diff
@@ -24,6 +24,24 @@ class State
...
+    {
+        assert(m_accounts.count(addr) == 1);
+        return m_accounts.find(addr)->second;
+    }
+
+    Account* get_or_null(const address& addr)
```
gumb0, line 34: some std containers (`variant`) call this `get_if`

test/state/state.hpp
gumb0: maybe this is just `operator[]` ?

test/state/state.hpp
gumb0: Also for other methods

gumb0: `noexcept`

test/state/state.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+namespace stdx
+{
+/// Implementation of std::erase_if from C++20.
+/// Taken from https://en.cppreference.com/w/cpp/container/unordered_map/erase_if.
+template <class Key, class T, class Hash, class KeyEqual, class Alloc, class Pred>
+typename std::unordered_map<Key, T, Hash, KeyEqual, Alloc>::size_type erase_if(
+    std::unordered_map<Key, T, Hash, KeyEqual, Alloc>& c, Pred pred)
+{
+    auto old_size = c.size();
```
gumb0, line 18: can be const

test/state/state.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    if (sender_acc.nonce == Account::NonceMax)
+        return {};
+
+    const auto tx_max_cost_512 = intx::umul(intx::uint256{tx.gas_limit}, tx.max_gas_price);
+    auto sender_balance_check = sender_acc.balance;
+    if (sender_balance_check < tx_max_cost_512)
+        return {};
+
+    sender_balance_check -= static_cast<intx::uint256>(tx_max_cost_512);
```
gumb0, line 114: why not check `sender_acc.balance >= tx_max_cost_512 + tx.value` ?

test/state/state.cpp
```diff
@@ -0,0 +1,169 @@
...
+        {},
+        recipient,
+    };
+}
+}  // namespace
+
+std::optional<std::vector<Log>> transition(
+    State& state, const BlockInfo& block, const Transaction& tx, evmc_revision rev, evmc::VM& vm)
+{
+    if (rev < EVMC_LONDON && tx.kind == Transaction::Kind::eip1559)
```
gumb0, line 86: Maybe makes sense to extract transaction-related checks into a separate transaction validity funtion.

test/state/state.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const auto effective_gas_price = base_fee + priority_gas_price;
+
+    assert(effective_gas_price <= tx.max_gas_price);
+    const auto tx_max_cost = tx.gas_limit * effective_gas_price;
+
+    sender_acc.balance -= tx_max_cost;  // Modify sender balance after all checks.
+
+    Host host{rev, vm, state, block, tx};
+
+    const auto result = host.call(build_message(tx, execution_gas_limit));
```
gumb0, line 135: Where do sender and recipient get their `touched` flag set? inside `call`?

chfast: In `Host::call()` for `CALL` only: `recipient_acc.touched = true;`.
Sender is not touched.
Are you confused by "touched" vs "warm"?

test/state/state.hpp
chfast: I replaced single `operator[]` with this ugly names because I want to know where new accounts are created (in the context of State cache). This already requires one more round of redesign and probably one more when state journal will be introduced.

From the current usage I need should replace this one with:
`get_or_create(address, Account{...})` because I often need to modify the created account straight away.

test/state/host.cpp
```diff
@@ -0,0 +1,369 @@
...
+    // Check collision as defined in pseudo-EIP https://github.com/ethereum/EIPs/issues/684.
+    // All combinations of conditions (nonce, code, storage) are tested.
+    if (const auto collision_acc = m_state.get_or_null(msg.recipient);
+        collision_acc != nullptr && (collision_acc->nonce != 0 || !collision_acc->code.empty()))
+        return evmc::Result{EVMC_OUT_OF_GAS};
+
+    auto& new_acc = m_state.get_or_create(msg.recipient);
+    if (m_rev >= EVMC_SPURIOUS_DRAGON)
+        new_acc.nonce = 1;
+    new_acc.storage.clear();  // In case of collision.
```
gumb0, line 189: confused why this is needed, doesn't it return error above in case of collision?

test/state/host.cpp
```diff
@@ -0,0 +1,369 @@
...
+    const auto value = intx::be::load<intx::uint256>(msg.value);
+    assert(sender_acc.balance >= value && "EVM must guarantee balance");
+    sender_acc.balance -= value;
+    new_acc.balance += value;  // The new account may be prefunded.
+
+    auto create_msg = msg;
+    create_msg.input_data = nullptr;
+    create_msg.input_size = 0;
+
+    // Execution can modify the state, iterators are invalidated.
```
gumb0, line 201: Do you mean `new_acc` and `sender_acc`? How can they be invalidated, they're in a unordered_map?

test/state/host.cpp
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        return r;
+    }
+
+    // Reject EF code.
+    if (m_rev >= EVMC_LONDON && !code.empty() && code[0] == 0xEF)
+        return evmc::Result{EVMC_OUT_OF_GAS};
+
+    // TODO: The new_acc pointer is invalid because of the state revert implementation,
```
gumb0, line 235: Also not sure what this means

test/state/host.cpp
chfast: The collision check ignores storage so you can have an "empty" account with non-empty storage (only in tests though). I have better version of this with better comments and I should bring it here.

test/state/host.cpp
chfast: They are invalidated when a call is reverted. We make a copy of unordered_map before the call and assign it back to the state in case of revert. This assign invalidates the pointers to accounts (they are replaced by the copy).

This is not happen if state journal is implemented. Then we keep the same unordered_map all the time.

test/state/host.cpp
chfast: This comment seems a bit off, probably imperfect description of the comment below. I will remove it.

test/state/host.cpp
```diff
@@ -0,0 +1,373 @@
...
+        result.create_address = msg.recipient;
+        return result;
+    }
+
+    auto gas_left = result.gas_left;
+    assert(gas_left >= 0);
+
+    bytes_view code{result.output_data, result.output_size};
+    if (m_rev >= EVMC_SPURIOUS_DRAGON && code.size() > 0x6000)
+        return evmc::Result{EVMC_OUT_OF_GAS};
```
chfast, line 214: We may add additional error codes to EVMC, but maybe for now we should use generic `EVMC_FAILURE` instead of `EVMC_OUT_OF_GAS`.

test/state/host.cpp
```diff
@@ -0,0 +1,370 @@
...
+            return EVMC_ACCESS_WARM;
+    }
+
+    return m_accessed_addresses.insert(addr).second ? EVMC_ACCESS_COLD : EVMC_ACCESS_WARM;
+}
+
+evmc_access_status Host::access_storage(const address& addr, const bytes32& key) noexcept
+{
+    // Check tx access list.
+    // TODO: Tx access list can be applied to the storage cache before execution.
```
chfast, line 361: I have it done on the side. Should I include it here?

test/state/host.cpp
```diff
@@ -0,0 +1,369 @@
...
+    const auto msg = prepare_message(orig_msg);
+    if (!msg.has_value())
+        return evmc::Result{EVMC_OUT_OF_GAS, orig_msg.gas};  // Light exception.
+
+    auto state_snapshot = m_state;
+    auto destructs_snapshot = m_destructs.size();
+    auto access_addresses_snapshot = m_accessed_addresses;
+    auto logs_snapshot = m_logs.size();
+
+    auto result = execute_message(*msg);
```
gumb0, line 279: can be const?

test/state/host.cpp
gumb0: Ah I think I got by reading further, `call` can replace entire `m_state` in case of exception, as I understand.

test/state/host.cpp
```diff
@@ -0,0 +1,369 @@
...
+            m_state.get_or_create(addr_03).touched = true;
+    }
+    return result;
+}
+
+evmc_tx_context Host::get_tx_context() const noexcept
+{
+    // TODO: The effective gas price is already computed in transaction validation.
+    const auto priority_gas_price =
+        std::min(m_tx.max_priority_gas_price, m_tx.max_gas_price - m_block.base_fee);
```
gumb0, line 304: maybe `assert(m_tx.max_gas_price >= m_block.base_fee)`

test/state/host.cpp
```diff
@@ -0,0 +1,369 @@
...
+{
+    m_logs.push_back({addr, {data, data_size}, {topics, topics + topics_count}});
+}
+
+evmc_access_status Host::access_account(const address& addr) noexcept
+{
+    // TODO: Predefined warm addresses can be applied to the state cache before execution.
+
+    // Transaction {sender,to} are always warm.
+    if (addr == m_tx.to)
```
gumb0, line 339: Why not add all these and access list to `m_accessed_addresses` in `Host` constructor?

test/state/host.cpp
chfast: No because we have to move it.

test/state/host.cpp
chfast: On todo list. I have it done already so I will include the change.

chfast: Done, but a lot of new code landed.

test/state/host.cpp
gumb0: Do you mean to return value? copy will be elided even if it's const I think

test/statetest/statetest_runner.cpp
```diff
@@ -0,0 +1,47 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2022 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "../state/mpt_hash.hpp"
```
gumb0, line 5: Better `<test/state/mpt_hash.hpp>` ?

test/statetest/statetest.cpp
```diff
@@ -28,6 +28,11 @@ class StateTest : public testing::Test
 
 int main(int argc, char* argv[])
 {
+    static constexpr std::string_view filter_flag_name = "--gtest_filter";
+    const auto has_user_filter = std::count_if(argv, argv + argc, [](const char* arg) noexcept {
```
gumb0, line 32: Seems `find_if` is more appropriate

test/state/state.hpp
```diff
@@ -24,6 +24,24 @@ class State
...
+    {
+        const auto it = m_accounts.find(addr);
+        if (it != m_accounts.end())
+            return &it->second;
+        return nullptr;
+    }
+
+    Account& get_or_create(const address& addr) { return m_accounts[addr]; }
+
+    auto& get_accounts() { return m_accounts; }
```
gumb0, line 44: noexcept

test/statetest/statetest_runner.cpp
chfast: Doesn't work. I probably would need to specify additional include paths what does not sound worth it.

test/state/host.cpp
chfast: Yes, but `Result` is move-only type

PR #493, opened 2022-08-18
Implement gas refund

chfast:
Calculate, aggregate and report the gas refund. This requires changes to `SSTORE`, `SELFDESTRUCT` and calls (propagate gas refund).

### DIFF
--- a/lib/evmone/instructions_storage.cpp
+++ b/lib/evmone/instructions_storage.cpp
@@ -6,6 +6,91 @@
 
 namespace evmone::instr::core
 {
+namespace
+{
+/// The gas cost specification for storage instructions.
+struct StorageCostSpec
+{
+    bool net_cost;        ///< Is this net gas cost metering schedule?
+    int16_t warm_access;  ///< Storage warm access cost, YP: G_{warmaccess}
+    int16_t set;          ///< Storage addition cost, YP: G_{sset}
+    int16_t reset;        ///< Storage modification cost, YP: G_{sreset}
+    int16_t clear;        ///< Storage deletion refund, YP: R_{sclear}
+};
+
+/// Table of gas cost specification for storage instructions per EVM revision.
+/// TODO: This can be moved to instruction traits and be used in other places: e.g.
+///       SLOAD cost, replacement for warm_storage_read_cost.
+constexpr auto storage_cost_spec = []() noexcept {
+    std::array<StorageCostSpec, EVMC_MAX_REVISION + 1> tbl{};
+
+    // Legacy cost schedule.
+    for (auto rev : {EVMC_FRONTIER, EVMC_HOMESTEAD, EVMC_TANGERINE_WHISTLE, EVMC_SPURIOUS_DRAGON,
+             EVMC_BYZANTIUM, EVMC_PETERSBURG})
+        tbl[rev] = {false, 200, 20000, 5000, 15000};
+
+    // Net cost schedule.
+    tbl[EVMC_CONSTANTINOPLE] = {true, 200, 20000, 5000, 15000};
+    tbl[EVMC_ISTANBUL] = {true, 800, 20000, 5000, 15000};
+    tbl[EVMC_BERLIN] = {
+        true, instr::warm_storage_read_cost, 20000, 5000 - instr::cold_sload_cost, 15000};
+    tbl[EVMC_LONDON] = {
+        true, instr::warm_storage_read_cost, 20000, 5000 - instr::cold_sload_cost, 4800};
+    tbl[EVMC_PARIS] = tbl[EVMC_LONDON];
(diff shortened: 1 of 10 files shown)


lib/evmone/instructions_storage.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    for (size_t rev = EVMC_FRONTIER; rev <= EVMC_MAX_REVISION; ++rev)
+    {
+        auto& e = tbl[rev];
+        if (const auto c = storage_cost_spec[rev]; !c.net_cost)  // legacy
+        {
+            e[EVMC_STORAGE_ASSIGNED] = {c.reset, 0};
+            e[EVMC_STORAGE_ADDED] = {c.set, 0};
+            e[EVMC_STORAGE_DELETED] = {c.reset, c.clear};
+            e[EVMC_STORAGE_MODIFIED] = {c.reset, 0};
```
gumb0, line 61: This could also be
```suggestion
            e[EVMC_STORAGE_MODIFIED] = e[EVMC_STORAGE_ASSIGNED];
```
because they are same cases for legacy.

chfast: Yes, maybe even the other way around.

chfast: I made it `e[EVMC_STORAGE_ASSIGNED] = e[EVMC_STORAGE_MODIFIED]`.

lib/evmone/instructions_storage.cpp
```diff
@@ -6,6 +6,86 @@
 
 namespace evmone::instr::core
 {
+namespace
+{
+struct StorageCostSpec
```
chfast, line 11: We can move this to instruction traits and use in other places (e.g. cost for SLOAD, replacement for `warm_storage_read_cost`.

test/unittests/evm_storage_test.cpp
```diff
@@ -169,6 +170,120 @@ TEST_P(evm, sstore_cost)
...
+        int64_t reset = 5000;
+        int64_t clear = 15000;
+    } c;
+
+    auto test = [this](const evmc::bytes32& original, const evmc::bytes32& current,
+                    const evmc::bytes32& value, int64_t expected_gas_used,
+                    int64_t expected_gas_refund) {
+        auto& storage = host.accounts[msg.recipient].storage;
+        storage[key] = {current, original};
+        execute(sstore(0xef, calldataload(0)), value);
```
gumb0, line 195: maybe could be simpler
```suggestion
        execute(sstore(0xef, value), {});
```

test/unittests/evm_storage_test.cpp
```diff
@@ -169,6 +169,120 @@ TEST_P(evm, sstore_cost)
...
+
+    struct CostConstants
+    {
+        int64_t warmaccess = -1;
+        int64_t set = -1;
+        int64_t reset = -1;
+        int64_t clear = -1;
+    };
+
+    auto test = [this](const evmc::bytes32& original, const evmc::bytes32& current,
```
gumb0, line 244: Could this helper be reused between two tests?

test/unittests/evm_state_test.cpp
```diff
@@ -397,6 +397,28 @@ TEST_P(evm, selfdestruct_with_balance)
...
 
+TEST_P(evm, selfdestruct_gas_refund)
+{
+    rev = EVMC_BERLIN;  // The last revision with gas refund.
+    const auto code = selfdestruct(0xbe);
+    execute(code);
+    EXPECT_GAS_USED(EVMC_SUCCESS, 7603);  // Cold access to 0xbe.
+    EXPECT_EQ(result.gas_refund, 24000);
+
+    // Second selfdestruct of the same account.
```
gumb0, line 408: Maybe also add a test for another selfdestruct of different account

test/unittests/evm_calls_test.cpp
```diff
@@ -738,3 +738,35 @@ TEST_P(evm, returndatacopy_outofrange)
...
+TEST_P(evm, call_gas_refund_aggregation)
+{
+    rev = EVMC_LONDON;
+    host.accounts[msg.recipient].set_balance(1);
+    host.call_result.status_code = EVMC_SUCCESS;
+    host.call_result.gas_refund = 1;
+
+    const auto a = 0xaa_address;
+    const auto code =
+        call(a) + callcode(a) + delegatecall(a) + staticcall(a) + create() + create2();
```
gumb0, line 768: may be redundant, but I thought of having two calls of each type to check that it's also aggregated for same type calls.

gumb0 (approved):
Looks good overall, some minor suggestions only in tests.

test/unittests/evm_storage_test.cpp
chfast: This is rather difficult because it accesses `this`.

PR #496, opened 2022-08-22
baseline: Refactor execute dispatch

chfast:
Limit the scope of the core dispatch functions.


lib/evmone/baseline.cpp
```diff
@@ -266,11 +254,30 @@ evmc_result execute(const VM& vm, ExecutionState& state, const CodeAnalysis& ana
...
+}  // namespace
+
+evmc_result execute(const VM& vm, ExecutionState& state, const CodeAnalysis& analysis) noexcept
+{
+    state.analysis.baseline = &analysis;  // Assign code analysis for instruction implementations.
+
+    // Use padded code.
+    state.code = {analysis.padded_code.get(), state.code.size()};
+
+    const auto& cost_table = get_baseline_cost_table(state.rev);
```
gumb0, line 270: Is it necessary to pass it from here to internal `execute`? It seems you could leave it there and then have just `state` and `tracer` parameters.

chfast: Yes, this is true. I tried both to check if it affects generated code but it doesn't.
I did it this way, because we will have another function like this for "computed goto". So I think the smaller the better.

gumb0 (approved):

PR #497, opened 2022-08-23
EIP-4750: EOF - Functions

gumb0:


lib/evmone/eof.cpp
```diff
@@ -18,11 +19,24 @@ constexpr uint8_t MAGIC[] = {0xef, 0x00};
...
+    const auto section_count =
+        headers[TYPE_SECTION].size() + headers[CODE_SECTION].size() + headers[DATA_SECTION].size();
+
+    constexpr auto section_header_size = 3;  // (SECTION_ID + SIZE) per each section
+
+    return sizeof(MAGIC) + 1 +  // 1 version byte
+           section_count * section_header_size + sizeof(TERMINATOR);
+}
+
+std::pair<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_view container)
```
axic, line 39: Why is this not `noexcept` anymore?

lib/evmone/eof.cpp
```diff
@@ -46,20 +60,30 @@ std::pair<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_view
...
                 break;
+            case TYPE_SECTION:
+                if (!section_headers[TYPE_SECTION].empty())
+                    return {{}, EOFValidationError::multiple_type_sections};
+                if (!section_headers[CODE_SECTION].empty())
+                    return {{}, EOFValidationError::code_section_before_type_section};
+                state = State::section_size;
+                break;
             case DATA_SECTION:
-                if (section_headers[CODE_SECTION] == 0)
```
axic, line 54: I think you can't remove this from here?

lib/evmone/eof.cpp
gumb0: Number of sections is arbitrary now and so function does dynamic allocation (to return all sections' sizes)

lib/evmone/eof.cpp
gumb0: It's replaced by another check in `case CODE_SECTION:`
```
if (!section_headers[DATA_SECTION].empty())
    return {{}, EOFValidationError::data_section_before_code_section};
```
(because now invalid case can be like code-data-code)

plus the old check in `case TERMINATOR` guarantees that at least one code section is present.

lib/evmone/instructions.hpp
```diff
@@ -701,6 +701,43 @@ inline code_iterator jumpi(StackTop stack, ExecutionState& state, code_iterator
...
+    const auto pc_post = pc + 1 + 1 /* count */ + count * REL_OFFSET_SIZE /* tbl */;
+
+    if (case_ >= count)
+    {
+        return pc_post;
+    }
+    else
+    {
+        const auto rel_offset_hi = pc[1 + static_cast<uint16_t>(case_) * REL_OFFSET_SIZE];
+        const auto rel_offset_lo = pc[2 + static_cast<uint16_t>(case_) * REL_OFFSET_SIZE];
```
gumb0, line 734: Is this off-by-one error? 
```suggestion
        const auto rel_offset_hi = pc[2 + static_cast<uint16_t>(case_) * REL_OFFSET_SIZE];
        const auto rel_offset_lo = pc[3 + static_cast<uint16_t>(case_) * REL_OFFSET_SIZE];
```
`pc[1]` is count, for `case == 0`, it should take `pc[2]` and `pc[3]`.

test/utils/bytecode.hpp
```diff
@@ -242,6 +244,24 @@ inline bytecode jumpi(bytecode target, bytecode condition)
...
+{
+    return OP_RJUMP + big_endian(offset);
+}
+
+inline bytecode rjumpi(int16_t offset, bytecode condition)
+{
+    return condition + OP_RJUMPI + bytecode{big_endian(offset)};
+}
+
+inline bytecode rjumpv(const std::initializer_list<uint16_t> offsets, bytecode condition)
```
gumb0, line 257: sorry this was wrong
```suggestion
inline bytecode rjumpv(const std::initializer_list<int16_t> offsets, bytecode condition)
```

lib/evmone/eof.cpp
```diff
@@ -88,83 +143,414 @@ std::pair<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_view
...
+            static_cast<uint16_t>(container[offset + 2] << 8 | container[offset + 3])});
+    }
+
+    // check 1st section is (0, 0)
+    if (types[0].inputs_num != 0 || types[0].outputs_num != 0)
+        return {{}, EOFValidationError::invalid_first_section_type};
+
+    for (const auto& t : types)
+    {
+        if (t.max_stack_height > MAX_STACK_HEIGHT)
```
gumb0, line 188: According to EIP-5450 max stack height doesn't exceed 1023
```suggestion
        if (t.max_stack_height >= MAX_STACK_HEIGHT)
```

axic:
Is this deprecated by #563?

PR #500, opened 2022-08-26
Make Baseline default, select Advanced with "advanced" flag

chfast:


README.md
```diff
@@ -31,12 +36,6 @@ The codebase of _evmone_ is optimized to provide fast and efficient execution of
    and applied once per block during execution.
 3. Performs extensive and expensive bytecode analysis before execution.
```
hugo-dc, line 37: Specify we need the `advanced` flag in order to enable the Advanced interpreter (?)

chfast: Yes, I forgot to add it in the second commit. Good now?

hugo-dc: Looks good.

PR #504, opened 2022-09-06
Reset gas refund counter in ExecutionState

gumb0:


chfast (approved):

chfast:
We may want to release this as 0.9.1 so if you don't mind the trouble, you can create `release/0.9` branch at `v0.9.0` and merge it there. Also add changelog entry.

yperbasis (approved):

PR #506, opened 2022-09-21
rlp: More tests and experimental uint64 encoding

chfast:


test/unittests/state_rlp_test.cpp
```diff
@@ -123,3 +123,69 @@ TEST(state_rlp, encode_custom_struct_list)
...
+    EXPECT_EQ(rlp::encode(uint64_t{0xffffffffffffffff}), "88ffffffffffffffff"_hex);
+}
+
+/// The "custom" implementation of RLP encoding of uint64. It trims leading zero bytes and
+/// manually constructs bytes with variadic-length encoding.
+inline bytes rlp_encode_uint64(uint64_t x)
+{
+    static constexpr uint8_t ShortBase = 0x80;
+    if (x < ShortBase)  // Short form.
+        return bytes{(x != 0) ? static_cast<uint8_t>(x) : ShortBase};
```
axic, line 157: Why not?
```suggestion
        return bytes{(x == 0) ? ShortBase : static_cast<uint8_t>(x)};
```

chfast: No particular reason.

test/unittests/state_rlp_test.cpp
```diff
@@ -123,3 +123,69 @@ TEST(state_rlp, encode_custom_struct_list)
...
+
+    const auto bit_width = sizeof(x) * 8 - intx::clz(x);  // TODO(c++20): Use std::bit_width.
+    const auto byte_width = (bit_width + 7) / 8;
+    const auto trimmed_x = x << (intx::clz(x) & 0xfffffff8);  // Significant bytes moved to the top.
```
axic, line 161:
```suggestion
    const auto trimmed_x = x << (intx::clz(x) & 0x38);  // Significant bytes moved to the top.
```

Since `clz(uint64_t)` can only return 64, and you want to skip bottom 8-bits. The current `0xfff..` is more general, but needs perhaps better documentation.

axic (approved):
Looks cool, but maybe a bit more explainer comments could be useful.

test/unittests/state_rlp_test.cpp
```diff
@@ -123,3 +123,75 @@ TEST(state_rlp, encode_custom_struct_list)
...
+    EXPECT_EQ(rlp::encode(uint64_t{0xffffffffffffff}), "87ffffffffffffff"_hex);
+    EXPECT_EQ(rlp::encode(uint64_t{0x0100000000000000}), "880100000000000000"_hex);
+    EXPECT_EQ(rlp::encode(uint64_t{0xffffffffffffffff}), "88ffffffffffffffff"_hex);
+}
+
+inline bytes to_significant_be_bytes(uint64_t x)
+{
+    const auto bit_width = sizeof(x) * 8 - intx::clz(x);  // TODO(c++20): Use std::bit_width.
+    const auto byte_width = (bit_width + 7) / 8;
+    const auto leading_zero_bytes = intx::clz(x) & ~7u;  // Leading bits rounded down to 8x.
```
axic, line 155:
```suggestion
    const auto leading_zero_bits = intx::clz(x) & ~7u;  // Leading bits rounded down to 8x.
```

test/unittests/state_rlp_test.cpp
```diff
@@ -123,3 +123,75 @@ TEST(state_rlp, encode_custom_struct_list)
...
+/// The "custom" implementation of RLP encoding of uint64. It trims leading zero bytes and
+/// manually constructs bytes with variadic-length encoding.
+inline bytes rlp_encode_uint64(uint64_t x)
+{
+    static constexpr uint8_t ShortBase = 0x80;
+    if (x < ShortBase)  // Single-byte encoding.
+        return bytes{(x == 0) ? ShortBase : static_cast<uint8_t>(x)};
+
+    const auto b = to_significant_be_bytes(x);
+    return static_cast<uint8_t>(ShortBase + b.size()) + b;
```
axic, line 172: Maybe for readability this could be nice (especially given it uses auto for `b`):
```suggestion
    return bytes{static_cast<uint8_t>(ShortBase + b.size())} + b;
```

chfast: This will use different overload though. We are using `operator+(uint8_t, bytes)`.

axic:
@chfast wen merge?

chfast:
Rebased and modernized.

PR #512, opened 2022-10-05
statetest: Load tests from multiple dirs/files

### LINKED ISSUE #510: statetest: Load tests from multiple dirs / files
Currently the `evmone-statetest` tool takes one argument and loads JSON tests from the dir specified by this argument.
Add support for loading tests from multiple dirs (multiple args).
Also handle the case when the argument points to a file not a dir (useful for fuzzing tools or just executing single test).

Test names are created from the file test path as a relative path to the dir root. Having mutiple dirs/files we probably should first find the common root of these and then create test names from relative path to the common root.

Start with [`statetests_multiple_dirs`](https://github.com/ethereum/evmone/tree/statetests_multiple_dirs) branch.

JSzymanskiJS:
Allow passing multiple paths to evmone-statetest to load tests from.

Closes #510

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -349,11 +349,7 @@ jobs:
       - run:
           name: "State tests"
           working_directory: ~/build
-          command: bin/evmone-statetest ~/tests/GeneralStateTests
-      - run:
-          name: "State tests (legacy)"
-          working_directory: ~/build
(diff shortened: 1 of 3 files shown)


chfast:
Please rebase this PR.

chfast (approved):

PR #513, opened 2022-10-06
Migrate benchmarks to the JSON State Test format

### LINKED ISSUE #511: Migrate to benchmarks in JSON format
The existing benchmark suite has been converted to JSON State Test format and is available in https://github.com/ipsilon/evm-benchmarks.

- [ ] Add evm-benchmarks as git submodule replacing `test/benchmarks`.
- [ ] Update evmone-bench tool to load benchmarks from JSON files. You can reuse `StateTransitionTest` and `load_state_test()` from https://github.com/ethereum/evmone/blob/master/test/statetest/statetest_loader.cpp#L198.

miles170:
Closes #511.

I haven't checked gas used in every benchmark loop iteration because I want to make sure current changes won't break anything.
After that, I'll add a check for gas used.


test/bench/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ target_sources(
     bench.cpp
     helpers.hpp
     synthetic_benchmarks.cpp synthetic_benchmarks.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/../statetest/statetest_loader.cpp
```
chfast, line 13: The statetest loader should be a separate support library. We should make this change before this PR. Let me know if you want to work on this. If not I can make the change quickly.

test/statetest/statetest_loader.cpp
```diff
@@ -126,7 +142,10 @@ static void from_json(const json::json& j, TestMultiTransaction& o)
         o.max_gas_price = from_json<intx::uint256>(j.at("maxFeePerGas"));
         o.max_priority_gas_price = from_json<intx::uint256>(j.at("maxPriorityFeePerGas"));
     }
-    o.sender = from_json<evmc::address>(j.at("sender"));
+
+    if (j.contains("sender"))
```
chfast, line 146: I updated `evm-benchmarks` so this is not needed any more.

test/state/state.hpp
```diff
@@ -24,6 +25,21 @@ class State
...
+struct ClientInfo
+{
+    std::string comment;
+    std::string rpc_server_version;
+    std::string tool_version;
+    std::string lllc_version;
+    hash256 generated_test_hash;
+    std::string source;
+    hash256 source_hash;
+    std::unordered_map<uint64_t, std::string> labels;
```
chfast, line 42: I think we are not interested in other metadata than input labels for now so try to place this directly in `class StateTranstionTest`.  This can also be done separately from this PR for existing state test loader.

test/bench/CMakeLists.txt
miles170: I'll propose a new pr.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "evmc"]
 	path = evmc
 	url = https://github.com/ethereum/evmc
+[submodule "test/benchmarks"]
```
chfast, line 4: Add it with original name: "evm-benchmarks"

test/bench/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 
 add_executable(evmone-bench)
 target_include_directories(evmone-bench PRIVATE ${evmone_private_include_dir})
-target_link_libraries(evmone-bench PRIVATE evmone testutils evmc::loader benchmark::benchmark)
+target_link_libraries(evmone-bench PRIVATE evmone testutils evmc::loader evmone::statetestutils benchmark::benchmark)
```
chfast, line 7: Place it before `evmc::loader`.

test/statetest/statetest_loader.cpp
```diff
@@ -186,6 +186,11 @@ static void from_json(const json::json& j, StateTransitionTest& o)
 
     o.block = from_json<state::BlockInfo>(j_t.at("env"));
 
+    for (const auto& [j_key, j_value] : j_t.at("_info").at("labels").items())
```
chfast, line 189: "labels" are optional.

miles170:
Should I add more test cases to `GeneralStateTests`?

chfast:
> Should I add more test cases to `GeneralStateTests`(https://github.com/ethereum/tests)?

If you are asking in context of code coverage, then this is not needed. We don't run the tests yet. But on the other hand it is always welcomed to add new tests if you have any ideas.

chfast (approved):

PR #514, opened 2022-10-19
statetest: Extract test registration to separate function

chfast:


axic:
How is this working with empty test JSONs?

chfast:
> How is this working with empty test JSONs?

The `--gtest_list_tests` do not execute tests (i.e. the files are not even loaded).

PR #515, opened 2022-10-19
Bump minimum required GCC version to 10

chfast:
This removes the need to link stdc++fs in case <filesystem> is used.
This also allows using of std::span once C++20 is enabled.


chfast:
Included in #502.

PR #518, opened 2022-10-26
baseline: Better stack overflow/underflow checks

chfast:
Change stack overflow/underflow checks to stack pointers comparisons.
This gives better compiler optimizations. This gives around ~2% performance improvements.


rakita (approved):
lgtm, it just shuffles some stuff around

chfast:
2% man!

PR #522, opened 2022-11-02
Change has_const_gas_cost() to consteval

chfast:
This function is not meant to be executed at runtime.


axic:
What is stopping this PR?

chfast:
> What is stopping this PR?

The `consteval` support in clang is broken it seems.

PR #525, opened 2022-11-08
EIP-3860: Limit and meter initcode

yperbasis:
[EIP-3860](https://eips.ethereum.org/EIPS/eip-3860) is included into [Shanghai](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md). This PR implements Rules 3 & 4 of the [specification](https://eips.ethereum.org/EIPS/eip-3860#specification). Rules 1 & 2 should be implemented on the host side. See  https://github.com/torquem-ch/silkworm/pull/808 for an example of host implementation.

### DIFF
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -137,10 +137,21 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
     const auto init_code_offset = static_cast<size_t>(init_code_offset_u256);
     const auto init_code_size = static_cast<size_t>(init_code_size_u256);
 
+    if (state.rev >= EVMC_SHANGHAI && init_code_size > 0xC000)
+        return EVMC_SUCCESS;  // "Light" failure.
+
+    const auto init_code_words = num_words(init_code_size);
     if constexpr (Op == OP_CREATE2)
     {
-        const auto salt_cost = num_words(init_code_size) * 6;
+        const auto salt_cost = init_code_words * 6;
         if ((state.gas_left -= salt_cost) < 0)
+            return EVMC_OUT_OF_GAS;
+    }
+    if (state.rev >= EVMC_SHANGHAI)
+    {
+        // TODO: Merge two init code cost checks.
+        const auto init_code_cost = init_code_words * 2;
+        if ((state.gas_left -= init_code_cost) < 0)
             return EVMC_OUT_OF_GAS;
     }
 
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -207,7 +207,7 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
     assert(gas_left >= 0);
 
     const bytes_view code{result.output_data, result.output_size};
-    if (m_rev >= EVMC_SPURIOUS_DRAGON && code.size() > 0x6000)
+    if (m_rev >= EVMC_SPURIOUS_DRAGON && code.size() > max_code_size)
         return evmc::Result{EVMC_FAILURE};
 
     // Code deployment cost.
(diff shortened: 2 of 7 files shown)


lib/evmone/instructions_calls.cpp
```diff
@@ -125,15 +125,24 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
...
         salt = stack.pop();
-        auto salt_cost = num_words(static_cast<size_t>(init_code_size)) * 6;
+        auto salt_cost = init_code_words * 6;
         if ((state.gas_left -= salt_cost) < 0)
             return EVMC_OUT_OF_GAS;
     }
 
+    if (state.rev >= EVMC_SHANGHAI)
+    {
+        auto init_code_cost = init_code_words * 2;
```
chfast, line 141:
```suggestion
        const auto init_code_cost = init_code_words * 2;
```

lib/evmone/instructions_calls.cpp
```diff
@@ -125,15 +125,24 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
...
         return EVMC_OUT_OF_GAS;
 
+    const auto init_code_words = num_words(static_cast<size_t>(init_code_size));
+
     auto salt = uint256{};
     if constexpr (Op == OP_CREATE2)
     {
         salt = stack.pop();
-        auto salt_cost = num_words(static_cast<size_t>(init_code_size)) * 6;
+        auto salt_cost = init_code_words * 6;
```
chfast, line 134:
```suggestion
        const auto salt_cost = init_code_words * 6;
```

lib/evmone/instructions_calls.cpp
```diff
@@ -125,15 +125,24 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
...
         if ((state.gas_left -= salt_cost) < 0)
             return EVMC_OUT_OF_GAS;
     }
 
+    if (state.rev >= EVMC_SHANGHAI)
```
chfast, line 139: Does the order of the checks matter? It may be better to check this before the salt cost check.

lib/evmone/instructions_calls.cpp
yperbasis: done

lib/evmone/instructions_calls.cpp
yperbasis: done

lib/evmone/instructions_calls.cpp
yperbasis: I've swapped the checks

chfast (commented):
I'd add rule 3 here as well (before rule 4). I like to have "light" checks on the EVM side. This also helps against multiplication overflow. Any counter-arguments?

yperbasis:
> I'd add rule 3 here as well (before rule 4). I like to have "light" checks on the EVM side. This also helps against multiplication overflow. Any counter-arguments?

Initially I implemented rule 3 in evmone, but that deviates from geth's implementation https://github.com/ethereum/go-ethereum/pull/23847 (see `core/vm/evm.go`). Specifically, that leads to an incorrect nonce in the `createInitCodeSizeLimit_d1g0v0_Merge+3860` consensus test.

yperbasis:
I've updated this PR according to https://github.com/ethereum/EIPs/pull/6040.

gumb0 (dismissed):

chfast:
I messed up with GitHub here. Can you re-create the PR?

yperbasis:
> I messed up with GitHub here. Can you re-create the PR?

Sure. Re-created as PR #545.

PR #529, opened 2022-11-24
Implement EIP-663: Unlimited SWAP and DUP instructions

rodiazet:
https://eips.ethereum.org/EIPS/eip-663

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -820,6 +820,41 @@ inline void swap(StackTop stack) noexcept
     a[3] = t3;
 }
 
+inline code_iterator dupn(StackTop stack, ExecutionState& state, code_iterator pos) noexcept
+{
+    const auto n = pos[1] + 1;
+
+    const auto stack_size = &stack.top() - state.stack_space.bottom();
+
+    if (stack_size < n)
+    {
+        state.status = EVMC_STACK_UNDERFLOW;
+        return nullptr;
+    }
+
+    stack.push(stack[n - 1]);
+
+    return pos + 2;
+}
+
+inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator pos) noexcept
+{
+    const auto n = pos[1] + 1;
+
+    const auto stack_size = &stack.top() - state.stack_space.bottom();
+
+    if (stack_size <= n)
+    {
+        state.status = EVMC_STACK_UNDERFLOW;
+        return nullptr;
+    }
+
+    // TODO: This may not be optimal, see instr::core::swap().
+    std::swap(stack.top(), stack[n]);
+
+    return pos + 2;
+}
+
 template <size_t NumTopics>
 inline evmc_status_code log(StackTop stack, ExecutionState& state) noexcept
 {
--- a/lib/evmone/instructions_opcodes.hpp
+++ b/lib/evmone/instructions_opcodes.hpp
@@ -152,6 +152,9 @@ enum Opcode
     OP_LOG3 = 0xa3,
     OP_LOG4 = 0xa4,
 
+    OP_DUPN = 0xb5,
+    OP_SWAPN = 0xb6,
+
     OP_CREATE = 0xf0,
     OP_CALL = 0xf1,
     OP_CALLCODE = 0xf2,
--- a/test/unittests/eof_validation_test.cpp
+++ b/test/unittests/eof_validation_test.cpp
@@ -219,7 +219,8 @@ TEST(eof_validation, EOF1_terminating_instructions)
     {
         const auto& op_traits = traits[opcode];
         // Skip undefined opcodes.
-        if (op_traits.name == nullptr)
+        // TODO: iterate over all EOF revisions.
+        if (op_traits.name == nullptr || op_traits.since == EVMC_CANCUN)
             continue;
 
         bytes code{static_cast<uint8_t>(opcode) + bytes(op_traits.immediate_size, 0)};
--- a/test/unittests/instructions_test.cpp
+++ b/test/unittests/instructions_test.cpp
@@ -55,6 +55,8 @@ constexpr void validate_traits_of() noexcept
     // immediate_size
     if constexpr (Op >= OP_PUSH1 && Op <= OP_PUSH32)
         static_assert(tr.immediate_size == Op - OP_PUSH1 + 1);
+    else if constexpr (Op == OP_DUPN || Op == OP_SWAPN)
+        static_assert(tr.immediate_size == 1);
     else
         static_assert(tr.immediate_size == 0);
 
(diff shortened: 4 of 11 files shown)


lib/evmone/instructions_opcodes.hpp
```diff
@@ -152,6 +152,9 @@ enum Opcode
     OP_LOG3 = 0xa3,
     OP_LOG4 = 0xa4,
 
+    OP_DUPN = 0xb2,
+    OP_SWAPN = 0xb3,
```
axic, line 156: Let's go for 0xb3 + 0xb4 to match the EIP.

lib/evmone/instructions.hpp
```diff
@@ -820,6 +820,40 @@ inline void swap(StackTop stack) noexcept
...
+    }
+
+    stack.push(stack[n]);
+
+    return pos + 2;
+}
+
+inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator pos) noexcept
+{
+    const auto n = pos[1] + 17;
```
axic, line 842: The EIP doesn't have this yet and people seem to lean against now.

rodiazet: Done

lib/evmone/instructions_opcodes.hpp
rodiazet: Done

axic: Sorry to ask again, but to avoid conflicts with EIP-1153, let's use 0xb5 and 0xb6. The EIP has been updated.

lib/evmone/instructions.hpp
```diff
@@ -820,6 +820,40 @@ inline void swap(StackTop stack) noexcept
...
+
+    const auto stack_size = &stack.top() - state.stack_space.bottom();
+
+    if (stack_size <= n)
+    {
+        state.status = EVMC_STACK_UNDERFLOW;
+        return nullptr;
+    }
+
+    stack.push(stack[n]);
```
axic, line 835: I think this misses the `< 1024` check. `push` does [not seem enforce the 1024 limit](https://github.com/ethereum/evmone/blob/master/lib/evmone/instructions.hpp#L38).

chfast: The stack overflow check is done automatically by the interpreter. This information is in the instruction traits.

test/unittests/eof_validation_test.cpp
```diff
@@ -230,6 +230,6 @@ TEST(eof_validation, EOF1_terminating_instructions)
                                    opcode == OP_INVALID || opcode == OP_SELFDESTRUCT) ?
                                    EOFValidationError::success :
                                    EOFValidationError::missing_terminating_instruction);
-        EXPECT_EQ(validate_eof(container), expected) << hex(code);
+        EXPECT_EQ(validate_eof(container, EVMC_CANCUN), expected) << hex(code);
```
gumb0, line 233: We still want to test `validate_eof` for Shanghai, better skip opcodes that are defined since Cancun on line 222.

It could be also a loop over revisions starting from Shanghai, but we want to drop this terminating instruction requirement anyway, so I'd say just "TODO iterate for all EOF revisions" is enough too.

test/unittests/instructions_test.cpp
```diff
@@ -138,6 +150,11 @@ TEST(instructions, compare_with_evmc_instruction_names)
     const auto* evmc_tbl = evmc_get_instruction_names_table(EVMC_MAX_REVISION);
     for (size_t i = 0; i < instr::traits.size(); ++i)
     {
+        // Skip DUPN and SWAPN for Cancun. They are not defined in evmc
+        // TODO: Define DUPN and SWAPN in evmc
+        if (instr::traits[i].since == EVMC_CANCUN &&
```
gumb0, line 155: Nit: checking for Cancun is redundant here, DUPN/SWAPN are only since Cancun always.

lib/evmone/instructions_opcodes.hpp
```diff
@@ -152,6 +152,9 @@ enum Opcode
     OP_LOG3 = 0xa3,
     OP_LOG4 = 0xa4,
 
+    OP_DUPN = 0xb3,
+    OP_SWAPN = 0xb4,
```
gumb0, line 156:
```suggestion
    OP_DUPN = 0xb5,
    OP_SWAPN = 0xb6,
```

lib/evmone/instructions.hpp
```diff
@@ -820,6 +820,40 @@ inline void swap(StackTop stack) noexcept
     a[3] = t3;
 }
 
+inline code_iterator dupn(StackTop stack, ExecutionState& state, code_iterator pos) noexcept
+{
+    const auto n = pos[1];
```
gumb0, line 825: Looks correct, but a bit confusing when you compare it to EIP spec.
Maybe `n = pos[1] + 1`, `if (stack_size < n)`, `stack.push(stack[n -1])` ?

test/unittests/evm_test.cpp
```diff
@@ -153,6 +153,173 @@ TEST_P(evm, swapsn_push)
...
+
+    execute(full_stack_code + OP_POP + OP_DUPN + "ff" + ret_top());
+    EXPECT_STATUS(EVMC_SUCCESS);
+    EXPECT_OUTPUT_INT(257);
+
+    execute(full_stack_code + OP_POP + OP_DUPN + "fe" + ret_top());
+    EXPECT_STATUS(EVMC_SUCCESS);
+    EXPECT_OUTPUT_INT(256);
+
+    execute(full_stack_code + OP_DUPN + "fe" + OP_DUPN + "ff");
```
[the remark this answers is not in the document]
gumb0: Ah I see OP_POP was needed because of `ret_top()` in the end

test/unittests/evm_test.cpp
```diff
@@ -153,6 +153,173 @@ TEST_P(evm, swapsn_push)
...
+    if (evm::isAdvanced())
+        return;
+
+    rev = EVMC_CANCUN;
+
+    auto pushes = bytecode{};
+    for (uint64_t i = 1; i <= 20; ++i)
+        pushes += push(i);
+
+    execute(pushes + OP_SWAPN + "00" + ret_top());
```
gumb0, line 196: You don't really check the other value that was swapped :smile: 
Maybe add another test with DUPN like
```cpp
    execute(pushes + OP_SWAPN + "00" + OP_DUPN + "01" + ret_top());
```

lib/evmone/instructions_opcodes.hpp
rodiazet: Done

lib/evmone/instructions_opcodes.hpp
rodiazet: Done

lib/evmone/instructions.hpp
rodiazet: Done

test/unittests/evm_test.cpp
rodiazet: Done.

gumb0 (approved):

chfast (approved):

PR #533, opened 2022-11-29
Remove dependency on EVMC instructions

rodiazet:


test/unittests/evm_test.cpp
```diff
@@ -8,6 +8,7 @@
 using namespace evmc::literals;
 using namespace intx;
 using evmone::test::evm;
+using namespace evmone;
```
chfast, line 11: Move the line up to be ordered alphabetically.

lib/evmone/instructions.hpp
```diff
@@ -918,11 +918,11 @@ inline StopToken selfdestruct(StackTop stack, ExecutionState& state) noexcept
 
 /// Maps an opcode to the instruction implementation.
 ///
-/// The set of template specializations which map opcodes `Op` to the function
+/// The set of template specializations which map opcodes `Opcode` to the function
```
chfast, line 921:
```suggestion
/// The set of template specializations which map opcodes `Op` to the function
```

test/utils/bytecode.hpp
```diff
@@ -17,6 +18,8 @@ inline bytecode push(uint64_t n);
 inline bytecode push(evmc::address addr);
 inline bytecode push(evmc::bytes32 bs);
 
+using namespace evmone;
```
chfast, line 21:
```suggestion
using enum evmone::Opcode;
```

chfast (changes requested):

chfast (changes requested):
In CMake: remove `evmc::instructions` dependency from `evmone` and `testutils`.

test/utils/bytecode.hpp
rodiazet: This change will require to change `Opcode opcode` to `evmone::Opcode opcode` in `Opcode` type param declaration.

lib/evmone/baseline.cpp
```diff
@@ -7,8 +7,8 @@
 #include "eof.hpp"
 #include "execution_state.hpp"
 #include "instructions.hpp"
+#include "instructions_opcodes.hpp"
```
chfast, line 10: Not needed. `instructions.hpp` has `Opcode` included already.

test/utils/bytecode.hpp
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 #pragma once
 
+#include "evmone/instructions_opcodes.hpp"
```
chfast, line 6: Not needed.

chfast (changes requested):

lib/evmone/baseline.cpp
chfast: You missed this one.

chfast (approved):

PR #534, opened 2022-11-29
Get instruction names from instr::traits[]

chfast:
This removes the dependency on `evmc::instructions` for getting instruction names for tracing and testing.


lib/evmone/tracing.cpp
```diff
@@ -25,21 +26,18 @@ class HistogramTracer : public Tracer
...
 
     std::stack<Context> m_contexts;
     std::ostream& m_out;
 
     void on_execution_start(
-        evmc_revision rev, const evmc_message& msg, bytes_view code) noexcept override
+        evmc_revision /*rev*/, const evmc_message& msg, bytes_view code) noexcept override
     {
-        m_contexts.emplace(msg.depth, code.data(), evmc_get_instruction_names_table(rev));
+        m_contexts.emplace(msg.depth, code.data());
```
rodiazet, line 40: In previous version we took only instructions from current revision. In the new one we take all instructions from all revisions. Not sure if it's ok.

chfast: Yes, this is right. We lost some precision, e.g. some not-yet-defined instructions may be printed by name. I think this is fine for now. Later we can create tables with revision dimension (similar to EVMC). There is TODO added for this.

rodiazet: Ok. Fine for me.

rodiazet (approved):

PR #539, opened 2022-12-14
ci: Windows build via circleci

rodiazet:


circle.yml
```diff
@@ -505,6 +545,12 @@ workflows:
           filters:
             tags:
               only: /.*/
+      - release-windows:
+          filters:
+            branches:
+              ignore: /.*/
```
chfast, line 551: Branches should not be ignored. You can also enabled builds of release-macos.

chfast (approved):

circle.yml
rodiazet: OK.

PR #542, opened 2022-12-28
Refactor CALL/CREATE instructions

chfast:
- Perform uint256 → size_t casts of memory offsets and sizes in single
place when it is safe to do so.
- Assume failure of CALL/CREATE by setting result to 0 and clearing
the return buffer before performing any checks.


rodiazet (approved):

lib/evmone/instructions_calls.cpp
```diff
@@ -82,20 +89,18 @@ evmc_status_code call_impl(StackTop stack, ExecutionState& state) noexcept
         state.gas_left += 2300;
     }
 
-    state.return_data.clear();
```
axic, line 85: This wasn't tested in the ethereum/tests?

chfast: What do you mean by "tested"? This is non-functional change. In particular, the return buffer clear was moved up to be consistent with create.

PR #544, opened 2023-01-04
ci: Add 32-bit Windows builds

axic:
Depends on https://github.com/chfast/intx/pull/283.

### DIFF
--- a/cmake/Hunter/config.cmake
+++ b/cmake/Hunter/config.cmake
@@ -8,3 +8,10 @@ hunter_cmake_args(
     ethash
     CMAKE_ARGS -DETHASH_BUILD_ETHASH=OFF
 )
+
+hunter_config(
+    intx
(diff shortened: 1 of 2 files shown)


cmake/Hunter/config.cmake
```diff
@@ -8,3 +8,10 @@ hunter_cmake_args(
     ethash
     CMAKE_ARGS -DETHASH_BUILD_ETHASH=OFF
 )
+
+hunter_config(
+    intx
+    VERSION 0.9.1
+    URL https://github.com/chfast/intx/archive/42750ef8e7575bf91bb0185dfe509352ba633554.tar.gz
```
axic, line 15: Should wait for a release.

chfast (approved):
It looks like windows builds work sequentially so this probably don't make sense long term.

PR #552, opened 2023-01-18
Add evmone t8n tool

chfast:
The t8n is "state transition" CLI interface used to check and generate
JSON tests. It reads and outputs specified JSON files.
See https://ethereum-tests.readthedocs.io/en/develop/t8ntool-ref.html.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "state.hpp"
+#include "errors.hpp"
 #include "host.hpp"
 #include <evmone/evmone.h>
 #include <evmone/execution_state.hpp>
@@ -53,32 +54,33 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
 /// Validates transaction and computes its execution gas limit (the amount of gas provided to EVM).
 /// @return  Non-negative execution gas limit for valid transaction
 ///          or negative value for invalid transaction.
-int64_t validate_transaction(const Account& sender_acc, const BlockInfo& block,
-    const Transaction& tx, evmc_revision rev) noexcept
+std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
+    const BlockInfo& block, const Transaction& tx, evmc_revision rev) noexcept
 {
     if (rev < EVMC_LONDON && tx.kind == Transaction::Kind::eip1559)
-        return -1;
+        return make_error_code(TX_TYPE_NOT_SUPPORTED);
 
     if (rev < EVMC_BERLIN && !tx.access_list.empty())
-        return -1;
+        return make_error_code(TX_TYPE_NOT_SUPPORTED);
 
     if (tx.max_priority_gas_price > tx.max_gas_price)
-        return -1;  // Priority gas price is too high.
+        return make_error_code(TIP_GT_FEE_CAP);  // Priority gas price is too high.
 
     if (tx.gas_limit > block.gas_limit)
-        return -1;
+        return make_error_code(GAS_LIMIT_REACHED);
 
     if (rev >= EVMC_LONDON && tx.max_gas_price < block.base_fee)
-        return -1;
+        return make_error_code(FEE_CAP_LESS_THEN_BLOCKS);
 
     if (!sender_acc.code.empty())
-        return -1;  // Origin must not be a contract (EIP-3607).
+        return make_error_code(SENDER_NOT_EOA);  // Origin must not be a contract (EIP-3607).
 
     if (sender_acc.nonce == Account::NonceMax)
-        return -1;
+        return make_error_code(NONCE_HAS_MAX_VALUE);
 
+    // initcode size is limited by EIP-3860.
     if (rev >= EVMC_SHANGHAI && !tx.to.has_value() && tx.data.size() > max_initcode_size)
-        return -1;  // initcode size is limited by EIP-3860.
+        return make_error_code(INIT_CODE_SIZE_LIMIT_EXCEEDED);
 
     // Compute and check if sender has enough balance for the theoretical maximum transaction cost.
     // Note this is different from tx_max_cost computed with effective gas price later.
@@ -86,9 +88,13 @@ int64_t validate_transaction(const Account& sender_acc, const BlockInfo& block,
     if (const auto tx_cost_limit_512 =
             umul(intx::uint256{tx.gas_limit}, tx.max_gas_price) + tx.value;
         sender_acc.balance < tx_cost_limit_512)
-        return -1;
+        return make_error_code(INSUFFICIENT_FUNDS);
 
-    return tx.gas_limit - compute_tx_intrinsic_cost(rev, tx);
+    const auto intrinsic_cost = compute_tx_intrinsic_cost(rev, tx);
+    if (intrinsic_cost > tx.gas_limit)
+        return make_error_code(INTRINSIC_GAS_TOO_LOW);
+
+    return tx.gas_limit - intrinsic_cost;
 }
 
 evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) noexcept
@@ -110,18 +116,21 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 }
 }  // namespace
 
-std::optional<std::vector<Log>> transition(
+std::variant<TransactionReceipt, std::error_code> transition(
     State& state, const BlockInfo& block, const Transaction& tx, evmc_revision rev, evmc::VM& vm)
 {
     auto& sender_acc = state.get(tx.sender);
-    const auto execution_gas_limit = validate_transaction(sender_acc, block, tx, rev);
-    if (execution_gas_limit < 0)
+    const auto validation_result = validate_transaction(sender_acc, block, tx, rev);
+
+    if (holds_alternative<std::error_code>(validation_result))
     {
         // Pre EIP-158 coinbase has to be touched also for invalid tx.
         if (rev <= EVMC_TANGERINE_WHISTLE)
             state.touch(block.coinbase);
-        return {};
+        return get<std::error_code>(validation_result);
     }
+
+    const auto execution_gas_limit = get<int64_t>(validation_result);
 
     const auto base_fee = (rev >= EVMC_LONDON) ? block.base_fee : 0;
     assert(tx.max_gas_price >= base_fee);                   // Checked at the front.
@@ -167,6 +176,6 @@ std::optional<std::vector<Log>> transition(
         return acc.destructed || (rev >= EVMC_SPURIOUS_DRAGON && acc.erasable && acc.is_empty());
     });
 
-    return host.take_logs();
+    return TransactionReceipt{gas_used, host.take_logs()};
 }
 }  // namespace evmone::state
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -74,6 +74,9 @@ state::BlockInfo from_json<state::BlockInfo>(const json::json& j);
 template <>
 state::State from_json<state::State>(const json::json& j);
 
+template <>
+state::Transaction from_json<state::Transaction>(const json::json& j);
+
 StateTransitionTest load_state_test(const fs::path& test_file);
 
 void run_state_test(const StateTransitionTest& test, evmc::VM& vm);
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -4,9 +4,11 @@
 #pragma once
 
 #include "../state/state.hpp"
+#include <nlohmann/json.hpp>
 #include <filesystem>
 
 namespace fs = std::filesystem;
+namespace json = nlohmann;
 
 namespace evmone::test
 {
@@ -62,6 +64,12 @@ struct StateTransitionTest
     std::vector<Case> cases;
     std::unordered_map<uint64_t, std::string> input_labels;
 };
+
+template <typename T>
+T from_json(const json::json& j) = delete;
+
+template <>
+state::BlockInfo from_json<state::BlockInfo>(const json::json& j);
 
 StateTransitionTest load_state_test(const fs::path& test_file);
 
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -81,4 +81,18 @@ StateTransitionTest load_state_test(const fs::path& test_file);
 
 void run_state_test(const StateTransitionTest& test, evmc::VM& vm);
 
+/// Converts an integer to hex string representation with 0x prefix.
+///
+/// This handles also builtin types like uint64_t. Not optimal but works for now.
+inline std::string hex0x(const intx::uint256& v)
+{
+    return "0x" + intx::hex(v);
+}
+
+/// Encodes bytes as hex with 0x prefix.
+inline std::string hex0x(const bytes_view& v)
+{
+    return "0x" + evmc::hex(v);
+}
+
 }  // namespace evmone::test
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -10,6 +10,10 @@ namespace fs = std::filesystem;
 
 namespace evmone::test
 {
+
+/// Translates tests fork name to EVM revision
+evmc_revision to_rev(std::string_view s);
+
 struct TestMultiTransaction : state::Transaction
 {
     struct Indexes
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -71,6 +71,9 @@ T from_json(const json::json& j) = delete;
 template <>
 state::BlockInfo from_json<state::BlockInfo>(const json::json& j);
 
+template <>
+state::State from_json<state::State>(const json::json& j);
+
 StateTransitionTest load_state_test(const fs::path& test_file);
 
 void run_state_test(const StateTransitionTest& test, evmc::VM& vm);
--- a/test/statetest/statetest_loader.cpp
+++ b/test/statetest/statetest_loader.cpp
@@ -79,6 +79,26 @@ state::BlockInfo from_json<state::BlockInfo>(const json::json& j)
     };
 }
 
+template <>
+state::State from_json<state::State>(const json::json& j)
+{
+    state::State o;
+    for (const auto& [j_addr, j_acc] : j.items())
+    {
+        auto& acc = o.insert(from_json<address>(j_addr),
+            {.nonce = from_json<uint64_t>(j_acc.at("nonce")),
+                .balance = from_json<intx::uint256>(j_acc.at("balance")),
+                .code = from_json<bytes>(j_acc.at("code"))});
+
+        for (const auto& [j_key, j_value] : j_acc.at("storage").items())
+        {
+            const auto value = from_json<bytes32>(j_value);
+            acc.storage.insert({from_json<bytes32>(j_key), {.current = value, .original = value}});
+        }
+    }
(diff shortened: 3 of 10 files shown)


chfast:
First example to run:

- geth
  ```
  retesteth -t 'GeneralStateTests/stExample' -- --clients t8ntool --testpath ~/Projects/ethereum/tests --singletest add11 --singlenet Berlin --verbosity 6
  ```

- evmone
  ```
  retesteth -t 'GeneralStateTests/stExample' -- --datadir ~/Projects/ethereum/evmone/config/retesteth --clients evmone --testpath ~/Projects/ethereum/tests --singletest add11 --singlenet Berlin --verbosity 6
  ```

chfast:
The example isn't working currently because evmone does not produce expected transaction receipts (empty array in the result). The receipt is match with a transaction by transaction hash which evmone is not able to compute yet.

test/statetest/secret_sender_map.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include "secret_sender_map.hpp"
+#include <cassert>
+#include <map>
+
+namespace evmone::test
+{
+using namespace evmc;
+
+evmc::address sender_address(const bytes32& secret_key)
```
axic, line 13: Is this `private_key` -> `address` lookup table?

test/statetest/statetest.hpp
```diff
@@ -59,8 +63,59 @@ struct StateTransitionTest
...
+}
+
+inline size_t countl_zero_bytes(uint64_t x)
+{
+    auto upper = uint32_t(x >> 32);
+    auto lower = uint32_t(x & 0xFFFFFFFF);
+    return upper ? countl_zero_bytes(upper) : 4 + countl_zero_bytes(lower);
+}
+
+inline std::string hex(uint64_t v)
```
axic, line 105: I'd call this something like `shortest_hex` or something similar to signal intent.

test/statetest/statetest_loader.cpp
```diff
@@ -11,11 +12,6 @@ namespace evmone::test
 namespace json = nlohmann;
 using evmc::from_hex;
 
-namespace
```
axic, line 14: Shouldn't these stay in the anonymous namespace?

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <string_view>
+
+namespace fs = std::filesystem;
+namespace json = nlohmann;
+using namespace evmone;
+
+namespace evmone::state
```
axic, line 15: Do we need this available there or can this be anonymous?

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,174 @@
...
+            }
+
+            j_result["logsHash"] = "0x" + hex(keccak256(rlp::encode(txs_logs)));
+            j_result["stateRoot"] = "0x" + hex(state::mpt_hash(state.get_accounts()));
+        }
+
+        j_result["logsBloom"] = "0x" + std::string(512, '0');
+        j_result["receiptsRoot"] =
+            "0x0000000000000000000000000000000000000000000000000000000000000000";
+        j_result["txRoot"] = "0x0000000000000000000000000000000000000000000000000000000000000000";
```
axic, line 154: Could we have an overloaded version of `hex` to support various types, and then use something like `hex(intx::uint256{0})`?

axic: Or perhaps have these constants pre-defined at the top of the file and reused multiple times (e.g. `NULL_HEXSTRING_160`, `NULL_HEXSTRING_256`, `NULL_HEXSTRING_512`)

test/statetest/statetest_loader.cpp
rodiazet: No, because `from_json` is now also defined in header.

test/t8n/t8n.cpp
rodiazet: Unfortunately it has to be there. Similar to `statetest_runner.cpp`. In other case template functions encode instantiation fails.

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+        std::ofstream{output_dir / output_result_file} << j_result;
+        std::cerr << "---\nRESULT:\n" << std::setw(2) << j_result << "\n";
+
+        json::json j_alloc;
+        for (const auto& [addr, acc] : state.get_accounts())
+        {
+            j_alloc["0x" + hex(addr)]["nonce"] = "0x" + test::utils::hex(acc.nonce);
```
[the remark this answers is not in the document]
rodiazet: I introduced changes to make it less confusing.

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,172 @@
...
+                for (auto& j_tx : j_txs)
+                {
+                    state::Transaction tx = test::from_json<state::Transaction>(j_tx);
+                    auto const [tx_logs, gas_used] = state::transition(state, block, tx, rev, vm);
+                    if (tx_logs.has_value())
+                    {
+                        txs_logs.insert(txs_logs.begin(), (*tx_logs).begin(), (*tx_logs).end());
+                        auto& j_receipt = j_result["receipts"][j_result["receipts"].size()];
+                        j_receipt["transactionHash"] = j_tx["hash"];
+                        j_receipt["gasUsed"] = tu::shortest_hex(uint64_t(gas_used));
```
axic, line 117: What if `int64_t gas_used` is negative (due to a failure)?

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,172 @@
...
+                        j_receipt["cumulativeGasUsed"] = j_receipt["gasUsed"];
+
+                        j_receipt["blockHash"] = NULL_HEXSTRING_32;
+                        j_receipt["contractAddress"] = NULL_HEXSTRING_20;
+                        j_receipt["logsBloom"] = NULL_HEXSTRING_256;
+                        j_receipt["logs"] = json::json::array();  // FIXME: Add to_json<state:Log>
+                        j_receipt["root"] = "";
```
axic, line 124: Is this really expected to be empty?

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,172 @@
...
+        // Print out current state to outAlloc file
+        json::json j_alloc;
+        for (const auto& [addr, acc] : state.get_accounts())
+        {
+            j_alloc[tu::hex(addr)]["nonce"] = test::utils::shortest_hex(acc.nonce);
+            for (const auto& [key, val] : acc.storage)
+                if (!is_zero(val.current))
+                    j_alloc[tu::hex(addr)]["storage"][tu::hex(key)] = tu::hex(val.current);
+
+            j_alloc[tu::hex(addr)]["code"] = tu::hex({acc.code.data(), acc.code.size()});
```
axic, line 160: This has no `0x` prefix?

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,172 @@
...
+namespace evmone::state
+{
+/// Defines how to RLP-encode a Log. This is only needed to compute "logs hash".
+inline bytes rlp_encode(const Log& log)
+{
+    return rlp::encode_tuple(log.addr, log.topics, log.data);
+}
+}  // namespace evmone::state
+
+int main(int argc, const char* argv[])
```
axic, line 29: We should have a unit test for this (mostly for checking the input -> output mapping is unchanged). Question is: is it better to use a CLI test or make this into an internal library first?

test/t8n/t8n.cpp
rodiazet: It's a minimal working version. A lot of fields are required by retesteth but never used.

test/t8n/t8n.cpp
rodiazet: I would prefer the second. @chfast What do you think?

test/t8n/t8n.cpp
axic: One more idea: we could have an overloaded helper `hex` defined in this file at the top for each type we care of, so that the body here can just use `hex()` solely. This way we can inspect what types are used and how they get translated.

rodiazet: But if these helper needs to use hex impl from different libs like evmc it will be a mess with specifying which one we want to use.

axic: I meant something like this:
```cpp
namespace {
inline bytes hex(string_view s) {
    return evmc::hex(s);
}
inline bytes hex(intx::u256 v) {
    return intx::hex(v);
}
}
```

rodiazet: Sure I understand but we will have to remove a couple of `using namespace evmc` etc.

test/statetest/statetest.hpp
```diff
@@ -59,8 +63,66 @@ struct StateTransitionTest
...
+    const auto prefix = std::string(add_0x ? "0x" : "");
+    size_t num_lz_bytes = countl_zero_bytes(v);
+
+    if (num_lz_bytes == 8)
+        return prefix + "0";
+
+    std::string str;
+
+    for (auto i = num_lz_bytes; i < 8; ++i)
+        str += evmc::hex(static_cast<uint8_t>(v >> 8 * (8 - i - 1)));
```
axic, line 116: Could also just translate it to string first and then doing a substring to the first non-`0` character.

test/statetest/statetest.hpp
```diff
@@ -59,8 +63,66 @@ struct StateTransitionTest
...
+    auto lower = uint32_t(x & 0xFFFFFFFF);
+    return upper ? countl_zero_bytes(upper) : 4 + countl_zero_bytes(lower);
+}
+
+inline std::string shortest_hex(uint64_t v, bool add_0x = true)
+{
+    const auto prefix = std::string(add_0x ? "0x" : "");
+    size_t num_lz_bytes = countl_zero_bytes(v);
+
+    if (num_lz_bytes == 8)
```
axic, line 110: This could be done with `if (v == 0)`.

test/statetest/statetest.hpp
axic:
```cpp
inline std::string shortest_hex(uint64_t v, bool add_prefix = true)
{
    if (v == 0) {
        return add_prefix ? "0x0" : "0";
    }

    auto value = hex(v, false);
    // String leading zeroes.
    if (size_t pos = value.find_first_not_of("0"); pos != string::npos) {
        value = value.substr(pos);
    }
    return (add_prefix ? "0x" : "") + value;
}
```

test/statetest/statetest_loader.cpp
```diff
@@ -69,19 +65,89 @@ state::AccessList from_json<state::AccessList>(const json::json& j)
     return o;
 }
 
+inline uint64_t calculate_current_base_fee_eip1559(
+    uint64_t parent_gas_used, uint64_t parent_gas_limit, uint64_t parent_base_fee)
+{
+    static const auto BASE_FEE_MAX_CHANGE_DENOMINATOR = 8;
+    static const auto ELASTICITY_MULTIPLIER = 2;
+
+    uint64_t base_fee = 0;
```
axic, line 74: Is 64-bit precision enough for these? Someone should double check EIP-1559.

axic:
Should remove `.config.swp` before merging. Also why is there a need to duplicate the config files, aren't they part of reteseth?

test/statetest/statetest_loader.cpp
rodiazet: TODO added.

rodiazet:
> Should remove `.config.swp` before merging. Also why is there a need to duplicate the config files, aren't they part of reteseth?

Definitely configs should be moved to retesteth.

test/statetest/statetest.hpp
```diff
@@ -59,8 +63,54 @@ struct StateTransitionTest
...
+inline std::string shortest_hex(uint64_t v, bool add_prefix = true)
+{
+    if (v == 0)
+    {
+        return add_prefix ? "0x0" : "0";
+    }
+
+    std::string value;
+    for (auto i = 0; i < 8; ++i)
+        value += evmc::hex(static_cast<uint8_t>(v >> 8 * (8 - i - 1)));
```
axic, line 94: Hm, can't this actually be done with?
```suggestion
    value = intx::hex(intx::uint256{v});
```

rodiazet: Done

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "../statetest/statetest.hpp"
+#include <evmone/evmone.h>
+#include <nlohmann/json.hpp>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <string_view>
+
+namespace fs = std::filesystem;
+namespace json = nlohmann;
```
axic, line 12: Why not?
```suggestion
namespace json = nlohmann::json;
```

Would remove the `json::json::...` cases.

axic: Ah ok, this is consistent with `statetest_loader`, can change later then.

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,177 @@
...
+            {
+                j_result["receipts"] = json::json::array();
+                j_result["rejected"] = json::json::array();
+                auto idx = 0;
+                for (auto& j_tx : j_txs)
+                {
+                    state::Transaction tx = test::from_json<state::Transaction>(j_tx);
+                    auto const result = state::transition(state, block, tx, rev, vm);
+                    const auto receipt = get_if<state::TransactionReceipt>(&result);
+                    if (receipt != nullptr)
```
axic, line 113: I'd swap and have the failure case first, as that is shorter.

test/statetest/statetest.hpp
```diff
@@ -59,8 +63,52 @@ struct StateTransitionTest
...
+state::Transaction from_json<state::Transaction>(const json::json& j);
+
 StateTransitionTest load_state_test(const fs::path& test_file);
 
 void run_state_test(const StateTransitionTest& test, evmc::VM& vm);
 
+namespace utils
+{
+
+inline std::string shortest_hex(uint64_t v, bool add_prefix = true)
```
chfast, line 85: Is this really needed? I believe `intx::hex` is already producing the "shorted" hex as expected for any integer.

test/statetest/statetest.hpp
```diff
@@ -59,8 +63,52 @@ struct StateTransitionTest
...
+    }
+    return (add_prefix ? "0x" : "") + value;
+}
+
+inline std::string hex(bytes_view bs, bool add_prefix = true)
+{
+    return (add_prefix ? "0x" : "") + evmc::hex(bs);
+}
+
+inline std::string hex(const intx::uint<256>& v, bool add_prefix = true)
```
chfast, line 107: It looks we are just wrapping existing `hex` functions to add a prefix. How about a template:
```cpp
template<typename T>
auto hex0x(const T& v)
{
    return "0x" + hex(v);
}
```

rodiazet: done

test/statetest/statetest.hpp
rodiazet: you are right. Fixed.

axic: In fairness, we slowly arrived at the point of using `intx::hex()` which then allowed this further simplification 😅

lib/evmone/baseline.cpp
```diff
@@ -356,6 +356,11 @@ evmc_result execute(const VM& vm, ExecutionState& state, const CodeAnalysis& ana
 evmc_result execute(evmc_vm* c_vm, const evmc_host_interface* host, evmc_host_context* ctx,
     evmc_revision rev, const evmc_message* msg, const uint8_t* code, size_t code_size) noexcept
 {
+    if (rev >= EVMC_SHANGHAI && is_eof_code({code, code_size}) &&
```
chfast, line 359: Why is this needed?

lib/evmone/errors.hpp
```diff
@@ -0,0 +1,80 @@
+// evmone: Fast Ethereum Virtual Machine implementation
```
chfast, line 1: This file should go to test/state.

test/state/state.cpp
```diff
@@ -53,42 +54,48 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
...
-        return -1;  // Origin must not be a contract (EIP-3607).
+        return make_error_code(SENDER_NOT_EOA);  // Origin must not be a contract (EIP-3607).
 
     if (sender_acc.nonce == Account::NonceMax)
-        return -1;
+        return make_error_code(NONCE_HAS_MAX_VALUE);
 
     if (rev >= EVMC_SHANGHAI && !tx.to.has_value() && tx.data.size() > max_initcode_size)
-        return -1;  // initcode size is limited by EIP-3860.
+        return make_error_code(INIT_CODE_SIZE_LIMIT_EXCEEDED);  // initcode size is limited by
```
chfast, line 82: Move comment before `if`.

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,171 @@
...
+        {
+            const auto j = json::json::parse(std::ifstream{env_file});
+            block = test::from_json<state::BlockInfo>(j);
+        }
+
+        auto state = pre_state;
+
+        json::json j_result;
+        // FIXME: Calculate difficulty properly
+        j_result["currentDifficulty"] = "0x20000";
```
chfast, line 86: Is this right?

config/retesteth/evmone/config
```diff
@@ -0,0 +1,239 @@
+{
+    "name" : "evmone",
+    "socketType" : "tranition-tool",
+    "socketAddress" : "start.sh",
+    "transactionsAsJson" : true,
+    "checkLogsHash" : false,
```
chfast, line 6: Do you want to try this?

lib/evmone/baseline.cpp
rodiazet: For code equals `0xfe00` (EOF) function `read_valid_eof1_header` (called by `analyze`) assumes that there is something else but it's the whole code so it tries to read data from the outside of the code section.

test/t8n/t8n.cpp
axic: Right, can we just include the namespace (`using namespace nlohmann`)?

test/t8n/t8n.cpp
```diff
@@ -0,0 +1,169 @@
...
+            evmc::VM vm{evmc_create_evmone(), {{"O", "0"}}};
+
+            std::vector<state::Log> txs_logs;
+
+            if (j_txs.is_array())
+            {
+                j_result["receipts"] = json::json::array();
+                j_result["rejected"] = json::json::array();
+                auto idx = 0;
+                for (auto& j_tx : j_txs)
```
axic, line 101: Is this modifying `j_txs` or can this be const?

rodiazet (approved):

test/statetest/statetest_loader.cpp
```diff
@@ -255,6 +255,38 @@ static void from_json(const json::json& j, StateTransitionTest& o)
...
+    state::Transaction o;
+    o.data = from_json<bytes>(j.at("input"));
+    o.gas_limit = from_json<int64_t>(j.at("gas"));
+    o.value = from_json<intx::uint256>(j.at("value"));
+
+    if (!j.at("to").get<std::string>().empty())
+        o.to = from_json<evmc::address>(j.at("to"));
+
+    o.sender = from_json<evmc::address>(j.at("sender"));
+    // FIXME: Get max gas price
```
axic, line 270: Ia this commit still valid?

chfast: No.

PR #553, opened 2023-02-06
EOFCREATE and RETURNCONTRACT implementation

gumb0:
TODO:
- [x] Validate CREATE3 to not allow it with truncated containers (currently implemented as run-time check, but should be done at validation, but requires parsing subcontainers before CREATE3 validation)
- [x] Fix some commented out tests
- ~~[ ] Optimization to replace validation recursion with iteration~~ created separate issue #794 
- [x] More transition tests and unit tests
- [x] Merge EVMC changes
- [x] **Update state tests**


lib/evmone/eof.cpp
```diff
@@ -576,9 +625,34 @@ EOF1Header read_valid_eof1_header(bytes_view container)
...
 }
 
+void append_data_section(bytes& container, bytes_view aux_data)
+{
+    const auto header = read_valid_eof1_header(container);
+    const auto insert_pos =
+        header.container_offsets.empty() ? container.size() : header.container_begin(0);
+    container.insert(insert_pos, aux_data);
+    const auto new_data_size = header.data_size + aux_data.size();
+
```
axic, line 647: Should check/assert for `new_data_size < 2^16`.

lib/evmone/instructions.hpp
```diff
@@ -968,6 +971,30 @@ inline StopToken return_impl(StackTop stack, ExecutionState& state) noexcept
...
+    const auto& size = stack[1];
+
+    if (!check_memory(state, offset, size))
+        return {EVMC_OUT_OF_GAS};
+
+    const auto deploy_container_index = size_t{pos[1]};
+
+    const auto header = read_valid_eof1_header(state.original_code);
+    bytes deploy_container{&state.original_code[header.container_begin(deploy_container_index)],
+        header.container_sizes[deploy_container_index]};
```
axic, line 986: Shouldn't we have a `container_size()` helper?

lib/evmone/instructions.hpp
```diff
@@ -968,6 +971,30 @@ inline StopToken return_impl(StackTop stack, ExecutionState& state) noexcept
...
+
+    // Append (offset, size) to data section
+    const auto deploy_header = read_valid_eof1_header(deploy_container);
```
axic, line 989: What is the assignment doing here?

lib/evmone/instructions_traits.hpp
```diff
@@ -171,6 +171,8 @@ constexpr inline GasCostTable gas_costs = []() noexcept {
     table[EVMC_CANCUN] = table[EVMC_SHANGHAI];
     table[EVMC_CANCUN][OP_DUPN] = 3;
     table[EVMC_CANCUN][OP_SWAPN] = 3;
+    table[EVMC_CANCUN][OP_CREATE3] = 32000;
+    table[EVMC_CANCUN][OP_RETURNCONTRACT] = 0;
```
axic, line 175: To think about this: probably neither of these prices are good?

lib/evmone/instructions.hpp
gumb0: removed

lib/evmone/eof.cpp
gumb0: Added the check.

lib/evmone/instructions.hpp
gumb0: Added the helper, but we also need to add `code_size()` to `eof` branch.

lib/evmone/eof.cpp
```diff
@@ -90,87 +197,494 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_eof_headers(
...
+    auto msh_it = std::max_element(stack_heights.begin(), stack_heights.end());
+
+    if (std::find(stack_heights.begin(), stack_heights.end(), -1) != stack_heights.end())
+        return {EOFValidationError::unreachable_instructions, -1};
+
+    return {EOFValidationError::success, *msh_it};
+}
+
+std::variant<EOF1Header, EOFValidationError> validate_eof1(  // NOLINT(misc-no-recursion)
```
gumb0, line 481: I counted already 7 places, including this, where we suppress recursion warning. Does it make sense to disable globally?

gumb0:
EOFv2 Design Space currently specifies this feature:

> Initcode is allowed terminate execution with STOP. It keeps the changes done to the state but the new contract is not created. This allows using initcode as ephemeral code.

Questions: 
- Do we really need this use case?
  - Without it for successful CREATE3 some valid container still has to be returned and deployed, even if not needed. 
  - It can be argued that for ephemeral case we still could 

- In case we do allow STOP which doesn't result in account creation, do we wanto to signal to CREATE3 caller that it succeeded without creation and how?
  - 0 is returned on stack for REVERT case
  - address of new contract is returned if CREATE3 finished with RETURNCONTRACT
  - Options for signalling STOP  
    - Some other magic address value could be returned for STOP case
    - Pushing two stack values (error and address separately)
    - Saving created address in memory
      - Can be input data buffer
      - Can be return data

lib/evmone/eof.cpp
```diff
@@ -507,9 +578,47 @@ EOF1Header read_valid_eof1_header(bytes_view container)
...
+        return false;
+
+    // Insert position is before first embedded container, or at the end if there're none.
+    const auto insert_pos =
+        header.container_offsets.empty() ? container.size() : header.container_begin(0);
+    container.insert(insert_pos, aux_data);
+
+    // find data size position
+    const auto num_code_sections = header.code_sizes.size();
+    const auto data_size_pos = std::size(MAGIC) + 1 + 3 + 3 + 2 * num_code_sections + 1;
```
axic, line 615: Why not store this in the header?

lib/evmone/eof.cpp
```diff
@@ -507,9 +578,47 @@ EOF1Header read_valid_eof1_header(bytes_view container)
...
+    const auto section_headers_or_error = validate_eof_headers(container);
+    if (std::holds_alternative<EOFValidationError>(section_headers_or_error))
+        return false;
+
+    const auto header = read_valid_eof1_header(container);
+    const auto new_data_size = header.data_size + aux_data.size();
+    if (new_data_size > std::numeric_limits<uint16_t>::max())
+        return false;
+
+    // Insert position is before first embedded container, or at the end if there're none.
```
axic, line 608: Hm, so container is last? Need to test this well.

This will really push having metadata at the beginning of the data section.

lib/evmone/instructions.hpp
```diff
@@ -966,6 +982,34 @@ inline StopToken return_impl(StackTop stack, ExecutionState& state) noexcept
...
+
+    const auto header = read_valid_eof1_header(state.original_code);
+    bytes deploy_container{&state.original_code[header.container_begin(deploy_container_index)],
+        header.container_size(deploy_container_index)};
+
+    // Append (offset, size) to data section
+    if (!append_data_section(deploy_container,
+            {&state.memory[static_cast<size_t>(offset)], static_cast<size_t>(size)}))
+        return {EVMC_OUT_OF_GAS};
+
```
axic, line 1007:
```suggestion

    // TODO: charge gas based on size of container?

```

lib/evmone/instructions_calls.cpp
```diff
@@ -180,6 +180,66 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
...
+        return EVMC_OUT_OF_GAS;
+
+    const auto input_offset = static_cast<size_t>(input_offset_u256);
+    const auto input_size = static_cast<size_t>(input_size_u256);
+
+    if (state.msg->depth >= 1024)
+        return EVMC_SUCCESS;  // "Light" failure.
+
+    if (endowment != 0 &&
+        intx::be::load<uint256>(state.host.get_balance(state.msg->recipient)) < endowment)
```
axic, line 209: Hm why is this recipient and not sender? I know `create_impl` does the same but naming is counter intuitive.

lib/evmone/instructions_calls.cpp
```diff
@@ -180,6 +180,66 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
     return EVMC_SUCCESS;
 }
 
+evmc_status_code create3(StackTop stack, ExecutionState& state, code_iterator& pos) noexcept
```
axic, line 183: Currently there is a lot of overlap with `create_impl`. Would it make sense templatizing it with some `if constexpr` branches?

lib/evmone/instructions.hpp
```diff
@@ -952,6 +961,13 @@ inline code_iterator retf(StackTop /*stack*/, ExecutionState& state, code_iterat
 template <evmc_status_code StatusCode>
 inline StopToken return_impl(StackTop stack, ExecutionState& state) noexcept
 {
+    // RETURN is forbidden inside CREATE3 context
+    if constexpr (StatusCode == EVMC_SUCCESS)
```
axic, line 965: Why check for this?

lib/evmone/eof.cpp
gumb0: Can be a helper, it will be used only here.

lib/evmone/instructions.hpp
axic: `stop_impl` has the same branch. All this is a bit confusing, maybe need to document it better.

axic:
Reading this diff I am unsure how account creation works in evmone/host 😅

lib/evmone/instructions_calls.cpp
```diff
@@ -180,6 +180,66 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
...
+        msg.input_size = input_size;
+    }
+
+    msg.sender = state.msg->recipient;
+    msg.depth = state.msg->depth + 1;
+    msg.create2_salt = intx::be::store<evmc::bytes32>(salt);
+    msg.value = intx::be::store<evmc::uint256be>(endowment);
+    msg.initcontainer_index = initcontainer_index;
+
+    const auto result = state.host.call(msg);
```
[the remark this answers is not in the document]
gumb0: Deployment cost is charged by host in `Host::create()`

lib/evmone/instructions_calls.cpp
```diff
@@ -180,6 +180,84 @@ evmc_status_code create_impl(StackTop stack, ExecutionState& state) noexcept
...
+    if ((state.gas_left -= initcode_cost) < 0)
+        return EVMC_OUT_OF_GAS;
+
+    if (!check_memory(state, input_offset_u256, input_size_u256))
+        return EVMC_OUT_OF_GAS;
+
+    const auto input_offset = static_cast<size_t>(input_offset_u256);
+    const auto input_size = static_cast<size_t>(input_size_u256);
+
+    // Charge for input data hashing.
```
chfast, line 215: We should not hash input so this is not needed.

test/unittests/evm_eof_test.cpp
chfast, line 1: We should create separate files for create and call EOF unit tests

lib/evmone/instructions_calls.cpp
gumb0: Why not? I think it's useful (calling constuctor with different arguments normally means someone wants to deploy another instance at diferrent address) and mimics the way CREATE2 works.

test/unittests/evm_eof_test.cpp
gumb0: Current approach to creation unit tests is pretty useless anyway, as `MockedHost` mocks all internal call frames.

chfast: I'm trying to do state-like tests in #589.

lib/evmone/instructions_calls.cpp
gumb0: Found the discussion where some people are in favor of not hashing arguments https://github.com/ethereum/eips/pull/3171

lib/evmone/instructions.hpp
```diff
@@ -1068,6 +1084,33 @@ inline TermResult return_impl(StackTop stack, int64_t gas_left, ExecutionState&
 inline constexpr auto return_ = return_impl<EVMC_SUCCESS>;
 inline constexpr auto revert = return_impl<EVMC_REVERT>;
 
+inline TermResult returncontract(
+    StackTop stack, int64_t gas_left, ExecutionState& state, code_iterator pos) noexcept
+{
+    const auto& offset = stack[0];
+    const auto& size = stack[1];
+
+    if (state.msg->kind != EVMC_CREATE3)
```
gumb0, line 1093: In case we end up allowing EOF creation transactions with empty `to` (as opposed to relying only on transaction to Creator Contract), this will have to be adjusted to allow `RETURNCONTRACT` in the context of creation transaction.

lib/evmone/instructions_calls.cpp
```diff
@@ -199,6 +199,75 @@ Result create_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noex
...
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    if (!check_memory(gas_left, state.memory, input_offset_u256, input_size_u256))
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    const auto input_offset = static_cast<size_t>(input_offset_u256);
+    const auto input_size = static_cast<size_t>(input_size_u256);
+
+    if (state.msg->depth >= 1024)
+        return {EVMC_SUCCESS, gas_left};  // "Light" failure.
```
gumb0, line 235: Should make sure to do `pos += 2`  in case of light failures.

gumb0: Fixed.

lib/evmone/eof.cpp
gumb0: Here I actually decided better to rewrite without recursion.

pdobacz:
>     * [ ]  Fix some commented out tests

They have been fixed and uncommented in `eof-create4` (I think all of them relied on CREATE4), can tick that one off.

gumb0:
Rebased.

gumb0:
Rebased and reworked subcontainers and truncated data support in `bytecode.hpp`

gumb0:
Rebased and squashed.

lib/evmone/eof.cpp
gumb0: > Here I actually decided better to rewrite without recursion.

I would do it in a separate PR #794

test/unittests/state_transition_eof_test.cpp
gumb0, line 1: Moved to `eof_create4` branch

lib/evmone/eof.cpp
```diff
@@ -167,12 +195,19 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_v
...
-                                     section_headers[DATA_SECTION].front();
+    const auto section_bodies_without_data =
+        section_headers[TYPE_SECTION].front() +
+        std::accumulate(
+            section_headers[CODE_SECTION].begin(), section_headers[CODE_SECTION].end(), 0) +
+        std::accumulate(section_headers[CONTAINER_SECTION].begin(),
+            section_headers[CONTAINER_SECTION].end(), 0);
     const auto remaining_container_size = container_end - it;
-    if (section_bodies_size != remaining_container_size)
+    // Only data section may be truncated, so remaining_container size must be in
```
hugo-dc, line 205: We need verify we are validating subcontainers ([here](https://github.com/ethereum/evmone/blob/3ebb3c3dc8df344c62853950be8688dd99e333eb/lib/evmone/eof.cpp#L595)), because for regular EOF code validation ([here](https://github.com/ethereum/evmone/blob/3ebb3c3dc8df344c62853950be8688dd99e333eb/lib/evmone/eof.cpp#L761)) data section must be the exact size indicated in the data section header

hugo-dc: Makes sense, eofparse tool would also need to validate the container is not truncated as well.

gumb0: > Makes sense, eofparse tool would also need to validate the container is not truncated as well.

I would say it should check that all CREATE3 use non-truncated container, because it is part of validation, but top-level container should be allowed to be truncated.

test/unittests/evm_eof_test.cpp
```diff
@@ -352,3 +353,98 @@ TEST_P(evm, datacopy_memory_cost)
...
+{
+    rev = EVMC_CANCUN;
+    const auto code =
+        calldatacopy(0, 0, OP_CALLDATASIZE) + OP_CALLDATASIZE + 0 + OP_RETURNCONTRACT + Opcode{0};
+
+    execute(code);
+    EXPECT_STATUS(EVMC_UNDEFINED_INSTRUCTION);
+}
+
+TEST_P(evm, returncontract_not_int_initcode)
```
pdobacz, line 423:
```suggestion
TEST_P(evm, returncontract_not_in_initcode)
```

test/unittests/state_tx_test.cpp
```diff
@@ -153,3 +154,39 @@ TEST(state_tx, validate_blob_tx)
...
+        .s = 0,
+    };
+
+    ASSERT_FALSE(holds_alternative<std::error_code>(
+        validate_transaction(acc, bi, tx, EVMC_CANCUN, 60000, BlockInfo::MAX_BLOB_GAS_PER_BLOCK)));
+
+    EXPECT_EQ(std::get<std::error_code>(validate_transaction(acc, bi, tx, EVMC_PRAGUE, 60000,
+                                            BlockInfo::MAX_BLOB_GAS_PER_BLOCK))
+                  .message(),
+        "EOF in creation transaction");
```
pdobacz, line 186:
```suggestion
        "EOF initcode in creation transaction");
```

to be more specific? Can also be "in creation transaction data"

pdobacz (approved):
Very nice, I appreciate the very thorough testing. Just a few nitpicks left, but pls someone else review and give a stamp too.

Late thought: do we have a test that a `CREATE3-RETURNCONTRACT`-ed contract can be successfully called, for example DATALOAD ing from the dynamic aux data part. To ensure that the deployment process doesn't garble the EOF container in any way.

gumb0:
> Late thought: do we have a test that a `CREATE3-RETURNCONTRACT`-ed contract can be successfully called, for example DATALOAD ing from the dynamic aux data part. To ensure that the deployment process doesn't garble the EOF container in any way.

Added such test.

gumb0:
Renamed CREATE3 to EOFCREATE.

gumb0:
Not sure yet what happened to the coverage, it was almost everything covered some time ago :disappointed:

gumb0:
Rebased.

pdobacz (approved):

PR #554, opened 2023-02-08
statetest: Extract logs_hash() function

chfast:


test/statetest/statetest_runner.cpp
```diff
@@ -18,6 +18,11 @@ inline bytes rlp_encode(const Log& log)
 
 namespace evmone::test
 {
+hash256 logs_hash(const std::vector<state::Log>& logs)
```
axic, line 21: Why dup here?

chfast: Mistake. It was my first attempted location.

axic (approved):

rodiazet (approved):

PR #555, opened 2023-02-08
Add project version printing for t8n executable

rodiazet:


test/t8n/t8n.cpp
```diff
@@ -47,7 +47,7 @@ int main(int argc, const char* argv[])
 
         if (arg == "-v")
         {
-            std::cout << "evmone dev\n";  // FIXME: Return proper version.
+            std::cout << "evmone " << PROJECT_VERSION << std::endl;
```
axic, line 50:
```suggestion
            std::cout << "evmone-t8n " << PROJECT_VERSION << std::endl;
```

Why not?

rodiazet: Right.

axic (approved):

test/t8n/t8n.cpp
chfast: Depends what you want to see in the generated tests.

rodiazet: I think -t8n is ok. We will probably will have also project versions for other executables so it's good to distinct them

test/t8n/t8n.cpp
```diff
@@ -47,7 +47,7 @@ int main(int argc, const char* argv[])
 
         if (arg == "-v")
         {
-            std::cout << "evmone dev\n";  // FIXME: Return proper version.
+            std::cout << "evmone-t8n " << PROJECT_VERSION << std::endl;
```
chfast, line 50:
```suggestion
            std::cout << "evmone-t8n " PROJECT_VERSION "\n";
```

chfast (changes requested):

test/t8n/t8n.cpp
rodiazet: Why is it better?

chfast: The string will be concatenated by the compiler and `std::cout` called only once.

rodiazet: Done

chfast (approved):

PR #557, opened 2023-02-08
ci: Run Windows builds in parallel

chfast:


circle.yml
```diff
@@ -585,9 +585,7 @@ workflows:
           filters:
             tags:
               only: /.*/
-      - release-windows-32bit:
-          requires:
-            - release-windows
+      - release-windows-32bit
```
axic, line 588: Probably need to add the filters then?

chfast: No need. It didn't have any filters before.

axic: The dependency implicitly caused filtering?

axic (approved):

PR #558, opened 2023-02-08
Add state.reward processing

rodiazet:


test/t8n/t8n.cpp
```diff
@@ -54,6 +55,13 @@ int main(int argc, const char* argv[])
             output_result_file = argv[i];
         else if (arg == "--output.alloc" && ++i < argc)
             output_alloc_file = argv[i];
+        else if (arg == "--state.reward" && ++i < argc)
+        {
+            if (std::string(argv[i]) == "-1")
+                state_reward = std::numeric_limits<intx::uint256>::max();
```
axic, line 61: This is for disabling it? Why not use `std::optional<intx::uint256>` in that case?

test/t8n/t8n.cpp
```diff
@@ -128,6 +136,16 @@ int main(int argc, const char* argv[])
                 }
             }
 
+            if (state_reward != 0 && state_reward != std::numeric_limits<intx::uint256>::max())
+                state.get_or_insert(block.coinbase).balance += state_reward;
+            else if (state_reward == 0)
+                state.touch(block.coinbase);
+            else if (state_reward == std::numeric_limits<intx::uint256>::max())
+            {
+                if (state.get_or_insert(block.coinbase).is_empty())
```
axic, line 145: Hm, cleaning up empty accounts is just an EIP-161 rule. Don't we have that implemented?

test/t8n/t8n.cpp
chfast: I agree here.

test/t8n/t8n.cpp
```diff
@@ -30,6 +30,7 @@ int main(int argc, const char* argv[])
     fs::path output_dir;
     fs::path output_result_file;
     fs::path output_alloc_file;
+    intx::uint256 state_reward;
```
chfast, line 33: Rename it to block reward.

test/t8n/t8n.cpp
rodiazet: For `<= EIP158` coinbase account is touched even it tx is rejected. retesteth passing -1 as a stare.reward requires zero state to be removed.

test/t8n/t8n.cpp
```diff
@@ -54,6 +55,13 @@ int main(int argc, const char* argv[])
             output_result_file = argv[i];
         else if (arg == "--output.alloc" && ++i < argc)
             output_alloc_file = argv[i];
+        else if (arg == "--state.reward" && ++i < argc)
+        {
+            if (std::string(argv[i]) == "-1")
```
chfast, line 60:
```suggestion
            if (std::string_view value{argv[i]}; value == "-1")
```

test/t8n/t8n.cpp
```diff
@@ -54,6 +55,13 @@ int main(int argc, const char* argv[])
...
+                block_reward = {};
+            else
+                block_reward = intx::from_string<intx::uint256>(argv[i]);
```
chfast, line 63:
```suggestion
                block_reward = intx::from_string<intx::uint256>(value);
```

test/t8n/t8n.cpp
```diff
@@ -128,6 +136,16 @@ int main(int argc, const char* argv[])
                 }
             }
 
+            if (block_reward.has_value())
+            {
+                if (*block_reward == 0)
+                    state.touch(block.coinbase);
```
chfast, line 142: This should handle both cases.

```suggestion
                    state.touch(block.coinbase).balance += *block_reward;
```

test/t8n/t8n.cpp
```diff
@@ -128,6 +136,16 @@ int main(int argc, const char* argv[])
...
+                else
+                    state.get_or_insert(block.coinbase).balance += *block_reward;
+            }
+            else if (rev <= EVMC_TANGERINE_WHISTLE)
```
chfast, line 146: Don't we do the same in transaction execution?

test/t8n/t8n.cpp
rodiazet: Adding 0 does not change so much :)

test/t8n/t8n.cpp
rodiazet: no, we are touching there here we are erasing.

test/t8n/t8n.cpp
```diff
@@ -54,6 +55,13 @@ int main(int argc, const char* argv[])
             output_result_file = argv[i];
         else if (arg == "--output.alloc" && ++i < argc)
             output_alloc_file = argv[i];
+        else if (arg == "--state.reward" && ++i < argc)
+        {
+            if (std::string_view(argv[i]) == "-1")
+                block_reward = {};
```
axic, line 61: Do we actually need to set this here? Can we just have `!= "-1"` case for setting the value?

test/t8n/t8n.cpp
```diff
@@ -128,6 +132,11 @@ int main(int argc, const char* argv[])
                 }
             }
 
+            if (block_reward.has_value())
+                state.touch(block.coinbase).balance += *block_reward;
+            else if (rev <= EVMC_TANGERINE_WHISTLE)  // This behaviour is required by retesteth
+                state.get_accounts().erase(block.coinbase);
```
axic, line 138: Hm, this will remove it even when other transactions touched this account. Is that desired?

rodiazet: Sorry. Fore some reason I remove also checking if it's empty. Fixed now

test/t8n/t8n.cpp
```diff
@@ -54,6 +55,9 @@ int main(int argc, const char* argv[])
             output_result_file = argv[i];
         else if (arg == "--output.alloc" && ++i < argc)
             output_alloc_file = argv[i];
+        else if (arg == "--state.reward" && ++i < argc)
+            if (std::string_view(argv[i]) != "-1")
```
chfast, line 59: You have to combine all 3 conditions. You can be fancy by doing `argv[i] != "-1"sv`.

test/t8n/t8n.cpp
```diff
@@ -128,6 +132,12 @@ int main(int argc, const char* argv[])
                 }
             }
 
+            if (block_reward.has_value())
+                state.touch(block.coinbase).balance += *block_reward;
```
chfast, line 136: This should be done _before_ any transaction execution.

test/t8n/t8n.cpp
chfast: I'm still not convinced by this quirk. It looks it replicates a bug in some client implementation. Can you provide a list of failing tests?

test/t8n/t8n.cpp
rodiazet: error: no matching literal operator for call to 'operator""sv' with arguments of types 'const char *' and 'unsigned long', and no matching literal operator template
        else if (arg == "--state.reward" && ++i < argc && argv[i] != "-1"sv)

chfast: `using namespace std::literals`.

test/t8n/t8n.cpp
rodiazet: I removed it. It looks that now previously failing test passed. I will check why and let know.

rodiazet: Are you sure? Moving block reward to coinbase before transaction made a lot of test started failing.

chfast: Should be _after_

chfast (approved):

PR #560, opened 2023-02-10
state: Implement EIP-3651: Warm COINBASE

chfast:
Spec: https://eips.ethereum.org/EIPS/eip-3651
Tested with: https://github.com/ethereum/tests/pull/1082.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -156,6 +156,11 @@ std::variant<TransactionReceipt, std::error_code> transition(
         for (const auto& key : storage_keys)
             storage[key].access_status = EVMC_ACCESS_WARM;
     }
+    // EIP-3651: Warm COINBASE.
+    // This may create an empty coinbase account. The account cannot be created unconditionally
+    // because this breaks old revisions.
+    if (rev >= EVMC_SHANGHAI)
+        host.access_account(block.coinbase);
 
     const auto result = host.call(build_message(tx, execution_gas_limit));
 


yperbasis (approved):

axic (commented):
Any unit test for this?

test/state/state.cpp
```diff
@@ -156,6 +156,10 @@ std::variant<TransactionReceipt, std::error_code> transition(
         for (const auto& key : storage_keys)
             storage[key].access_status = EVMC_ACCESS_WARM;
     }
+    // EIP-3651: Warm COINBASE.
+    // We cannot touch it here because it may create an account in pre Spurious Dragon.
```
axic, line 160: Hm? The condition is >=Shanghai, so not sure how Spurious Dragon is related.

chfast: This refers to the implementation detail that the coinbase is touched after the transaction execution (see some lines below). This gives 2 coinbase account lookups, but it looks we cannot do better because we are not allowed to create the account here for pre Spurious Dragon revisions.

chfast:
> Any unit test for this?

We don't have unit tests for State/Host 😬. We mostly relay on state tests.

rodiazet (approved):

test/state/state.cpp
axic: > because we are not allowed to create the account here for pre Spurious Dragon revisions.

Since the condition is >=Shanghai, this can never happen. Or the point of the comment is to explain the condition?

chfast: I changed the comment.

axic (approved):

chfast:
> We don't have unit tests for State/Host grimacing. We mostly relay on state tests.

Actually, we have some in test/integration/statetest.

PR #563, opened 2023-02-14
EOF "1.0"

gumb0:
Adds implementation of EVM Object Format (EOF) version "1.0". Which is a collection of 5 EIPs originally proposed for Shanghai Ethereum upgrade:

- [EIP-3540: EOF - EVM Object Format v1](https://eips.ethereum.org/EIPS/eip-3540) ([cad968c](https://github.com/ethereum/EIPs/blob/cad968c6c4be303cd0ef6cddd8f0513a3b20003b/EIPS/eip-3540.md))
- [EIP-3670: EOF - Code Validation](https://eips.ethereum.org/EIPS/eip-3670) ([f29a424](https://github.com/ethereum/EIPs/blob/f29a4248a93e1f89839023b790e20ad39b426aaf/EIPS/eip-3670.md))
- [EIP-4200: EOF - Static relative jumps](https://eips.ethereum.org/EIPS/eip-4200) ([1b4112a](https://github.com/ethereum/EIPs/blob/1b4112a8207ffe0e54910c772e6d12d2886300d8/EIPS/eip-4200.md))
- [EIP-4750: EOF - Functions](https://eips.ethereum.org/EIPS/eip-4750) ([ce9748c](https://github.com/ethereum/EIPs/blob/ce9748cc0881632974943ad55f67518dde8a8de1/EIPS/eip-4750.md))
- [EIP-5450: EOF - Stack Validation](https://eips.ethereum.org/EIPS/eip-5450) ([0f3191f](https://github.com/ethereum/EIPs/blob/0f3191f6e8a1613972d7fdaecc01470c50bd00b8/EIPS/eip-5450.md))

PR #572, opened 2023-02-20
eof: Change handling of errors from pair<val, err> to variant<val, err>

axic:
Pulled out of #563, but not sure who the author is.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -8,6 +8,7 @@
 #include <array>
 #include <cassert>
 #include <limits>
+#include <variant>
 
 namespace evmone
 {
@@ -21,7 +22,8 @@ constexpr uint8_t MAX_SECTION = DATA_SECTION;
 
 using EOFSectionHeaders = std::array<uint16_t, MAX_SECTION + 1>;
 
-std::pair<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_view container) noexcept
+std::variant<EOFSectionHeaders, EOFValidationError> validate_eof_headers(
+    bytes_view container) noexcept
 {
     enum class State
     {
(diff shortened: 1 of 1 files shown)


lib/evmone/eof.cpp
```diff
@@ -186,7 +189,12 @@ EOFValidationError validate_eof(evmc_revision rev, bytes_view container) noexcep
     {
         if (rev < EVMC_CANCUN)
             return EOFValidationError::eof_version_unknown;
-        return validate_eof1(rev, container).second;
+
+        const auto header_or_error = validate_eof1(rev, container);
```
axic, line 193: It seems this function isn't even using the returned `EOF1Header` and so the entire deal with `pair`/`variant` seems useless.

axic:
@chfast @gumb0 @rodiazet it is draft because I'm unsure of usefulness, but please review.

gumb0:
Have you seen https://github.com/ethereum/evmone/pull/563/commits/6acd9d321003ecf04261a752e552714bfb1e7dcb ?

lib/evmone/eof.cpp
gumb0: For simplicity we can change `validate_eof1()`  to return only error.

gumb0: We can also make pubic `validate_eof()` return variant, in general it seems useful to be able to validate and parse in one go.

But currently I found that it could be useful only in `eofparse` (which does `validate_eof()` + `read_valid_eof1_header()` combo)

State test runner seems to do these two operations in different places.

axic:
> Have you seen https://github.com/ethereum/evmone/commit/6acd9d321003ecf04261a752e552714bfb1e7dcb ?
> 
> Ah I see it's the same commit (I was the author)

I actually duplicated the change from the final diff and was lazy to find the actual commit. Will change the author then.

gumb0 (approved):

PR #575, opened 2023-02-27
state: Refactor `compute_new_account_address()`

chfast:
Refactor `compute_new_account_address()` to match 𝐀𝐃𝐃𝐑 from Yellow Paper.
This makes it a separate utility independent of `evmc_message`.

### DIFF
--- a/test/state/host.hpp
+++ b/test/state/host.hpp
@@ -15,6 +15,12 @@ using evmc::uint256be;
 inline constexpr size_t max_code_size = 0x6000;
 inline constexpr size_t max_initcode_size = 2 * max_code_size;
 
+/// Computes the address of to-be-created contract.
+///
+/// Computes the new account address for the contract creation context
+/// as defined by `ADDR()` in Yellow Paper, 7. Contract Creation, (86).
+address compute_new_account_address(const evmc_message& msg, uint64_t sender_nonce) noexcept;
+
 class Host : public evmc::Host
 {
     evmc_revision m_rev;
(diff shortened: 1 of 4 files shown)


test/state/host.hpp
```diff
@@ -18,8 +18,15 @@ inline constexpr size_t max_initcode_size = 2 * max_code_size;
...
-/// as defined by `ADDR()` in Yellow Paper, 7. Contract Creation, (86).
-address compute_new_account_address(const evmc_message& msg, uint64_t sender_nonce) noexcept;
+/// as defined by 𝐀𝐃𝐃𝐑 in Yellow Paper, 7. Contract Creation, (86).
+///
+/// @param sender        The address of the message sender. YP: 𝑠.
+/// @param sender_nonce  The sender's nonce before the increase. YP: 𝑛.
+/// @param salt          The salt for CREATE2. If null, CREATE address is computed. YP: ζ.
+/// @param init_code     The contract creation init code. Value only affects CREATE2. YP: 𝐢.
+/// @return              The computed address for CREATE or CREATE2 scheme.
+address compute_new_account_address(const address& sender, uint64_t sender_nonce,
```
axic, line 28: Why not have two overloads:
```
address compute_new_account_address(const evmc_message& msg, uint64_t sender_nonce) noexcept;
address compute_new_account_address(const address& sender, bytes32& salt, bytes_view init_code) noexcept;
```

The `sender_nonce` is irrelevant for create2 and this separation makes it clear which values are used in which.

chfast: Mostly so that a user don't need to know which overload to select. Now it only needs to provide arguments (sometimes redundantly) and will get the right answer. Non-functional change from what we had.

axic (approved):

PR #578, opened 2023-03-03
Simplify validate_eof_headers()

gumb0:
This refactors checking for the order of sections, replaces all specific section-out-of-order checks with one generic check, and so many error messages are also replaced with generic `expected_section_missing`.

Missing terminator also generates `expected_section_missing`.


lib/evmone/eof.cpp
```diff
@@ -61,48 +61,33 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_v
...
     while (it != container_end && state != State::terminated)
     {
         switch (state)
         {
         case State::section_id:
         {
             section_id = *it++;
+
+            if (section_id != expected_section_id)
+                return EOFValidationError::expected_section_missing;
```
chfast, line 74: Here you can select error code by `expected_section_id` so we preserve more detailed error information. If you want to be fancy, you can do something like
`return errc::type_section_missing + expected_section_id`.

lib/evmone/eof.hpp
```diff
@@ -63,24 +63,15 @@ enum class EOFValidationError
 
     incomplete_section_size,
     incomplete_section_number,
-    code_section_missing,
-    type_section_missing,
-    data_section_missing,
-    multiple_data_sections,
-    unknown_section_id,
+    expected_section_missing,
```
chfast, line 66: "missing" implies "expected" so... `section_missing`?

lib/evmone/eof.cpp
```diff
@@ -112,11 +97,12 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_eof_headers(bytes_v
                 section_num = static_cast<uint16_t>((section_number_hi << 8) | section_number_lo);
                 if (section_num == 0)
                     return EOFValidationError::zero_section_size;
+                expected_section_id = DATA_SECTION;
                 state = State::section_size;
                 break;
             }
             default:
-                return EOFValidationError::unknown_section_id;
```
chfast, line 67: Was this executable?

chfast (commented):
Looks good to me to integrate to the main PR.

lib/evmone/eof.cpp
gumb0: Yes it was returned when section_id is unknown, but now it fails earlier with new error.

PR #580, opened 2023-03-11
Modernize test utils

chfast:
Apply changes suggested by clang-tidy-15.


axic (commented):
Not sure what benefit moving to `auto` makes, but fine with it because `base_fee` is still a fixed type.

axic (approved):

PR #581, opened 2023-03-11
ci: Upgrade compilers (incl. clang-15)

chfast:


circle.yml
```diff
@@ -236,8 +236,8 @@ commands:
         type: string
     steps:
       - run:
-          name: "Upgrade codecov"
-          command: sudo pip3 install --upgrade --quiet --no-cache-dir codecov
+          name: "Install codecov"
+          command: pip3 install --break-system-packages --upgrade --user --quiet --no-cache-dir codecov
```
axic, line 240: What is this setting doing?

circle.yml
```diff
@@ -427,7 +427,7 @@ jobs:
     steps:
       - build
       - download_execution_tests:
-          commit: e279c3d0a53e2f3055c7d7f72f993b62dde18e36
+          commit: v11.3
```
axic, line 430: The diff between e279c3d0a53e2f3055c7d7f72f993b62dde18e36 only seems to be concerned with EOF-related test changes.

axic (approved):
Looks okayish.

circle.yml
chfast: Installs codecov python script. In the new version of pip/Debian it is disallowed to install pip packges "globally". I have not found better way than forcing this with `--break-system-packages`.

circle.yml
chfast: I just picked the latest release which was newer than the previous hash.

circle.yml
axic: > I have not found better way than forcing this with --break-system-packages.

😬

gumb0 (approved):

axic:
Argh, didn't notice this wasn't on master. Sorry for merging.

PR #583, opened 2023-03-15
Load BlockInfo fixes required by exec-spec-tests

rodiazet:
-Add tests cases for BlockInfo loader according to bugs reported.
-Extend `from_json` implementation for `64` bits integers to support `hex` and `dec` strings and values
-Add special case handling to suport `currentRandom` passed as `0`. `execution-spec-tests` requires it.

### DIFF
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -69,6 +69,12 @@ template <typename T>
 T from_json(const json::json& j) = delete;
 
 template <>
+uint64_t from_json<uint64_t>(const json::json& j);
+
+template <>
+int64_t from_json<int64_t>(const json::json& j);
+
+template <>
 state::BlockInfo from_json<state::BlockInfo>(const json::json& j);
 
 template <>
--- a/test/unittests/statetest_loader_block_info_test.cpp
+++ b/test/unittests/statetest_loader_block_info_test.cpp
@@ -1,0 +1,114 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+#include <test/statetest/statetest.hpp>
+
+using namespace evmone;
+
+TEST(statetest_loader, block_info)
+{
+    constexpr std::string_view input = R"({
+            "currentCoinbase": "0x1111111111111111111111111111111111111111",
+            "currentDifficulty": "0x0",
+            "currentGasLimit": "0x0",
+            "currentNumber": "0",
+            "currentTimestamp": "0",
+            "currentBaseFee": "7",
+            "currentRandom": "0x00",
+            "withdrawals": []
+        })";
+
+    const auto bi = test::from_json<state::BlockInfo>(json::json::parse(input));
+    EXPECT_EQ(bi.coinbase, 0x1111111111111111111111111111111111111111_address);
+    EXPECT_EQ(bi.prev_randao, 0x00_bytes32);
+    EXPECT_EQ(bi.gas_limit, 0x0);
+    EXPECT_EQ(bi.base_fee, 7);
+    EXPECT_EQ(bi.timestamp, 0);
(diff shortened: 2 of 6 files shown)


test/unittests/statetest_loader_block_info_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+    EXPECT_EQ(bi.coinbase, 0x1111111111111111111111111111111111111111_address);
+    EXPECT_EQ(
+        bi.prev_randao, 0x0000000000000000000000000000000000000000000000000000000000000000_bytes32);
+    EXPECT_EQ(bi.gas_limit, 0x0);
+    EXPECT_EQ(bi.base_fee, 7);
+    EXPECT_EQ(bi.timestamp, 0);
+    EXPECT_EQ(bi.number, 0);
+}
+
+// TODO: Enable this test when this case will be properly handled.
```
chfast, line 33: What case is it?

test/unittests/statetest_loader_block_info_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+            "currentTimestamp": "0",
+            "currentBaseFee": "7",
+            "currentRandom": "0x00",
+            "withdrawals": []
+        })";
+
+    const auto bi = test::from_json<state::BlockInfo>(json::json::parse(input));
+    EXPECT_EQ(bi.coinbase, 0x1111111111111111111111111111111111111111_address);
+    EXPECT_EQ(
+        bi.prev_randao, 0x0000000000000000000000000000000000000000000000000000000000000000_bytes32);
```
chfast, line 26:
```suggestion
        bi.prev_randao, 0x00_bytes32);
```
Shorter literal should work too.

test/statetest/statetest.hpp
```diff
@@ -68,6 +68,12 @@ struct StateTransitionTest
 template <typename T>
 T from_json(const json::json& j) = delete;
 
+template <>
+uint64_t from_json<uint64_t>(const json::json& j);
+
+template <>
+int64_t from_json<int64_t>(const json::json& j);
```
chfast, line 75: I think we should remove `int64_t` in JSON. I expect we only use it for gas. So let's load it as `uint64_t`.

chfast: Ok, this is bigger change, so let's keep it as is.

test/unittests/statetest_loader_tx_test.cpp
```diff
@@ -46,7 +46,7 @@ TEST(statetest_loader, tx_eip1559)
 {
     constexpr std::string_view input = R"({
         "input": "b0b1",
-        "gas": "9091",
+        "gas": "0x9091",
```
axic, line 49: Why was this passing before?

rodiazet: Because we assumed earlier that every field is hex string. Now it can be both and hex has to be prefixed with `0x`

rodiazet: This should be also changed probably for bytes, but I don't would prefer to make a separated PR for this

test/unittests/statetest_loader_block_info_test.cpp
rodiazet: done

test/unittests/statetest_loader_tx_test.cpp
axic: Yes I see base was fixed at 16, and now it is auto-detect. Can we add more unit tests for that (to have multiple version of the inputs)?

rodiazet: ok

axic: Nice! Can you add also octal inputs because `stoull` supports that too.

chfast: Added one test each.

chfast (approved):

PR #586, opened 2023-03-18
EOF DATA* opcodes

gumb0:


lib/evmone/instructions.hpp
```diff
@@ -890,6 +890,31 @@ inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator
     return pos + 2;
 }
 
+inline evmc_status_code dataload(StackTop stack, ExecutionState& state) noexcept
+{
+    auto& index = stack.top();
+
+    if (state.msg->input_size < index)
```
axic, line 897: Why not just `input_size < index + 32`? That catches both cases?

lib/evmone/instructions.hpp
```diff
@@ -890,6 +890,22 @@ inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator
...
+
+    if (state.data.size() < 32 || (state.data.size() - 32) < index)
+        return EVMC_INVALID_MEMORY_ACCESS;  // TODO better error
+
+    const auto begin = static_cast<size_t>(index);
+    uint8_t data[32] = {};
+    for (size_t i = 0; i < 32; ++i)
+        data[i] = state.data[begin + i];
+
+    index = intx::be::load<uint256>(data);
```
axic, line 905: Is this taking an array (and not iterators)? Otherwise could just use
```suggestion
    index = intx::be::load<uint256>(begin + index);
```

gumb0: It takes array

axic: Need an overload in intx 😅

axic: Made an issue: https://github.com/chfast/intx/issues/288

lib/evmone/instructions.hpp
```diff
@@ -890,6 +890,68 @@ inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator
...
+
+    if (!check_memory(state, mem_index, size))
+        return EVMC_OUT_OF_GAS;
+
+    auto s = static_cast<size_t>(size);
+
+    if (state.data.size() < s || state.data.size() - s < data_index)
+        return EVMC_INVALID_MEMORY_ACCESS;  // TODO better error
+
+    const auto copy_cost = num_words(s) * 3;
```
axic, line 936: Hm, we could have some helper for this, given it is used by so many instructions.

lib/evmone/instructions_opcodes.hpp
```diff
@@ -163,6 +163,11 @@ enum Opcode : uint8_t
     OP_DUPN = 0xb5,
     OP_SWAPN = 0xb6,
 
+    OP_DATALOAD = 0xb7,
+    OP_DATASIZE = 0xb8,
+    OP_DATACOPY = 0xb9,
+    OP_DATALOADI = 0xba,
```
axic, line 169: So are we going with `DATALOADI` (I for immediate) or `DATALOADN` (for being in-line with PUSHn/SWAPN/DUPN).

gumb0: I think I like `DATALOADN` now

lib/evmone/instructions_opcodes.hpp
```diff
@@ -163,6 +163,11 @@ enum Opcode : uint8_t
     OP_DUPN = 0xb5,
     OP_SWAPN = 0xb6,
 
+    OP_DATALOAD = 0xb7,
```
axic, line 166: Can we reorder this so that `DATALOAD` + `DATALOADI` are next to eachother?

I think you chose this to match the order of `CALLDATA*` and to not having renumber in case `DATALOADI` is rejected.

Let's be optmistic however.

lib/evmone/instructions_opcodes.hpp
axic: Let's go with `DATALOADN`. @chfast any objection?

chfast: No.

axic:
What is the blocker on merging this?

gumb0:
> What is the blocker on merging this?

I think only 1 weird GCC compilation failure.

lib/evmone/baseline.cpp
```diff
@@ -365,9 +365,10 @@ evmc_result execute(evmc_vm* c_vm, const evmc_host_interface* host, evmc_host_co
     evmc_revision rev, const evmc_message* msg, const uint8_t* code, size_t code_size) noexcept
 {
     auto vm = static_cast<VM*>(c_vm);
-    const auto jumpdest_map = analyze(rev, {code, code_size});
-    auto state =
-        std::make_unique<ExecutionState>(*msg, rev, *host, ctx, bytes_view{code, code_size});
+    const bytes_view container{code, code_size};
+    const auto jumpdest_map = analyze(rev, container);
+    const auto data = jumpdest_map.eof_header.get_data(container);
```
axic, line 370: The naming of `jumpdest_map` now becomes weird.

lib/evmone/instructions.hpp
```diff
@@ -899,6 +899,68 @@ inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator
     return pos + 2;
 }
 
+inline Result dataload(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    auto& index = stack.top();
+
+    if (state.data.size() < 32 || (state.data.size() - 32) < index)
+        return {EVMC_INVALID_MEMORY_ACCESS, gas_left};  // TODO better error
```
axic, line 907: What would be a better error? Specific DATALOAD offset error?

lib/evmone/instructions.hpp
```diff
@@ -899,6 +899,68 @@ inline code_iterator swapn(StackTop stack, ExecutionState& state, code_iterator
...
+    const auto& data_index = stack.pop();
+    const auto& size = stack.pop();
+
+    if (!check_memory(gas_left, state.memory, mem_index, size))
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    auto s = static_cast<size_t>(size);
+
+    if (state.data.size() < s || state.data.size() - s < data_index)
+        return {EVMC_INVALID_MEMORY_ACCESS, gas_left};  // TODO better error
```
axic, line 948: Same here.

lib/evmone/instructions.hpp
gumb0: I was thinking it could be something like `EVMC_INVALID_DATA_ACCESS`, but not sure anymore if it's important, will just remove TODO for now.

lib/evmone/baseline.cpp
gumb0: Renamed.

lib/evmone/instructions.hpp
axic: @gumb0 how about this?

axic:
@gumb0 needs a rebase 😬

lib/evmone/instructions.hpp
axic: Right, I think that could make sense, but even `EVMC_INVALID_MEMORY_ACCESS` is okay probably. I'm fine keeping a comment mentioning "TODO: Introduce EVMC_INVALID_DATA_ACCESS".

lib/evmone/instructions.hpp
axic: By helper I mean `copy_cost` calculation / check.

axic (commented):
This mostly looks good to me, only two small comments (the `copy_cost` helper and the error code).

lib/evmone/instructions.hpp
gumb0: Added comments.

lib/evmone/instructions.hpp
gumb0: Added a helper.

axic (approved):
Looks good to me, probably want a review from @chfast or @rodiazet too.

axic:
@chfast @rodiazet okay to merge?

lib/evmone/instructions.hpp
chfast: Use `intx::be::unsafe::load(&state.data[begin])`.

lib/evmone/instructions.hpp
```diff
@@ -923,6 +923,22 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
     return {EVMC_SUCCESS, gas_left};
 }
 
+inline Result dataload(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    auto& index = stack.top();
+
+    if (state.data.size() < 32 || (state.data.size() - 32) < index)
+        return {EVMC_INVALID_MEMORY_ACCESS, gas_left};  // TODO: Introduce EVMC_INVALID_DATA_ACCESS
```
chfast, line 936: Not everyone is happy about "invalid memory access" in _return data_. Do we want to have it here?

lib/evmone/instructions.hpp
```diff
@@ -923,6 +928,68 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
...
+}
+
+inline Result datacopy(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    const auto& mem_index = stack.pop();
+    const auto& data_index = stack.pop();
+    const auto& size = stack.pop();
+
+    if (!check_memory(gas_left, state.memory, mem_index, size))
+        return {EVMC_OUT_OF_GAS, gas_left};  // TODO: Introduce EVMC_INVALID_DATA_ACCESS
```
chfast, line 972:
```suggestion
        return {EVMC_OUT_OF_GAS, gas_left};
```
This is memory expansion only so we should not have different error message here.

chfast: codecov says this is not covered by tests.

lib/evmone/instructions.hpp
```diff
@@ -923,6 +928,68 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
...
+
+    auto s = static_cast<size_t>(size);
+
+    if (state.data.size() < s || state.data.size() - s < data_index)
```
chfast, line 976: After memory check the `data_index + s` will not overflow.

lib/evmone/instructions.hpp
```diff
@@ -925,6 +925,59 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
...
+    const auto& mem_index = stack.pop();
+    const auto& data_index = stack.pop();
+    const auto& size = stack.pop();
+
+    if (!check_memory(gas_left, state.memory, mem_index, size))
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    auto s = static_cast<size_t>(size);
+
+    if (data_index + s > state.data.size())
```
chfast, line 965: `data_index + s` can overflow.

gumb0:
> @hugo-dc Any idea why EIP-3670 tests don't fail here? Don't we check that all undefined opcodes are rejected?
> 
> We change the validation to accept some new opcodes that tests consider invalid, so supposedly this change shouldn't pass current tests.

I investigated this a bit and I think that codes with new opcodes are most certainly still invalid for some other reason.
The tests I found check the code generated from a pattern `0xef000101000402000100020300000000000000<invalid_opcode>00` - this would be invalid either because of wrong stack for new instruction, or for wrong max stack height declared as 0.

So this makes those tests (`stEIP3670/*InvalidOpcodes`) kinda useless.

lib/evmone/instructions.hpp
gumb0: True, I wrongly changed it after this comment https://github.com/ethereum/evmone/pull/586#discussion_r1192195251

lib/evmone/execution_state.hpp
```diff
@@ -135,6 +135,9 @@ class ExecutionState
     /// For EOF-formatted code this is a reference to entire container.
     bytes_view original_code;
 
+    /// Reference to the EOF data section. Maybe be empty.
```
chfast, line 138:
```suggestion
    /// Reference to the EOF data section. May be empty.
```

lib/evmone/instructions.hpp
```diff
@@ -925,6 +925,59 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
...
+inline Result datacopy(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    const auto& mem_index = stack.pop();
+    const auto& data_index = stack.pop();
+    const auto& size = stack.pop();
+
+    if (!check_memory(gas_left, state.memory, mem_index, size))
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    auto s = static_cast<size_t>(size);
```
chfast, line 963:
```suggestion
    const auto s = static_cast<size_t>(size);
```

test/unittests/evm_eof_test.cpp
```diff
@@ -149,3 +149,195 @@ TEST_P(evm, eof_data_only_contract)
...
+    execute(code);
+    EXPECT_STATUS(EVMC_SUCCESS);
+    EXPECT_EQ(bytes_view(result.output_data, result.output_size),
+        "111111111111111122222222222222223333333333333333aaaaaaaaaaaaaaaa"_hex);
+
+    code = eof1_bytecode(bytecode(8) + 63 + 0 + OP_DATACOPY + ret(0, 32), 3, data);
+    execute(code);
+    EXPECT_STATUS(EVMC_INVALID_MEMORY_ACCESS);
+
+    code = eof1_bytecode(bytecode(0) + 65 + 0 + OP_DATACOPY + ret(0, 32), 3, data);
```
chfast, line 324: A copy of size `0` should always be fine.

lib/evmone/instructions.hpp
```diff
@@ -925,6 +925,59 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
...
+
+    if (state.data.size() < s || state.data.size() - s < data_index)
```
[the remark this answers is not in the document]
gumb0: RETURNDATACOPY still fails if size is zero but src_offset > returndata_size

test/unittests/evm_eof_test.cpp
gumb0: It's not for RETURNDATACOPY

axic:
@gumb0 what is outstanding here?

gumb0:
> @gumb0 what is outstanding here?

I think another round of review form @chfast

lib/evmone/instructions.hpp
chfast: I mean general design: maybe padding with zero is better. But I think this is good staring point.

chfast (approved):

PR #587, opened 2023-03-18
host: Make EOF opaque for EXTCODE* instructions

axic:


axic:
@chfast @gumb0 any feedback on the direction (before I start adding tests / more code)?

test/state/host.cpp
```diff
@@ -74,24 +74,38 @@ uint256be Host::get_balance(const address& addr) const noexcept
...
-    return (acc != nullptr) ? acc->code.size() : 0;
+    return (acc != nullptr) ? extcode(acc->code).size() : 0;
 }
 
 bytes32 Host::get_code_hash(const address& addr) const noexcept
 {
     // TODO: Cache code hash. It will be needed also to compute the MPT hash.
     const auto* const acc = m_state.find(addr);
-    return (acc != nullptr && !acc->is_empty()) ? keccak256(acc->code) : bytes32{};
+    return (acc != nullptr && !acc->is_empty()) ? keccak256(extcode(acc->code)) : bytes32{};
```
chfast, line 101: Patching `EXTCODEHASH` will be a headache for real implementations. The code hash is available directly in the account node. Now you need to additionally load the code to obtain a possibly different hash.

axic: > Now you need to additionally load the code to obtain a possibly different hash.

This was always the case with this approach. Need to figure out if it is an EOF account or not (by loading the account).

test/state/host.cpp
```diff
@@ -74,24 +74,38 @@ uint256be Host::get_balance(const address& addr) const noexcept
...
 
+namespace
+{
+// For EXTCODE* instructions if the target is an EOF account, then only return EF00.
+// While we only do this if the caller is legacy, it is not a problem doing this
+// unconditionally, because EOF contracts dot no have EXTCODE* instructions.
+bytes_view extcode(bytes_view code) noexcept
+{
+    if (code.size() >= 2 && code[0] == 0xEF && code[1] == 0x00)
+        return code.substr(2);
```
axic, line 85: Ah this is actually broken, need to get first 2 bytes, not exclude first 2 bytes.

axic:
```suggestion
        return code.substr(0, 2);
```

test/state/host.cpp
gumb0: `keccak256("ef00")` should better be memoized or hard-coded

chfast: I think this is generally fine for now although it conflicts with the TODO comment above.

test/state/host.cpp
```diff
@@ -77,24 +77,35 @@ uint256be Host::get_balance(const address& addr) const noexcept
...
 }
 
+namespace
+{
+/// For EXTCODE* instructions if the target is an EOF account, then only return EF00.
+/// While we only do this if the caller is legacy, it is not a problem doing this
+/// unconditionally, because EOF contracts dot no have EXTCODE* instructions.
+bytes_view extcode(bytes_view code, evmc_revision rev) noexcept
+{
+    return (rev >= EVMC_PRAGUE && is_eof_container(code)) ? code.substr(0, 2) : code;
```
chfast, line 87: Unless I'm missing something, the check `rev >= EVMC_PRAGUE` is not needed. The EOF prefix is picked in a way to guarantee that there is legacy code with such prefix in the state. In other words, the `is_eof_container(code)` implies `rev >= EVMC_PRAGUE`.

Maybe we should also drop the check in https://github.com/ethereum/evmone/blob/master/test/statetest/statetest_loader.cpp#L445.

pdobacz: > there is legacy code with such prefix in the state
Presuming you've meant "no legacy code with such prefix", I see the point. I think the rule is too weak though. We can warn against activating EOF EIPs on networks which might have `0xEF00` legacy bytecodes (we only our networks don't), but they should function correctly if they don't activate EOF, i.e. pre-Prague.

This is maybe dogmatic, but seems safer, WDYT?

pdobacz: Decided on the call to assume as suggested and remove the `rev` check. Also EIP-3540 will include a note that no non-EOF MAGIC prefixed code can be present on a chain which adopts the eip.

chfast (approved):

PR #588, opened 2023-03-18
Reject DELEGATECALL EOF → legacy 

axic:
The code targeted by DELEGATECALL from an EOF must also be an EOF.
This restriction has been added to EIP-3540 in
https://github.com/ethereum/EIPs/pull/7131.

### DIFF
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -2,6 +2,7 @@
 // Copyright 2019 The evmone Authors.
 // SPDX-License-Identifier: Apache-2.0
 
+#include "eof.hpp"
 #include "instructions.hpp"
 
 namespace evmone::instr::core
@@ -94,6 +95,21 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
 
     if (has_value && intx::be::load<uint256>(state.host.get_balance(state.msg->recipient)) < value)
         return {EVMC_SUCCESS, gas_left};  // "Light" failure.
+
+    if constexpr (Op == OP_DELEGATECALL)
+    {
+        if (state.rev >= EVMC_CANCUN && is_eof_container(state.original_code))
+        {
+            // The code targeted by DELEGATECALL must also be an EOF.
+            // This restriction has been added to EIP-3540 in
+            // https://github.com/ethereum/EIPs/pull/7131
+            uint8_t target_code_prefix[2];
+            const auto s = state.host.copy_code(
+                msg.code_address, 0, target_code_prefix, std::size(target_code_prefix));
+            if (!is_eof_container({target_code_prefix, s}))
+                return {EVMC_SUCCESS, gas_left};
+        }
+    }
 
     const auto result = state.host.call(msg);
     state.return_data.assign(result.output_data, result.output_size);
--- a/circle.yml
+++ b/circle.yml
@@ -455,9 +455,9 @@ jobs:
             export EVMONE_PRECOMPILES_STUB=~/project/test/state/precompiles_stub.json 
             bin/evmone-statetest ~/tests/GeneralStateTests ~/tests/LegacyTests/Constantinople/GeneralStateTests
       - download_execution_tests:
-          repo: ipsilon/tests
-          rev: eof
-          commit: b25623d4d7df10e38498cace7adc7eb413c4b20d
+          repo: ethereum/tests
+          rev: develop
+          commit: 0ec53d024c2b1f5175e70aba8f750cd4d37d9999
           legacy: false
       - run:
           name: "State tests (EOF)"
--- a/test/unittests/CMakeLists.txt
+++ b/test/unittests/CMakeLists.txt
@@ -21,6 +21,7 @@ target_sources(
     evm_eip3855_push0_test.cpp
     evm_eip3860_initcode_test.cpp
     evm_eof_test.cpp
+    evm_eof_calls_test.cpp
     evm_eof_function_test.cpp
     evm_eof_rjump_test.cpp
     evm_memory_test.cpp
--- a/test/unittests/evm_eof_calls_test.cpp
+++ b/test/unittests/evm_eof_calls_test.cpp
@@ -1,0 +1,48 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "evm_fixture.hpp"
+#include "evmone/eof.hpp"
+
+using evmone::test::evm;
+using namespace evmc::literals;
+
+TEST_P(evm, eof1_delegatecall_eof1)
+{
+    rev = EVMC_CANCUN;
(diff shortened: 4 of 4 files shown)


test/state/host.cpp
```diff
@@ -280,6 +282,17 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
 
     // Copy of the code. Revert will invalidate the account.
     const auto code = dst_acc != nullptr ? dst_acc->code : bytes{};
+
+    if (msg.kind == EVMC_DELEGATECALL)
+    {
+        // TODO: does the sender always have code here? or is this function used for EOAs too?
+        const auto sender = m_state.get(msg.sender);
```
chfast, line 289:
```suggestion
        const auto& sender = m_state.get(msg.sender);
```

test/state/host.cpp
```diff
@@ -76,15 +76,17 @@ uint256be Host::get_balance(const address& addr) const noexcept
 
 namespace
 {
+bool is_eof(bytes_view code) noexcept
```
chfast, line 79: There is `is_eof_container()` in eof.hpp

chfast (commented):
Looks good in general.

I think we should do an experiment to pass "EOF mode" flag instead of checking code prefix in many places. But this can be done after this PR to asses full impact of the re-design.

test/state/host.cpp
axic: Yup, wanted to refactor before merging.

test/state/host.cpp
```diff
@@ -266,6 +266,17 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
 
     // Copy of the code. Revert will invalidate the account.
     const auto code = dst_acc != nullptr ? dst_acc->code : bytes{};
+
+    if (msg.kind == EVMC_DELEGATECALL)
+    {
+        // TODO: does the sender always have code here? or is this function used for EOAs too?
```
gumb0, line 272: used for EOAs too, as I understand

test/state/host.cpp
```diff
@@ -266,6 +266,17 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
 
     // Copy of the code. Revert will invalidate the account.
     const auto code = dst_acc != nullptr ? dst_acc->code : bytes{};
+
+    if (msg.kind == EVMC_DELEGATECALL)
```
axic, line 270: This has to be changed since it was decided on some EOF calls that we want to only block EOF->legacy delegatecall.

test/state/host.cpp
```diff
@@ -266,6 +266,17 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
...
+    {
+        // TODO: does the sender always have code here? or is this function used for EOAs too?
+        const auto& sender = m_state.get(msg.sender);
+
+        // DELEGATECALL initiator and destination both must be either legacy or EOF
+        if (is_eof_container(code) != is_eof_container(sender.code))
+            return evmc::Result{EVMC_FAILURE};
```
chfast, line 277: This will consume all gas. I think this is therefore bad staring point. We should move the check to call instruction implementation where we have more flexibility when and how to handle the check.

lib/evmone/instructions_calls.cpp
```diff
@@ -95,6 +96,21 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
...
 
+    if constexpr (Op == OP_DELEGATECALL)
+    {
+        if (state.rev >= EVMC_CANCUN && is_eof_container(state.original_code))
+        {
+            // The code targeted by DELEGATECALL must also be an EOF.
+            // This restriction has been added to EIP-3540 in
+            // https://github.com/ethereum/EIPs/pull/6359
+            uint8_t target_code_prefix[2];
+            const auto s = state.host.copy_code(
```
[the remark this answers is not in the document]
chfast: The `s` is number of bytes copied and can be any of 0–2. Then `is_eof_container` also check the length.

test/state/host.cpp
chfast: This commit should be removed.

lib/evmone/instructions_calls.cpp
```diff
@@ -95,6 +96,21 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
...
+                msg.code_address, 0, target_code_prefix, std::size(target_code_prefix));
+            if (!is_eof_container({target_code_prefix, s}))
+                return {EVMC_SUCCESS, gas_left};
```
axic, line 110: Hm, why is this a success?

chfast: Success here means we don't revert the current call frame. But the instruction still returns `0`.

axic: Ah you mean because we already push `0` on the stack in the beginning, and we only overwrite that on L100 in case of "actual success".

lib/evmone/instructions_calls.cpp
```diff
@@ -95,6 +96,21 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
     if (has_value && intx::be::load<uint256>(state.host.get_balance(state.msg->recipient)) < value)
         return {EVMC_SUCCESS, gas_left};  // "Light" failure.
 
+    if constexpr (Op == OP_DELEGATECALL)
+    {
+        if (state.rev >= EVMC_CANCUN && is_eof_container(state.original_code))
```
axic, line 101: Assuming `original_code` is the "caller's code"?

chfast: Yes. The code DELEGATECALL instruction belongs to.

lib/evmone/instructions_calls.cpp
chfast: Yep. Common pattern used here for some other checks too.

hugo-dc, line 110: Is this consuming all gas passed to DELEGATECALL?

chfast: No. Only the amount consumed so far (i.e. call initial cost). The updated spec is in https://github.com/ethereum/EIPs/pull/7131.

lib/evmone/instructions_calls.cpp
```diff
@@ -95,6 +96,21 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
...
+        {
+            // The code targeted by DELEGATECALL must also be an EOF.
+            // This restriction has been added to EIP-3540 in
+            // https://github.com/ethereum/EIPs/pull/6359
```
axic, line 105:
```suggestion
            // https://github.com/ethereum/EIPs/pull/7131
```

chfast: Done.

test/state/host.cpp
chfast: Obsolete as the implementation has been moved to call instructions.

gumb0 (approved):

PR #589, opened 2023-03-21
test: Declarative state transition test suite

chfast:
Add unit test fixture based on state transition. The fixture is designed in a way the test cases have structure similar to JSON state tests. This should make it easier to export the test cases to JSON format in the future.

### DIFF
--- a/test/unittests/state_transition.cpp
+++ b/test/unittests/state_transition.cpp
@@ -1,0 +1,73 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "state_transition.hpp"
+
+namespace evmone::test
+{
+void state_transition::SetUp()
+{
+    pre.insert(tx.sender, {.nonce = 1, .balance = tx.gas_limit * tx.max_gas_price + tx.value + 1});
+
+    // Default expectations.
+    expect.post[Coinbase].exists = true;
+    expect.post[tx.sender].exists = true;
+}
+
+void state_transition::TearDown()
+{
+    auto& state = pre;
+    const auto res = evmone::state::transition(state, block, tx, rev, vm);
+    ASSERT_TRUE(holds_alternative<TransactionReceipt>(res))
+        << std::get<std::error_code>(res).message();
+    const auto& receipt = std::get<TransactionReceipt>(res);
+
+    EXPECT_EQ(receipt.status, expect.status);
+    if (expect.gas_used.has_value())
+    {
+        EXPECT_EQ(receipt.gas_used, *expect.gas_used);
+    }
+
+    for (const auto& [addr, expected_acc] : expect.post)
+    {
+        const auto acc = state.find(addr);
+        if (!expected_acc.exists)
+        {
+            EXPECT_EQ(acc, nullptr) << "account " << addr << " should not exist";
+        }
+        else
+        {
+            ASSERT_NE(acc, nullptr) << "account " << addr << " should exist";
+            if (expected_acc.nonce.has_value())
+            {
+                EXPECT_EQ(acc->nonce, *expected_acc.nonce);
+            }
+            if (expected_acc.balance.has_value())
+            {
+                EXPECT_EQ(acc->balance, *expected_acc.balance)
+                    << to_string(acc->balance) << " vs " << to_string(*expected_acc.balance);
+            }
+            if (expected_acc.code.has_value())
+            {
+                EXPECT_EQ(acc->code, *expected_acc.code);
+            }
+            for (const auto& [key, value] : expected_acc.storage)
+            {
+                EXPECT_EQ(acc->storage[key].current, value);
+            }
+            for (const auto& [key, value] : acc->storage)
+            {
+                // Find unexpected storage keys. This will also report entries with value 0.
+                EXPECT_TRUE(expected_acc.storage.contains(key))
+                    << "unexpected storage key " << key << "=" << value.current << " in " << addr;
+            }
+        }
+    }
+
+    for (const auto& [addr, _] : state.get_accounts())
+    {
+        EXPECT_TRUE(expect.post.contains(addr)) << "unexpected account " << addr;
+    }
+}
+}  // namespace evmone::test
--- a/test/unittests/state_transition.hpp
+++ b/test/unittests/state_transition.hpp
@@ -1,0 +1,74 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <evmone/evmone.h>
+#include <gtest/gtest.h>
+#include <test/state/host.hpp>
+
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+
+namespace evmone::test
+{
+using namespace evmone;
+using namespace evmone::state;
+
+/// Fixture to defining test cases in form similar to JSON State Tests.
+///
+/// It takes the "pre" state and produces "post" state by applying the defined "tx" transaction.
+/// Then expectations declared in "except" are checked in the "post" state.
+class state_transition : public testing::Test
+{
+protected:
+    /// The default sender address of the test transaction.
+    /// Private key: 0xa45355879.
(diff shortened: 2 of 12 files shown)


test/unittests/state_transition.hpp
```diff
@@ -0,0 +1,68 @@
...
+    struct Expectation
+    {
+        evmc_status_code status = EVMC_SUCCESS;
+        std::optional<int64_t> gas_used;
+
+        std::unordered_map<evmc::address, ExpectedAccount> post;
+    };
+
+
+    BlockInfo block{.gas_limit = 1'000'000, .base_fee = 10};
```
gumb0, line 49: Is this GCC extension to have these initializers?

chfast: No, this is part of C++20.

test/unittests/state_transition.cpp
```diff
@@ -0,0 +1,49 @@
...
+        << std::get<std::error_code>(res).message();
+    const auto& receipt = std::get<TransactionReceipt>(res);
+
+    EXPECT_EQ(receipt.status, expect.status);
+    if (expect.gas_used.has_value())
+    {
+        EXPECT_EQ(receipt.gas_used, *expect.gas_used);
+    }
+
+    for (const auto& [a, e] : expect.post)
```
winsvega, line 23: better use more descriptive names for the vars `a` and `e`

test/unittests/state_transition.cpp
```diff
@@ -0,0 +1,49 @@
...
+            ASSERT_NE(p, nullptr) << "account " << hex(a) << " should exist";
+            if (e.nonce.has_value())
+            {
+                EXPECT_EQ(p->nonce, *e.nonce);
+            }
+            if (e.balance.has_value())
+            {
+                EXPECT_EQ(p->balance, *e.balance);
+            }
+            for (const auto& [k, v] : e.storage)
```
winsvega, line 42: same here key, value  expectedAcc.storage

test/unittests/state_transition_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "state_transition.hpp"
+
+using namespace evmc::literals;
+using namespace evmone::test;
+
+TEST_F(state_transition, eof_invalid_initcode)
+{
+    static constexpr auto create_address = 0x864bbda5c698ac34b47a9ea3bd4228802cc5ce3b_address;
+
+    pre.get(*tx.to).code = eof1_bytecode(create() + push(1) + OP_SSTORE + OP_STOP, 3);
```
winsvega, line 15: the ongoing discussion is wether to have an individual tool that will verify validity of EOF 
cause strictly there are EOF codes that are valid, but represent an invalid evm code sequence. so it can't be tested through state transition

gumb0: what are you talking about, what's invalid evm code sequence?

test/unittests/state_transition.cpp
```diff
@@ -0,0 +1,49 @@
...
+                EXPECT_EQ(acc->nonce, *expected_acc.nonce);
+            }
+            if (expected_acc.balance.has_value())
+            {
+                EXPECT_EQ(acc->balance, *expected_acc.balance);
+            }
+            for (const auto& [key, value] : expected_acc.storage)
+            {
+                EXPECT_EQ(acc->storage.at(key).current, value);
+            }
```
gumb0, line 45: Should this also check for extra values in account storage that are not listed in expected?

chfast: Not sure. For accounts I wanted to use `std::nullopt` to mark an account which should not exist. For storage we have null value. I guess we can extend the checks as you suggested until proven annoying.

test/unittests/state_transition_test.cpp
winsvega: In eth rnd execution layer testing channel.

test/unittests/state_transition.cpp
chfast: Added this for now.

Same could be done for accounts and possible eliminating `.exists` flag.

test/unittests/state_transition.hpp
```diff
@@ -0,0 +1,81 @@
...
+        .max_priority_gas_price = 1,
+        .sender = Sender,
+        .to = To,
+    };
+    State pre;
+    Expectation expect;
+
+    state_transition()
+    {
+        pre.insert(
```
gumb0, line 69: Maybe this initialization should be in `SetUp()` for symmetry with `TearDown()`

test/unittests/state_transition.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    static inline evmc::VM vm{evmc_create_evmone()};
+
+    struct ExpectedAccount
+    {
+        bool exists = true;
+        std::optional<uint64_t> nonce;
+        std::optional<intx::uint256> balance;
+        std::unordered_map<bytes32, bytes32> storage;
+    };
```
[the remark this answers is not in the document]
chfast: Added.

test/unittests/state_transition.hpp
chfast: Moved.

test/unittests/state_transition.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+void state_transition::SetUp()
+{
+    pre.insert(tx.sender, {.nonce = 1, .balance = tx.gas_limit * tx.max_gas_price + tx.value + 1});
+    pre.insert(*tx.to, {.nonce = 1});
+
+    // Default expectations.
+    expect.post[Coinbase].exists = true;
+    expect.post[tx.sender].exists = true;
+    expect.post[*tx.to].exists = true;
```
gumb0, line 17: A bit annoying that to make creation transaction, we would need to unset `tx.to` but also not forget to remove it from `expect`. But maybe okay, as it's much less often than using default.

test/utils/bytecode.hpp
```diff
@@ -297,6 +297,17 @@ inline bytecode calldataload(bytecode index)
     return index + OP_CALLDATALOAD;
 }
 
+inline bytecode calldatasize()
+{
+    return OP_CALLDATASIZE;
+}
+
+inline bytecode calldatacopy(
+    bytecode src = push(0), bytecode dst = push(0), bytecode size = calldatasize())
```
gumb0, line 306: I wouldn't use defaults, seems useful to see explicitly what goes into the code (on the other hand calls and creates already have implicit defaults)

test/unittests/state_transition.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void state_transition::TearDown()
+{
+    auto& state = pre;
+    const auto res = evmone::state::transition(state, block, tx, EVMC_CANCUN, vm);
```
gumb0, line 23: Would be nice to have revision configurable.

test/unittests/state_transition_create_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+TEST_F(state_transition, eof_invalid_initcode)
+{
+    // TODO: Correction of this address is not verified.
+    static constexpr auto create_address = 0x864bbda5c698ac34b47a9ea3bd4228802cc5ce3b_address;
+
+    pre.get(*tx.to).code = eof1_bytecode(create() + push(1) + OP_SSTORE + OP_STOP, 3);
+    pre.get(*tx.to).storage[0x01_bytes32] = {.current = 0x01_bytes32, .original = 0x01_bytes32};
+
+    EXPECT_EQ(pre.get(tx.sender).balance, 1000000'001);
```
gumb0, line 32:
```suggestion
    EXPECT_EQ(pre.get(tx.sender).balance, 1'000'000'001);
```

test/unittests/state_transition.cpp
gumb0: Maybe a member function helper like `SetupCreationTransaction()` could help, but not sure.

test/utils/bytecode.hpp
chfast: Done.

test/unittests/state_transition.cpp
chfast: Added. For later I'm thinking about revision ranges support.

gumb0 (approved):
Looks good.

PR #590, opened 2023-03-21
t8n: Add tx hash calculation for tx receipt

rodiazet:
- Change is required by `execution-spec-tests`. The tool doesn't set `hash` field for transaction in `txs.json`. We have to calculate it.
- Add better error printing to console.

### DIFF
--- a/test/t8n/t8n.cpp
+++ b/test/t8n/t8n.cpp
@@ -103,11 +103,25 @@ int main(int argc, const char* argv[])
                     tx.chain_id = chain_id;
 
                     auto res = state::transition(state, block, tx, rev, vm);
+
+                    const auto computed_tx_hash = keccak256(rlp::encode(tx));
+
+                    if (j_txs[i].contains("hash"))
+                    {
+                        const auto loaded_tx_hash_opt =
+                            evmc::from_hex<bytes32>(j_txs[i]["hash"].get<std::string>());
+
+                        if (loaded_tx_hash_opt != computed_tx_hash)
+                            throw std::logic_error("transaction hash mismatched: computed " +
+                                                   hex0x(computed_tx_hash) + ", expected " +
+                                                   hex0x(loaded_tx_hash_opt.value()));
+                    }
+
                     if (holds_alternative<std::error_code>(res))
                     {
                         const auto ec = std::get<std::error_code>(res);
                         json::json j_rejected_tx;
-                        j_rejected_tx["hash"] = j_txs[i]["hash"];
+                        j_rejected_tx["hash"] = hex0x(computed_tx_hash);
                         j_rejected_tx["index"] = i;
                         j_rejected_tx["error"] = ec.message();
                         j_result["rejected"].push_back(j_rejected_tx);
@@ -120,7 +134,8 @@ int main(int argc, const char* argv[])
 
                         txs_logs.insert(txs_logs.end(), tx_logs.begin(), tx_logs.end());
                         auto& j_receipt = j_result["receipts"][j_result["receipts"].size()];
-                        j_receipt["transactionHash"] = j_txs[i]["hash"];
+
+                        j_receipt["transactionHash"] = hex0x(computed_tx_hash);
                         j_receipt["gasUsed"] = hex0x(static_cast<uint64_t>(receipt.gas_used));
                         j_receipt["cumulativeGasUsed"] = j_receipt["gasUsed"];
 
@@ -165,9 +180,10 @@ int main(int argc, const char* argv[])
 
         std::ofstream{output_dir / output_alloc_file} << std::setw(2) << j_alloc;
     }
-    catch (...)
+    catch (const std::exception& e)
     {
-        std::cerr << "Unhandled exception" << std::endl;
+        std::cerr << e.what() << std::endl;
+        return 1;
     }
 
     return 0;


test/t8n/t8n.cpp
```diff
@@ -120,7 +120,8 @@ int main(int argc, const char* argv[])
 
                         txs_logs.insert(txs_logs.end(), tx_logs.begin(), tx_logs.end());
                         auto& j_receipt = j_result["receipts"][j_result["receipts"].size()];
-                        j_receipt["transactionHash"] = j_txs[i]["hash"];
```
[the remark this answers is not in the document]
axic: `j_rejected_tx` above also uses this field, probably need to change that too?

test/t8n/t8n.cpp
```diff
@@ -165,9 +166,9 @@ int main(int argc, const char* argv[])
 
         std::ofstream{output_dir / output_alloc_file} << j_alloc;
     }
-    catch (...)
+    catch (const std::exception& e)
     {
-        std::cerr << "Unhandled exception" << std::endl;
+        std::cerr << e.what() << std::endl;
```
chfast, line 171: I don't think we should exit with `0` in this case.

test/t8n/t8n.cpp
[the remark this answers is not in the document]
chfast: The `"hash"` was added by our request. But it is fine to compute it now.
If `"hash"` is available we should compare it with the computed hash and throw an exception if mismatch.

chfast (changes requested):

test/t8n/t8n.cpp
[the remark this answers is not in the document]
rodiazet: done

test/t8n/t8n.cpp
```diff
@@ -103,11 +103,16 @@ int main(int argc, const char* argv[])
                     tx.chain_id = chain_id;
 
                     auto res = state::transition(state, block, tx, rev, vm);
+
+                    const auto tx_hash = hex0x(keccak256(rlp::encode(tx)));
+                    if (j_txs[i].contains("hash") && j_txs[i]["hash"] != tx_hash)
```
chfast, line 108: Comparing string is risky. They may differ by "0x" prefix or hex digit case. Keep them as `hash256` type and only convert to string for printing.

test/t8n/t8n.cpp
```diff
@@ -103,11 +103,16 @@ int main(int argc, const char* argv[])
...
+                        throw std::invalid_argument("hash mismatch");
```
chfast, line 109: Provide more information: computed hash, loaded hash, "transaction hash".
And maybe use some other `logic_error` exception. `std::domain_error`?

axic (approved):
Looks good to me otherwise.

chfast (changes requested):

test/t8n/t8n.cpp
```diff
@@ -103,11 +103,16 @@ int main(int argc, const char* argv[])
...
+
                     if (holds_alternative<std::error_code>(res))
                     {
                         const auto ec = std::get<std::error_code>(res);
                         json::json j_rejected_tx;
-                        j_rejected_tx["hash"] = j_txs[i]["hash"];
+                        j_rejected_tx["hash"] = tx_hash;
```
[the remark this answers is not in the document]
rodiazet: `hash` https://github.com/ethereum/retesteth/blob/master/retesteth/testStructures/types/RPC/SubElements/ToolResponseRejected.cpp#L13

test/t8n/t8n.cpp
rodiazet: Done

test/t8n/t8n.cpp
rodiazet: Done

test/t8n/t8n.cpp
```diff
@@ -103,11 +103,24 @@ int main(int argc, const char* argv[])
                     tx.chain_id = chain_id;
 
                     auto res = state::transition(state, block, tx, rev, vm);
+
+                    const auto tx_computed_hash = keccak256(rlp::encode(tx));
```
chfast, line 107:
```suggestion
                    const auto computed_tx_hash = keccak256(rlp::encode(tx));
```
Keep "tx hash" words together.

test/t8n/t8n.cpp
```diff
@@ -103,11 +103,24 @@ int main(int argc, const char* argv[])
...
+
+                    if (j_txs[i].contains("hash") && j_txs[i]["hash"].is_string())
```
chfast, line 109:
```suggestion
                    if (j_txs[i].contains("hash"))
```
Too defensive.

chfast (changes requested):

chfast (approved):

test/t8n/t8n.cpp
```diff
@@ -103,11 +103,25 @@ int main(int argc, const char* argv[])
...
+
+                    if (j_txs[i].contains("hash"))
+                    {
+                        const auto loaded_tx_hash_opt =
+                            evmc::from_hex<bytes32>(j_txs[i]["hash"].get<std::string>());
+
+                        if (loaded_tx_hash_opt != computed_tx_hash)
+                            throw std::logic_error("transaction hash mismatched: computed " +
+                                                   hex0x(computed_tx_hash) + ", expected " +
+                                                   hex0x(loaded_tx_hash_opt.value()));
```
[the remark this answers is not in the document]
rodiazet: Merged. I will add new PR.

PR #592, opened 2023-03-22
trace: Print gas and gasUsed in hex

rodiazet:


rodiazet:
> Is this a requirement of the tracing format?

No. I just want to have it similar to geth to avoid unnecessary problems when comparing tests results.

axic:
Any chance to change geth instead? Decimal makes a lot more sense for gas usage.

lib/evmone/tracing.cpp
```diff
@@ -135,8 +135,9 @@ class InstructionTracer : public Tracer
             m_out << "null";
         else
             m_out << '"' << result.status_code << '"';
-        m_out << R"(,"gas":)" << result.gas_left;
-        m_out << R"(,"gasUsed":)" << (ctx.start_gas - result.gas_left);
+        m_out << R"(,"gas":)" << std::hex << "0x" << result.gas_left;
+        m_out << R"(,"gasUsed":)" << std::hex << "0x" << (ctx.start_gas - result.gas_left)
+              << std::dec;
         m_out << R"(,"output":")" << evmc::hex({result.output_data, result.output_size}) << '"';
```
axic, line 141: Apparently this field is not an even listed as optional in the EIP, @chfast.

axic (approved):

lib/evmone/tracing.cpp
rodiazet: I can remove it and check once again if we have it implemented according to spec, but let's first confirm with @chfast

chfast: It is useful for debugging. Let's keep it.

rodiazet: @chfast Approve please if no other comments.

axic: Can one of you open a PR to the EIP adding `output` as an optional field?

chfast (changes requested):

lib/evmone/tracing.cpp
```diff
@@ -115,7 +115,7 @@ class InstructionTracer : public Tracer
         m_out << R"("pc":)" << pc;
         m_out << R"(,"op":)" << int{opcode};
         m_out << R"(,"opName":")" << get_name(opcode) << '"';
-        m_out << R"(,"gas":)" << state.gas_left;
+        m_out << R"(,"gas":)" << std::hex << "0x" << state.gas_left << std::dec;
```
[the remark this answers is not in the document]
rodiazet: done.

chfast (approved):

PR #596, opened 2023-03-25
bench: Lambda capture structured bindings (C++20)

chfast:
Refactor lambda usage to capture structured bindings (finally available in C++20) without declaring additional lambda capture variable. This also fixes clang-17 "declaration shadows a structured binding" warnings.


chfast:
This uses [P1381R1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1381r1.html) which seems to be supported only from Clang 16. See https://en.cppreference.com/w/cpp/compiler_support.

PR #598, opened 2023-03-27
Rework memory grow/check helpers - pass gas counter by value

chfast:
This is preparation to remove `gas` from `ExecutionState` and passing `gas` by value everywhere.

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -83,8 +83,9 @@ inline constexpr int64_t num_words(uint64_t size_in_bytes) noexcept
     return gas_left;
 }
 
-// Check memory requirements of a reasonable size.
-inline bool check_memory(ExecutionState& state, const uint256& offset, uint64_t size) noexcept
+/// Check memory requirements of a reasonable size.
+inline bool check_memory(
+    int64_t& gas_left, Memory& memory, const uint256& offset, uint64_t size) noexcept
 {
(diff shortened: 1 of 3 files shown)


lib/evmone/instructions.hpp
```diff
@@ -93,14 +94,15 @@ inline bool check_memory(ExecutionState& state, const uint256& offset, uint64_t
...
+        gas_left = grow_memory(gas_left, memory, new_size);
 
-    return state.gas_left >= 0;  // Always true for no-grow case.
+    return gas_left >= 0;  // Always true for no-grow case.
 }
 
-// Check memory requirements for "copy" instructions.
-inline bool check_memory(ExecutionState& state, const uint256& offset, const uint256& size) noexcept
+/// Check memory requirements for "copy" instructions.
+inline bool check_memory(
```
axic, line 104: Not sure the asymmetry between `grow_memory` and `check_memory` of how `gas_left` is used is good. Should probably keep one option, pass-by-reference or return, for both.

chfast: The `check_memory` is `inline` so it does not matter. The `grow_memory` is `noinline` so calling convention will force `int64_t& gas` to be passed via memory.

The`check_memory` can be changed too but it has much more users and the changed version wasn't looking very good. So this should be a separate piece of work. Not needed for performance though.

gumb0 (approved):

PR #602, opened 2023-03-28
evmone-t8n: Temporary fix for currentDifficulty field for exec-spec-tests

spencer-tb:
This PR adds a similar workaround to that of `prevRandao` but for `currentDifficulty`.

I think the main issue here from the exec-spec-tests format perspective relates to the `from_json<evmc::bytes32>` function. Our inputs here are string int and string hex respectively so maybe the `evmc::bytes32` can't be cast from these types.


rodiazet:
It's done by https://github.com/ethereum/evmone/pull/603

PR #603, opened 2023-03-28
statetest: Fix loading difficulty for BlockInfo

rodiazet:
Following Spencer's fix added in https://github.com/ethereum/evmone/pull/602

### DIFF
--- a/test/statetest/statetest_loader.cpp
+++ b/test/statetest/statetest_loader.cpp
@@ -75,7 +75,12 @@ address from_json<address>(const json::json& j)
 template <>
 hash256 from_json<hash256>(const json::json& j)
 {
-    return evmc::from_hex<hash256>(j.get<std::string>()).value();
+    // Special case to handle "0". Required by exec-spec-tests.
+    // TODO: Get rid of it.
+    if (j.is_string() && (j == "0" || j == "0x0"))
+        return 0x00_bytes32;
+    else
+        return evmc::from_hex<hash256>(j.get<std::string>()).value();
 }
 
 template <>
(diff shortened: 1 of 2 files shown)


test/statetest/statetest_loader.cpp
```diff
@@ -140,6 +140,20 @@ inline uint64_t calculate_current_base_fee_eip1559(
     return base_fee;
 }
 
+namespace
+{
+bytes32 load_bytes32(const json::json::const_iterator& json_it)
```
chfast, line 145:
```suggestion
bytes32 load_bytes32(const json::json& j)
```
Just take the "json" because it is unconditionally dereferenced anyway.

test/statetest/statetest_loader.cpp
```diff
@@ -140,6 +140,20 @@ inline uint64_t calculate_current_base_fee_eip1559(
...
+{
+    // Special case to handle "0". Required by exec-spec-tests.
+    // TODO: Get rid of it.
+    if (json_it->is_string() &&
+        (json_it->get<std::string>() == "0" || json_it->get<std::string>() == "0x0"))
```
chfast, line 150:
```suggestion
        (*json_it == "0" || *json_it == "0x0"))
```
I believe you can shorten this.

chfast (changes requested):

test/unittests/statetest_loader_block_info_test.cpp
```diff
@@ -91,6 +91,64 @@ TEST(statetest_loader, block_info_dec)
...
+}
+
+TEST(statetest_loader, block_info_0_parent_difficulty)
+{
+    constexpr std::string_view input = R"({
+        "currentCoinbase": "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
+        "currentGasLimit": "100000000000000000",
+        "currentNumber": "1",
+        "currentTimestamp": "1000",
+        "parentDifficulty": "0",
```
chfast, line 130: Change one of the tests to `0x0`.

test/statetest/statetest_loader.cpp
```diff
@@ -140,6 +140,19 @@ inline uint64_t calculate_current_base_fee_eip1559(
     return base_fee;
 }
 
+namespace
+{
+bytes32 load_bytes32(const json::json& j)
```
chfast, line 145: All things considered, I think we should just add this quirk to `from_json<hash256>` unconditionally. evmone is not test format sanitizer and probably never will be. So let's keep logic simple.

chfast (changes requested):

test/statetest/statetest_loader.cpp
rodiazet: OK

test/unittests/statetest_loader_block_info_test.cpp
rodiazet: done

chfast (approved):

PR #604, opened 2023-03-29
Add t8n features required by exe-spec-tests

rodiazet:


test/t8n/t8n.cpp
```diff
@@ -137,6 +138,7 @@ int main(int argc, const char* argv[])
 
                         j_receipt["transactionHash"] = hex0x(computed_tx_hash);
                         j_receipt["gasUsed"] = hex0x(static_cast<uint64_t>(receipt.gas_used));
+                        blockGasUsed += receipt.gas_used;
```
chfast, line 141: This is also `"cumulativeGasUsed"` so I propose to name the variable like this and assign correct value to the receipt. The block gas used it the value of the last receipt.

test/t8n/t8n.cpp
```diff
@@ -82,6 +82,7 @@ int main(int argc, const char* argv[])
 
         std::vector<state::Transaction> transactions;
         std::vector<state::TransactionReceipt> receipts;
+        int64_t blockGasUsed = 0;
```
chfast, line 85: Put it before vectors.

test/t8n/t8n.cpp
```diff
@@ -80,6 +83,7 @@ int main(int argc, const char* argv[])
         j_result["currentDifficulty"] = "0x20000";
         j_result["currentBaseFee"] = hex0x(block.base_fee);
 
+        int64_t cumulativeGasUsed = 0;
```
chfast, line 86:
```suggestion
        int64_t cumulative_gas_used = 0;
```

chfast (commented):
Looks good, just let's merge #603 first.

test/t8n/t8n.cpp
rodiazet: Shit. Sorry for this. Too fast as always.

chfast (approved):

PR #606, opened 2023-03-30
Support optional `storage` field in input alloc.

rodiazet:


test/statetest/statetest_loader.cpp
```diff
@@ -184,10 +184,14 @@ state::State from_json<state::State>(const json::json& j)
                 .balance = from_json<intx::uint256>(j_acc.at("balance")),
                 .code = from_json<bytes>(j_acc.at("code"))});
 
-        for (const auto& [j_key, j_value] : j_acc.at("storage").items())
+        if (j_acc.contains("storage"))
```
chfast, line 187: I prefer to use `find` so you don't have to look it up twice.

chfast (approved):
You should not put full stop in a git commit title.

PR #607, opened 2023-03-30
Support optional `to` field in txs input

rodiazet:


chfast (approved):
There usually require a unit test.

PR #609, opened 2023-04-04
state: Clean up state finalization

rodiazet:
- Separate selfdestruct handling from empty account clearing.
- Apply block reward to coinbase consistently.
- Use it to properly clean state in `t8n` and `state-test-runner`.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -117,6 +117,22 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 }
 }  // namespace
 
+void finalize(
+    State& state, evmc_revision rev, const address& coinbase, std::optional<uint64_t> block_reward)
+{
+    if (block_reward.has_value())
+        state.touch(coinbase).balance += *block_reward;
+
+    if (rev >= EVMC_SPURIOUS_DRAGON)
+    {
+        std::erase_if(
+            state.get_accounts(), [](const std::pair<const address, Account>& p) noexcept {
+                const auto& acc = p.second;
+                return acc.erasable && acc.is_empty();
+            });
+    }
+}
+
 std::variant<TransactionReceipt, std::error_code> transition(
     State& state, const BlockInfo& block, const Transaction& tx, evmc_revision rev, evmc::VM& vm)
 {
@@ -124,12 +140,7 @@ std::variant<TransactionReceipt, std::error_code> transition(
     const auto validation_result = validate_transaction(sender_acc, block, tx, rev);
 
     if (holds_alternative<std::error_code>(validation_result))
-    {
-        // Pre EIP-158 coinbase has to be touched also for invalid tx.
-        if (rev <= EVMC_TANGERINE_WHISTLE)
-            state.touch(block.coinbase);
         return get<std::error_code>(validation_result);
-    }
 
     const auto execution_gas_limit = get<int64_t>(validation_result);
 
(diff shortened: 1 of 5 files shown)


test/state/state.cpp
```diff
@@ -117,19 +117,22 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 }
 }  // namespace
 
+void clear_empty_or_destructed_accounts(State& state, evmc_revision rev)
```
chfast, line 120: image

Maybe not very specific, but can we name it `finalize()`?

rodiazet: Good idea :)

test/state/state.cpp
```diff
@@ -177,10 +180,7 @@ std::variant<TransactionReceipt, std::error_code> transition(
     state.touch(block.coinbase).balance += gas_used * priority_gas_price;
 
     // Apply destructs and clear erasable empty accounts.
```
axic, line 182: I'd move this comment to the definition of `finalize`.

rodiazet: We already have some doc comment there https://github.com/ethereum/evmone/pull/609/files#diff-5e1562f74d0f9de52fe46c9a338ee6cdbbf9b573aaac29f4232a4d8183eac44f
I can add this one too,.

test/state/state.cpp
```diff
@@ -117,19 +117,22 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 }
 }  // namespace
 
+void finalize(State& state, evmc_revision rev)
+{
+    std::erase_if(state.get_accounts(), [rev](const std::pair<const address, Account>& p) noexcept {
```
chfast, line 122: This is incorrect. It seems the implementation practice is to split handling selfdestructed accounts from empty-but-touched. We should delete selfdestructed after each transaction and clear empty-but-touched in the very end after applying block reward and withdrawals. See https://github.com/ethereum/yellowpaper/issues/885.

test/state/state.cpp
chfast: This YP part is actually wrong: https://github.com/ethereum/yellowpaper/issues/885.

test/state/state.cpp
```diff
@@ -117,19 +117,30 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
...
+
 std::variant<TransactionReceipt, std::error_code> transition(
     State& state, const BlockInfo& block, const Transaction& tx, evmc_revision rev, evmc::VM& vm)
 {
     auto& sender_acc = state.get(tx.sender);
     const auto validation_result = validate_transaction(sender_acc, block, tx, rev);
 
     if (holds_alternative<std::error_code>(validation_result))
-    {
-        // Pre EIP-158 coinbase has to be touched also for invalid tx.
```
gumb0, line 128: Why is this not needed now?

chfast: This emulates applying block reward of 0. Previously, I wasn't exactly sure why this is needed. Now the block reward is "properly" applied in the end.

test/statetest/statetest_runner.cpp
```diff
@@ -28,6 +28,10 @@ void run_state_test(const StateTransitionTest& test, evmc::VM& vm)
             validate_deployed_code(state, rev);
 
             const auto res = state::transition(state, test.block, tx, rev, vm);
+
+            // Finalize block with reward 0.
```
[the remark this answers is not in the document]
chfast: For state tests this is always 0. This effectively only touches/creates the coinbase account.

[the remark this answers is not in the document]
gumb0: True, apparently it's non-zero only in blockchain tests.

gumb0 (approved):

AungThuSoe24 (commented):
Coinbase wallet

PR #615, opened 2023-04-12
ci: Switch to new codecov uploader

chfast:
Migrates to the new codecov uploader (old one is deprecated).
Good news is that we don't need python any more.
Instead we need to download some "binaries" (project is nodejs based).
We do some integration checks, but I'm not sure how good they are.

I also had to add `CODECOV_TOKEN` to Circle CI. Some docs say it shouldn't be needed for public repos on Circle CI, but upload wasn't working and worked with the token.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -236,19 +236,27 @@ commands:
     steps:
       - run:
           name: "Install codecov"
-          command: sudo pip3 install --break-system-packages --upgrade --quiet --no-cache-dir codecov
+          command: |
+            # TODO: This should go to cpp-build-env image
+            gpg --no-default-keyring --keyring trustedkeys.gpg --keyserver keyserver.ubuntu.com --recv-key ED779869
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -236,7 +236,22 @@ commands:
     steps:
       - run:
           name: "Install codecov"
-          command: sudo pip3 install --break-system-packages --upgrade --quiet --no-cache-dir codecov
+          command: |
+            # TODO: This should go to cpp-build-env image
+            # gpg --no-default-keyring --keyring trustedkeys.gpg --keyserver keyring.debian.org --recv-key ED779869
```
chfast, line 241: I wanted to download the PGP public key by ID but this command is not working on CI (works locally).

chfast: Done. keyserver.ubuntu.com works.

PR #616, opened 2023-04-13
Add tracing output for t8n

rodiazet:
Add `--trace` option for `evmone-t8n` to output traces to specific files compatible with `retesteth`.

Modify the tracing format to follow closer the specification at [EIP-3155: EVM trace specification](https://eips.ethereum.org/EIPS/eip-3155).

### DIFF
--- a/lib/evmone/tracing.cpp
+++ b/lib/evmone/tracing.cpp
@@ -70,10 +70,12 @@ class InstructionTracer : public Tracer
 {
     struct Context
     {
+        const int32_t depth;
         const uint8_t* const code;  ///< Reference to the code being executed.
         const int64_t start_gas;
 
-        Context(const uint8_t* c, int64_t g) noexcept : code{c}, start_gas{g} {}
+        Context(int32_t d, const uint8_t* c, int64_t g) noexcept : depth{d}, code{c}, start_gas{g}
+        {}
     };
 
     std::stack<Context> m_contexts;
@@ -94,15 +96,9 @@ class InstructionTracer : public Tracer
     }
 
     void on_execution_start(
-        evmc_revision rev, const evmc_message& msg, bytes_view code) noexcept override
+        evmc_revision /*rev*/, const evmc_message& msg, bytes_view code) noexcept override
     {
-        m_contexts.emplace(code.data(), msg.gas);
-
-        m_out << "{";
-        m_out << R"("depth":)" << msg.depth;
-        m_out << R"(,"rev":")" << rev << '"';
-        m_out << R"(,"static":)" << (((msg.flags & EVMC_STATIC) != 0) ? "true" : "false");
-        m_out << "}\n";
+        m_contexts.emplace(msg.depth, code.data(), msg.gas);
     }
 
     void on_instruction_start(uint32_t pc, const intx::uint256* stack_top, int stack_height,
@@ -114,34 +110,24 @@ class InstructionTracer : public Tracer
         m_out << "{";
         m_out << R"("pc":)" << std::dec << pc;
         m_out << R"(,"op":)" << std::dec << int{opcode};
-        m_out << R"(,"opName":")" << get_name(opcode) << '"';
-        m_out << R"(,"gas":0x)" << std::hex << gas;
-        output_stack(stack_top, stack_height);
+        m_out << R"(,"gas":"0x)" << std::hex << gas << '"';
+        m_out << R"(,"gasCost":"0x)" << std::hex << instr::gas_costs[state.rev][opcode] << '"';
 
         // Full memory can be dumped as evmc::hex({state.memory.data(), state.memory.size()}),
         // but this should not be done by default. Adding --tracing=+memory option would be nice.
-        m_out << R"(,"memorySize":)" << std::dec << state.memory.size();
+        m_out << R"(,"memSize":)" << std::dec << state.memory.size();
+
+        output_stack(stack_top, stack_height);
+        if (!state.return_data.empty())
+            m_out << R"(,"returnData":"0x)" << evmc::hex(state.return_data) << '"';
+        m_out << R"(,"depth":)" << std::dec << (ctx.depth + 1);
+        m_out << R"(,"refund":)" << std::dec << state.gas_refund;
+        m_out << R"(,"opName":")" << get_name(opcode) << '"';
 
         m_out << "}\n";
     }
 
-    void on_execution_end(const evmc_result& result) noexcept override
-    {
-        const auto& ctx = m_contexts.top();
-
-        m_out << "{";
-        m_out << R"("error":)";
-        if (result.status_code == EVMC_SUCCESS)
-            m_out << "null";
-        else
-            m_out << '"' << result.status_code << '"';
-        m_out << R"(,"gas":)" << std::hex << "0x" << result.gas_left;
-        m_out << R"(,"gasUsed":)" << std::hex << "0x" << (ctx.start_gas - result.gas_left);
-        m_out << R"(,"output":")" << evmc::hex({result.output_data, result.output_size}) << '"';
-        m_out << "}\n";
-
-        m_contexts.pop();
-    }
+    void on_execution_end(const evmc_result& /*result*/) noexcept override { m_contexts.pop(); }
 
 public:
     explicit InstructionTracer(std::ostream& out) noexcept : m_out{out}
--- a/test/t8n/t8n.cpp
+++ b/test/t8n/t8n.cpp
@@ -32,6 +32,7 @@ int main(int argc, const char* argv[])
     fs::path output_body_file;
     std::optional<uint64_t> block_reward;
     uint64_t chain_id = 0;
+    bool trace = false;
 
     try
     {
@@ -64,6 +65,8 @@ int main(int argc, const char* argv[])
                 chain_id = intx::from_string<uint64_t>(argv[i]);
             else if (arg == "--output.body" && ++i < argc)
                 output_body_file = argv[i];
+            else if (arg == "--trace")
+                trace = true;
         }
 
         state::BlockInfo block;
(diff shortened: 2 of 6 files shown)


lib/evmone/tracing.cpp
```diff
@@ -149,6 +150,60 @@ class InstructionTracer : public Tracer
         m_out << std::dec;  // Set number formatting to dec, JSON does not support other forms.
     }
 };
+
+
+class StandardTracer : public InstructionTracer
```
axic, line 155: Standard according to what? Is this matching some other EIP? Please add a description to clarify.

test/t8n/t8n.cpp
```diff
@@ -106,9 +109,8 @@ int main(int argc, const char* argv[])
                     auto tx = test::from_json<state::Transaction>(j_txs[i]);
                     tx.chain_id = chain_id;
 
-                    auto res = state::transition(state, block, tx, rev, vm);
-
                     const auto computed_tx_hash = keccak256(rlp::encode(tx));
+                    const auto computed_tx_hash_str = hex0x(computed_tx_hash);
```
axic, line 113: This is a refactoring, perhaps merge it separately?

test/t8n/t8n.cpp
```diff
@@ -117,15 +119,28 @@ int main(int argc, const char* argv[])
...
 
+                    if (tracing_enabled)
+                    {
+                        // Remove old tracers
+                        vm.set_option("stdtrace", "no");
+                        auto output_filename =
+                            fs::path(output_dir / ("trace-" + std::to_string(i) + "-" +
+                                                      computed_tx_hash_str + ".jsonl"));
+                        // Add 'to file' tracer. Casting required by windows build.
+                        vm.set_option("stdtrace", (const char*)(output_filename.c_str()));
```
axic, line 134: According to [this](https://cplusplus.com/reference/string/string/c_str/) `c_str` returns `const char*` so no need to typecast?

test/t8n/t8n.cpp
```diff
@@ -117,15 +119,28 @@ int main(int argc, const char* argv[])
...
-                                                   hex0x(computed_tx_hash) + ", expected " +
+                                                   computed_tx_hash_str + ", expected " +
                                                    hex0x(loaded_tx_hash_opt.value()));
                     }
 
+                    if (tracing_enabled)
+                    {
+                        // Remove old tracers
+                        vm.set_option("stdtrace", "no");
+                        auto output_filename =
```
axic, line 130: Can we use some tmpfile feature so it gets deleted afterwards?

axic: Could use [std::tmpfile()](https://en.cppreference.com/w/cpp/io/c/tmpfile).

test/t8n/t8n.cpp
rodiazet: Unfortunately this is required by MSVC https://app.circleci.com/pipelines/github/ethereum/evmone/4313/workflows/9e36645f-85db-4644-b7e3-1efdef3e6553/jobs/46959

test/t8n/t8n.cpp
rodiazet: This returns a file and opens it. We need only a file name here. Moreover we cannot delete it because retesteth needs to read it.

test/t8n/t8n.cpp
rodiazet: How about separated commit?

test/t8n/t8n.cpp
chfast: No, you definitely cannot do it like this. You cast `wchar_t*` to `char* what is totally wrong. `std::path` has a method to convert to `std::string` from "native" string.

chfast (commented):
The tracer management seems wrong. Why are you adding multiple standard tracers? You should have one with option to change its output file.

lib/evmone/tracing.cpp
rodiazet: Added a doc comment

test/unittests/tracing_test.cpp
```diff
@@ -234,38 +226,36 @@ TEST_F(tracing, trace_revert)
...
-{
-    vm.add_tracer(evmone::create_instruction_tracer(trace_stream));
-
-    trace_stream << '\n';
-    EXPECT_EQ(trace({}, 2, EVMC_STATIC), R"(
-{"depth":2,"rev":"Berlin","static":true}
-{"error":null,"gas":0xf4240,"gasUsed":0x0,"output":""}
-)");
-}
+// TEST_F(tracing, trace_create)
```
chfast, line 238: update test.

lib/evmone/tracing.cpp
```diff
@@ -70,39 +71,36 @@ class InstructionTracer : public Tracer
...
         const int64_t start_gas;
 
-        Context(const uint8_t* c, int64_t g) noexcept : code{c}, start_gas{g} {}
+        Context(int32_t d, const uint8_t* c, int64_t g) noexcept : depth{d}, code{c}, start_gas{g}
+        {}
     };
 
     std::stack<Context> m_contexts;
     std::ostream& m_out;  ///< Output stream.
+    std::ostringstream m_buf;
```
chfast, line 84: This improved performance ~2x but I'm not sure if this is the best option. Maybe there is an option to tune `clog` itself. I'm thinking of omitting this change for now.

@gumb0 do you have any hints about iostream performance?

test/t8n/t8n.cpp
```diff
@@ -156,6 +176,9 @@ int main(int argc, const char* argv[])
                         transactions.emplace_back(std::move(tx));
                         receipts.emplace_back(std::move(receipt));
                     }
+                    // Set original std::clob buffer
```
chfast, line 179: typo.
Also this is not exception safe and probably not needed. I'd go with just leaving a comment when rdbuf is modified that we will not restore the original.

test/t8n/t8n.cpp
```diff
@@ -123,11 +127,27 @@ int main(int argc, const char* argv[])
...
+                    }
+
+                    auto res = state::transition(state, block, tx, rev, vm);
+
                     if (holds_alternative<std::error_code>(res))
                     {
                         const auto ec = std::get<std::error_code>(res);
                         json::json j_rejected_tx;
-                        j_rejected_tx["hash"] = hex0x(computed_tx_hash);
+                        j_rejected_tx["hash"] = computed_tx_hash_str;
```
chfast, line 150: This should go to prev commit?

test/t8n/t8n.cpp
```diff
@@ -123,11 +127,27 @@ int main(int argc, const char* argv[])
                                                    hex0x(loaded_tx_hash_opt.value()));
                     }
 
+                    std::ofstream trace_file_output;
+                    const auto tmp_clog_buf = std::clog.rdbuf();
+                    if (tracing_enabled)
+                    {
+                        auto output_filename =
+                            fs::path(output_dir / ("trace-" + std::to_string(i) + "-" +
```
chfast, line 135: You don't need `fs::path` explicitly as operator `/` produces a path.

test/t8n/t8n.cpp
```diff
@@ -123,11 +127,27 @@ int main(int argc, const char* argv[])
                                                    hex0x(loaded_tx_hash_opt.value()));
                     }
 
+                    std::ofstream trace_file_output;
+                    const auto tmp_clog_buf = std::clog.rdbuf();
+                    if (tracing_enabled)
+                    {
+                        auto output_filename =
```
chfast, line 134: `const auto`.

test/t8n/t8n.cpp
rodiazet: When don't do it the program is crashing randomly.

test/unittests/tracing_test.cpp
rodiazet: I think better to remove them. Tracing return empty string now.

lib/evmone/tracing.cpp
```diff
@@ -111,38 +109,31 @@ class InstructionTracer : public Tracer
...
-        m_out << "}\n";
-    }
+        m_buf << R"(,"memSize":)" << std::dec << state.memory.size();
 
-    void on_execution_end(const evmc_result& result) noexcept override
-    {
-        const auto& ctx = m_contexts.top();
+        output_stack(stack_top, stack_height);
+        if (!state.return_data.empty())
+            m_buf << R"("returnData":"0x)" << evmc::hex(state.return_data) << '"';
```
chfast, line 124:
```suggestion
            m_buf << R"(,"returnData":"0x)" << evmc::hex(state.return_data) << '"';
```
This line is not covered by tests and in the same time there is a bug in JSON syntax.

chfast (changes requested):

lib/evmone/tracing.cpp
```diff
@@ -111,38 +109,31 @@ class InstructionTracer : public Tracer
...
+        m_buf << R"(,"depth":)" << std::dec << (ctx.depth + 1);
+        m_buf << R"(,"refund":)" << std::dec << state.gas_refund;
```
chfast, line 126: goevmlab complains about this because the value can go negative, not sure what they expect to get, but they try to fit it into `uint64`.

rodiazet: This field is required be `retesteth` for every test as far as I understand.

test/t8n/t8n.cpp
rodiazet: Done

lib/evmone/tracing.cpp
rodiazet: We change the default `clog` buffer to `std::ofstream` in `t8n` implementation to redirect the output to file (to satisfy retesteth requirements). No sure then how the suggested change with `clog` tuning will work here. We probably would have to tune `std::ofstream` buffer

chfast: Yes, omit this change for now.

chfast (approved):

PR #618, opened 2023-04-18
eofparse: Allow to specify forkname

hugo-dc:
Return `err: invalid_code`  for all Forks before ~~Cancun~~ Prague.

Needed by retesteth's new validation format ([example](https://github.com/ethereum/tests/blob/c8d47d2324f686994d8b01b3550e19b3c6332677/src/EOFTestsFiller/EIP5450/validInvalidFiller.yml#L5805-L5807))

~~This branch is based on branch: `coinbase-touching-cleanup`~~

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -619,6 +619,9 @@ uint8_t get_eof_version(bytes_view container) noexcept
 
 EOFValidationError validate_eof(evmc_revision rev, bytes_view container) noexcept
 {
+    if (rev < EVMC_PRAGUE)
+        return EOFValidationError::invalid_code;
+
     if (!is_eof_container(container))
         return EOFValidationError::invalid_prefix;
 
@@ -626,9 +629,6 @@ EOFValidationError validate_eof(evmc_revision rev, bytes_view container) noexcep
 
     if (version == 1)
     {
-        if (rev < EVMC_PRAGUE)
-            return EOFValidationError::eof_version_unknown;
-
         const auto header_or_error = validate_eof1(rev, container);
         if (const auto* error = std::get_if<EOFValidationError>(&header_or_error))
             return *error;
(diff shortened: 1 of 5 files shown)


test/eofparse/eofparse.cpp
```diff
@@ -35,8 +40,23 @@ std::optional<evmc::bytes> from_hex_skip_nonalnum(InputIterator begin, InputIter
...
 {
+    evmc_revision rev = EVMC_CANCUN;
+    if (argc == 3)
+    {
+        if (std::string_view{argv[1]} != "--fork")
+        {
+            std::cerr << "Usage: " << argv[0] << " [--fork <forkname>] < EOF1" << "\n";
+            return 1;
+        }
+        rev = evmone::test::to_rev(argv[2]);
```
axic, line 53: Can't this emit an exception?

hugo-dc: It is catching the exception now.

lib/evmone/eof.cpp
```diff
@@ -619,16 +619,16 @@ uint8_t get_eof_version(bytes_view container) noexcept
 
 EOFValidationError validate_eof(evmc_revision rev, bytes_view container) noexcept
 {
+    if (rev < EVMC_PRAGUE)
```
gumb0, line 622: I'm not very convinced about this change, and whether we should test validation for pre-Prague at all.

It made a bit more sense in 2022 EOF, when normal CREATEs were used to deploy EOF, but only after EOF activation. The check verified that CREATEs don't start to allow EOF before the fork.

Now CREATE/CREATE2 must always fail with EOF, so maybe these tests don't make sense anymore.

test/eofparse/eofparse.cpp
```diff
@@ -52,22 +67,15 @@ int main()
...
-            const auto err = evmone::validate_eof(EVMC_PRAGUE, eof);
+            const auto err = evmone::validate_eof(rev, eof);
             if (err != evmone::EOFValidationError::success)
             {
                 std::cout << "err: " << evmone::get_error_message(err) << "\n";
                 continue;
             }
 
             const auto header = evmone::read_valid_eof1_header(eof);
-            std::cout << "OK ";
```
gumb0, line 63: I believe this format was used by some cross-client fuzzing, maybe we should better change retesteth to accept it.

lib/evmone/eof.cpp
hugo-dc: It makes sense, and would help to avoid previous confusions in tests.

hugo-dc:
Closing this PR. Retesteth needs to be changed in order to parse evmone-eofparse result correctly.

PR #619, opened 2023-04-19
Check stack height for OP_CALLF validation

rodiazet:
- Add validation rule.
 - Add unit tests.
 - Fix already existing unit test.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -32,6 +32,7 @@ constexpr auto CODE_SECTION_NUMBER_LIMIT = 1024;
 constexpr auto MAX_STACK_HEIGHT = 0x03FF;
 constexpr auto OUTPUTS_INPUTS_NUMBER_LIMIT = 0x7F;
 constexpr auto REL_OFFSET_SIZE = sizeof(int16_t);
+constexpr auto STACK_SIZE_LIMIT = 1024;
 
 using EOFSectionHeaders = std::array<std::vector<uint16_t>, MAX_SECTION + 1>;
 
@@ -341,17 +342,22 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         auto stack_height_required = instr::traits[opcode].stack_height_required;
         auto stack_height_change = instr::traits[opcode].stack_height_change;
 
+        auto stack_height = stack_heights[i];
+        assert(stack_height != LOC_UNVISITED);
+
         if (opcode == OP_CALLF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
 
             stack_height_required = static_cast<int8_t>(code_types[fid].inputs);
+
+            if (stack_height + code_types[fid].max_stack_height - stack_height_required >
+                STACK_SIZE_LIMIT)
+                return EOFValidationError::stack_overflow;
+
             stack_height_change =
                 static_cast<int8_t>(code_types[fid].outputs - stack_height_required);
         }
-
-        auto stack_height = stack_heights[i];
-        assert(stack_height != LOC_UNVISITED);
 
         if (stack_height < stack_height_required)
             return EOFValidationError::stack_underflow;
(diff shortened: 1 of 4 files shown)


lib/evmone/eof.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "eof.hpp"
 #include "baseline_instruction_table.hpp"
+// TODO: Remove this include. It's needed only to use StackSpace::limit
```
gumb0, line 7: Probably better to redefine it in this file.

test/unittests/evm_eof_function_test.cpp
```diff
@@ -82,9 +82,12 @@ TEST_P(evm, callf_stack_overflow)
 
     rev = EVMC_CANCUN;
     const auto code =
-        bytecode{"ef0001 010008 020002 0BFF 0007 030000 00 000003FF 00000002"_hex} +  // EOF header
-        1023 * push(1) + OP_CALLF + bytecode{"0x0001"_hex} + 1021 * OP_POP + OP_RETURN +
-        2 * push(1) + 2 * OP_POP + OP_RETF;
+        bytecode{
+            "ef0001 01000c 020003 0BFF 0007 0004 030000 00 000003FF 00000001 00000001"_hex} +  // EOF
+                                                                                               // header
```
gumb0, line 87: formatting went funky here, maybe better just delete the comment

gumb0:
State tests currently seem to have a bug https://github.com/ethereum/tests/pull/1238#discussion_r1270786226

gumb0:
Similar validation-time check should be added to JUMPF (but JUMPF is not merged yet: https://github.com/ethereum/evmone/pull/644)

lib/evmone/eof.cpp
```diff
@@ -341,18 +342,23 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
+        auto stack_height = stack_heights[i];
+        assert(stack_height != LOC_UNVISITED);
+
         if (opcode == OP_CALLF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
 
             stack_height_required = static_cast<int8_t>(code_types[fid].inputs);
+
+            if (stack_height + code_types[fid].max_stack_height - stack_height_required >
```
gumb0, line 354: In case of stack underflow this expression can be negative, I think it's ok, it's signed and won't underflow

gumb0:
Rebased.

lib/evmone/eof.cpp
gumb0: I haven't found any tests for stack underflow validation at all. I will add some at least for CALLF.

gumb0: I was wrong it cannot be negative, because callee's `max_stack_height >= inputs`

lib/evmone/eof.cpp
```diff
@@ -341,18 +342,23 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         auto stack_height_required = instr::traits[opcode].stack_height_required;
         auto stack_height_change = instr::traits[opcode].stack_height_change;
 
+        auto stack_height = stack_heights[i];
```
chfast, line 345:
```suggestion
        const auto stack_height = stack_heights[i];
```

test/unittests/eof_validation_test.cpp
```diff
@@ -733,6 +734,75 @@ TEST(eof_validation, callf_invalid_code_section_index)
         EOFValidationError::invalid_code_section_index);
 }
 
+TEST(eof_validation, callf_stack_overflow)
+{
+    {
+        auto code = eof1_bytecode(
+            512 * push(1) + OP_CALLF + bytecode{"0x0000"_hex} + 510 * OP_POP + OP_RETURN, 512);
```
chfast, line 741:
```suggestion
            512 * push(1) + OP_CALLF + "0000" + 510 * OP_POP + OP_RETURN, 512);
```
`bytecode` can consume strings and will interpret them as hex.

chfast (approved):

lib/evmone/eof.cpp
rodiazet: It's modified in line 365

PR #621, opened 2023-04-24
Print account address in state transition test failures

gumb0:


chfast (approved):

test/unittests/state_transition.cpp
```diff
@@ -42,20 +42,21 @@ void state_transition::TearDown()
...
             {
-                EXPECT_EQ(acc->nonce, *expected_acc.nonce);
+                EXPECT_EQ(acc->nonce, *expected_acc.nonce) << "account " << addr;
             }
             if (expected_acc.balance.has_value())
             {
                 EXPECT_EQ(acc->balance, *expected_acc.balance)
-                    << to_string(acc->balance) << " vs " << to_string(*expected_acc.balance);
+                    << to_string(acc->balance) << " vs " << to_string(*expected_acc.balance)
+                    << " account " << addr;
```
rodiazet, line 51: Is it possible to ass unit test here? I think I was adding in a different PR

gumb0: I doubt so, as this is a fixture code... other lines here are also not tested, they are just "covered" because they are being run during (succeeding) tests in unit test sute.

PR #630, opened 2023-05-04
Implement "CALLS 2.0"

axic:
Closes https://github.com/ethereum/evmone/issues/773

Implements [EIP-7069](https://eips.ethereum.org/EIPS/eip-7069), as part of the [greater MegaEOF spec](https://github.com/ipsilon/eof/blob/main/spec/eof.md)


lib/evmone/instructions_calls.cpp
```diff
@@ -117,6 +117,105 @@ template Result call_impl<OP_DELEGATECALL>(
...
+            cost += 25000;
+    }
+
+    if ((gas_left -= cost) < 0)
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    msg.gas = std::numeric_limits<int64_t>::max();
+    msg.gas = std::min(msg.gas, gas_left - gas_left / 64);
+
+    if (has_value)
```
axic, line 180: Can we just remove the stipend logic since we pass everything?

lib/evmone/instructions.hpp
```diff
@@ -938,6 +938,12 @@ inline constexpr auto callcode = call_impl<OP_CALLCODE>;
 inline constexpr auto delegatecall = call_impl<OP_DELEGATECALL>;
 inline constexpr auto staticcall = call_impl<OP_STATICCALL>;
 
+template <Opcode Op>
+Result newcall_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept;
```
gumb0, line 942:
```suggestion
Result call2_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept;
```

lib/evmone/instructions_opcodes.hpp
```diff
@@ -170,7 +170,10 @@ enum Opcode : uint8_t
     OP_DELEGATECALL = 0xf4,
     OP_CREATE2 = 0xf5,
 
+    OP_CALL2 = 0xf8,
```
gumb0, line 173: `RETURNCONTRACT` is currently `0xf8` in the spec, so either calls should be `f9`, `fb`, `fc`, or `RETURNCONTRACT` should be moved to `0xfc`

lib/evmone/instructions_calls.cpp
```diff
@@ -117,6 +117,105 @@ template Result call_impl<OP_DELEGATECALL>(
...
+
+        if (has_value && !state.host.account_exists(dst))
+            cost += 25000;
+    }
+
+    if ((gas_left -= cost) < 0)
+        return {EVMC_OUT_OF_GAS, gas_left};
+
+    msg.gas = std::numeric_limits<int64_t>::max();
+    msg.gas = std::min(msg.gas, gas_left - gas_left / 64);
```
gumb0, line 178: I think no sense in comparison against max here
```suggestion
    msg.gas = gas_left - gas_left / 64;
```

lib/evmone/instructions_calls.cpp
gumb0: Yes I think we should remove it.

lib/evmone/instructions_calls.cpp
```diff
@@ -117,6 +117,105 @@ template Result call_impl<OP_DELEGATECALL>(
...
+
+    if (state.msg->depth >= 1024)
+        return {EVMC_SUCCESS, gas_left};  // "Light" failure.
+
+    if (has_value && intx::be::load<uint256>(state.host.get_balance(state.msg->recipient)) < value)
+        return {EVMC_SUCCESS, gas_left};  // "Light" failure.
+
+    const auto result = state.host.call(msg);
+    state.return_data.assign(result.output_data, result.output_size);
+    if (result.status_code == EVMC_SUCCESS)
```
gumb0, line 194:
```suggestion
    stack.top() =  (result.status_code == EVMC_SUCCESS ? 0 : (result.status_code == EVMC_REVERT ? 1 : 2));
```

lib/evmone/instructions.hpp
axic: I was thinking about that too, not sure which is better, especially since we have `OP_CALL2`.

lib/evmone/instructions_calls.cpp
axic: Right, that was a remnant, fixed.

pdobacz:
Uses `tests` from https://github.com/ethereum/tests/pull/1350, which contains some changes on top of the currently used tag for testing EOF in CI `state-tests`.

pdobacz:
Also mind that the PR has been adjusted to the suggestion from here https://github.com/ethereum/EIPs/pull/8287#discussion_r1526777815

test/utils/bytecode.hpp
```diff
@@ -503,6 +503,58 @@ inline call_instruction<OP_CALLCODE> callcode(bytecode address)
...
+    bytecode m_value = 0;
+    bytecode m_input = 0;
+    bytecode m_input_size = 0;
+
+public:
+    explicit extcall_instruction(bytecode address) : m_address{std::move(address)} {}
+
+
+    template <Opcode k = kind>
+    typename std::enable_if<k == OP_EXTCALL, extcall_instruction&>::type value(bytecode v)
```
chfast, line 520: You can use C++20's `requires(k == OP_EXTCALL)`.

lib/evmone/instructions_calls.cpp
```diff
@@ -133,6 +121,115 @@ template Result call_impl<OP_DELEGATECALL>(
...
+        msg.input_data = &state.memory[input_offset];
+        msg.input_size = input_size;
+    }
+
+    auto cost = has_value ? 9000 : 0;
+
+    if constexpr (Op == OP_EXTCALL)
+    {
+        if (has_value && state.in_static_mode())
+            return {EVMC_STATIC_MODE_VIOLATION, gas_left};
```
chfast, line 172: This is not covered by tests + there are some other uncovered cases. Check the coverage report inline comments.

chfast (changes requested):

test/utils/bytecode.hpp
pdobacz: :eyes: presuming you've meant 

```
    template <Opcode k = kind>
        requires requires { k == OP_EXTCALL; }
    extcall_instruction& value(bytecode v)
```

as this compiles and works. I'll leave deciphering the docs for `requires` for later :joy: Crazy stuff

lib/evmone/instructions_calls.cpp
```diff
@@ -133,6 +121,115 @@ template Result call_impl<OP_DELEGATECALL>(
...
+        (Op == OP_EXTDELEGATECALL) ? state.msg->value : intx::be::store<evmc::uint256be>(value);
+
+    if (input_size > 0)
+    {
+        // input_offset may be garbage if input_size == 0.
+        msg.input_data = &state.memory[input_offset];
+        msg.input_size = input_size;
+    }
+
+    auto cost = has_value ? 9000 : 0;
```
gumb0, line 167: Define constant `CALL_VALUE_COST` for this?

lib/evmone/instructions_calls.cpp
```diff
@@ -133,6 +121,115 @@ template Result call_impl<OP_DELEGATECALL>(
...
+
+    if constexpr (Op == OP_EXTCALL)
+    {
+        if (has_value && state.in_static_mode())
+            return {EVMC_STATIC_MODE_VIOLATION, gas_left};
+
+        if (has_value && !state.host.account_exists(dst))
+            cost += 25000;
```
gumb0, line 175: Define `ACCOUNT_CREATION_COST` constant for this maybe?

lib/evmone/instructions_calls.cpp
```diff
@@ -133,6 +121,115 @@ template Result call_impl<OP_DELEGATECALL>(
...
+{
+    static_assert(Op == OP_EXTCALL || Op == OP_EXTDELEGATECALL || Op == OP_EXTSTATICCALL);
+
+    const auto dst = intx::be::trunc<evmc::address>(stack.pop());
+    const auto input_offset_u256 = stack.pop();
+    const auto input_size_u256 = stack.pop();
+    const auto value = (Op == OP_EXTSTATICCALL || Op == OP_EXTDELEGATECALL) ? 0 : stack.pop();
+    const auto has_value = value != 0;
+
+    stack.push(2);  // Assume (hard) failure.
```
gumb0, line 135: Also I think would be good to have named constants for status codes

lib/evmone/instructions_calls.cpp
gumb0: was this a bad idea?

lib/evmone/instructions_calls.cpp
pdobacz: heh, this invites a heated debate about naming, but let's try. I'll push a suggestion in a sec.

lib/evmone/instructions_calls.cpp
```diff
@@ -133,6 +127,109 @@ template Result call_impl<OP_DELEGATECALL>(
...
```
chfast, line 141:
```suggestion
    stack.push(EXTCALL_ABORT);  // Assume (hard) failure.
```

chfast (approved):
You should clean the git history and write better PR description

pdobacz:
> You should clean the git history and write better PR description

I don't think the git history is very useful in this case, it's a pretty atomic change, and the individual commits are mostly feedback addressing, renaming, bumps and backandforwards. This looks like a squash merge to me.

PR #632, opened 2023-05-04
Change EOF data section kind to 0x04 and activate EOF in Prague

gumb0:
Preparation to introduce container sections with kind 0x03.


axic (approved):
Looks okay to me.

lib/evmone/eof.hpp
```diff
@@ -69,6 +69,7 @@ enum class EOFValidationError
     header_terminator_missing,
     type_section_missing,
     code_section_missing,
+    unknown_section_id,
```
chfast: Why is this added?

gumb0: I remove it in the 2nd commit

chfast: I think you should remove this change from the first commit.

gumb0: 1st commit doesn't work right without it, I wanted to squash them, if second approach is okay.

chfast: I don't see where this is used in the first commit. But the second approach is ok.

gumb0:
Current status: tests from https://github.com/ethereum/tests/pull/1238 are passing, but they shouldn't be, stack overflow check is incorrect in CALLF currently and there are wrong expectations in 5450 tests.

Plan: 
- [ ] Merge the fix in evmone https://github.com/ethereum/evmone/pull/677
- [ ] Fix expectations in tests in https://github.com/ethereum/tests/pull/1238
- [ ] Rebase this, then it should be good to go

test/unittests/evm_eof_function_test.cpp
```diff
@@ -79,7 +79,7 @@ TEST_P(evm, callf_with_inputs_stack_size_1024)
         return;
 
     rev = EVMC_CANCUN;
-    const auto code = bytecode{"ef0001 010008 020002 0BFF 0004 030000 00 000003FF 03030004"_hex} +
```
chfast, line 82: This seems to belong to the prev commit.

circle.yml
```diff
@@ -474,7 +474,7 @@ jobs:
             bin/evmone-statetest ~/tests/GeneralStateTests ~/tests/LegacyTests/Constantinople/GeneralStateTests
       - download_execution_tests:
           repo: ipsilon/tests
-          rev: runtime-stackoverflow-fix
+          rev: update-tests
```
chfast, line 477: Because of the risk of rebase or branch deletion, I try to use tags, e.g. `eof-opcodes-20230619`. But it's up to you in the end.

test/unittests/evm_eof_function_test.cpp
gumb0: Fixed

circle.yml
gumb0: Made a tag

chfast (approved):

PR #637, opened 2023-05-07
ci: Upgrade compilers: GCC 13 and Clang 16

chfast:


.clang-tidy
```diff
@@ -24,6 +24,7 @@ Checks: >
   clang-analyzer-*,
   cppcoreguidelines-*,
   -cppcoreguidelines-avoid-c-arrays,
+  -cppcoreguidelines-avoid-const-or-ref-data-members,
```
yperbasis, line 27: Why disable this check?

chfast: We have some structs with `const` data members. I can also suppress the warning locally.

gumb0: Maybe would be better to suppress locally, depending on how many there are such members. I couldn't find them yet.

chfast: There are 6 instances in this file (4 consts and 2 refs): https://github.com/ethereum/evmone/blob/master/lib/evmone/tracing.cpp

chfast: I think in this context all uses are actually good and neat, so I'm a bit skeptical about this warning.

yperbasis (approved):

.clang-tidy
chfast: clang-tidy has recently added proper YAML format for these checks so we can later update the config, review the checks and at least add some comments why this is disabled.

chfast: For the reference, I noticed clang-tidy supports disabling a check in a block of code with `NOLINTBEGIN` / `NOLINTEND`. Yet this check is still not useful as it kills any kind of OOP. E.g. we have some valid uses of member as `evmc::VM& m_vm;`.

gumb0: Core guidelines' recommendation is to use a pointer in such cases (potentially wrapped in `not_null`)

PR #639, opened 2023-05-08
Add output operator for EOFValidationError to be used by Google.Test

gumb0:
This makes EOF validation tests output better failures.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -11,6 +11,7 @@
 #include <cassert>
 #include <limits>
 #include <numeric>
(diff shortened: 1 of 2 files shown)


gumb0:
> Why not just `operator<<`?

yeah, it works too, then I'll put it into `eof.hpp`

chfast:
> > Why not just `operator<<`?
> 
> yeah, it works too, then I'll put it into `eof.hpp`

Go ahead.

chfast (changes requested):

chfast (approved):

PR #640, opened 2023-05-08
Change RJUMPV immediate argument to max_index

gumb0:
Change according to https://github.com/ethereum/EIPs/pull/6546/files

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -229,10 +229,8 @@ EOFValidationError validate_instructions(
 
         if (op == OP_RJUMPV)
         {
-            const auto count = code[i + 1];
-            if (count < 1)
-                return EOFValidationError::invalid_rjumpv_count;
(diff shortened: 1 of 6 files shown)


lib/evmone/eof.cpp
```diff
@@ -264,8 +262,8 @@ bool validate_rjump_destinations(bytes_view code) noexcept
         }
         else if (op == OP_RJUMPV)
         {
-            const auto count = code[i + 1];
-            imm_size += size_t{1} /* count */ + count * REL_OFFSET_SIZE /* tbl */;
+            const auto count = static_cast<size_t>(code[i + 1] + 1);
```
chfast, line 265:
```suggestion
            const auto count = size_t{code[i + 1]} + 1;
```

chfast (changes requested):

chfast (approved):

PR #644, opened 2023-05-11
Implement JUMPF and non-returning functions

gumb0:
https://eips.ethereum.org/EIPS/eip-6206
+
https://github.com/ethereum/EIPs/pull/7330


gumb0:
Need to add the check similar to https://github.com/ethereum/evmone/pull/619

gumb0:
Current status: state tests need update to change all section 0's type to non-returning.

lib/evmone/eof.cpp
```diff
@@ -349,9 +376,30 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
             stack_height_change =
```
hugo-dc, line 376: Consider the called function can be a Non Returning Function (0x80). In this case the number `0x80` should not be considered as the total number of outputs.

gumb0: I fixed this now by treating CALLF to non-returning as if it has 0 outputs.

But I'm thinking it would be better to disallow CALLF to non-returning altogether (so that only JUMPF is allowed)

lib/evmone/advanced_instructions.cpp
```diff
@@ -269,6 +269,7 @@ constexpr std::array<instruction_exec_fn, 256> instruction_implementations = [](
     table[OP_DATALOADN] = op_undefined;
     table[OP_DATASIZE] = op_undefined;
     table[OP_DATACOPY] = op_undefined;
+    table[OP_JUMPF] = op_undefined;
```
pdobacz, line 272: requires a similar addition in `legacy_cost_tables` in `baseline_instruction_table.cpp`?

gumb0: Right, good catch, thanks. Fixed and added a test.

pdobacz (approved):

lib/evmone/instructions.hpp
```diff
@@ -1088,6 +1088,25 @@ inline code_iterator retf(StackTop /*stack*/, ExecutionState& state, code_iterat
...
+    const auto callee_required_stack_size =
+        header.types[index].max_stack_height - header.types[index].inputs;
+    if (stack_size + callee_required_stack_size > StackSpace::limit)
+    {
+        state.status = EVMC_STACK_OVERFLOW;
+        return nullptr;
+    }
+
+    const auto offset = header.code_offsets[index] - header.code_offsets[0];
+    auto code = state.analysis.baseline->executable_code;
```
chfast, line 1106: Not obvious to me why this isn't `const`.

test/unittests/eof_validation_test.cpp
```diff
@@ -811,588 +813,294 @@ TEST(eof_validation, incomplete_section_size)
...
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-        "0000000000000000000000000000000000000000";
+    auto code = bytecode{"ef0001 010ffc 0203ff"} + 1023 * bytecode{"0001"} + "040000 00" +
```
chfast, line 829: Nice compression.

lib/evmone/instructions.hpp
gumb0: Looks like it can be, fixed.

PR #645, opened 2023-05-12
Add copy_cost() helper for memory copying gas cost

gumb0:
Extracted from #586


axic (approved):
Looks good to me, but can also apply @chfast's suggestion of compacting the conditional.

PR #629, opened 2023-04-29
Implement MCOPY (EIP-5656)

axic:
Implements [EIP-5656: Memory copying instruction](https://eips.ethereum.org/EIPS/eip-5656).

PR #648, opened 2023-05-15
Use references to stack values in MCOPY

gumb0:
`pop()` returns a reference and we don't need to make an extra copy.

Previous discussion: https://github.com/ethereum/evmone/pull/629#pullrequestreview-1422204520


chfast:
This disabled inlining of `evmone::instr::core::mcopy`.

gumb0:
> This disabled inlining of `evmone::instr::core::mcopy`.

Is it strong enough reason to do extra copying of 3 `u256`s and be inconsistent with other instruction implementations?

chfast (approved):

PR #650, opened 2023-05-16
test: Modernize `RETURNDATA*` unit tests

chfast:
Introduce returndatasize() and returndatacopy() bytecode helpers and rewrite/extend unit tests for `RETURNDATASIZE` and `RETURNDATACOPY`.

### DIFF
--- a/test/utils/bytecode.hpp
+++ b/test/utils/bytecode.hpp
@@ -307,9 +307,19 @@ inline bytecode calldatasize()
     return OP_CALLDATASIZE;
 }
 
-inline bytecode calldatacopy(bytecode src, bytecode dst, bytecode size)
+inline bytecode calldatacopy(bytecode dst, bytecode src, bytecode size)
 {
-    return std::move(size) + std::move(dst) + std::move(src) + OP_CALLDATACOPY;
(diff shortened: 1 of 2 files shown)


test/utils/bytecode.hpp
```diff
@@ -312,6 +312,16 @@ inline bytecode calldatacopy(bytecode src, bytecode dst, bytecode size)
...
 }
 
+inline bytecode returndatasize()
+{
+    return OP_RETURNDATASIZE;
+}
+
+inline bytecode returndatacopy(bytecode src, bytecode dst, bytecode size)
+{
+    return std::move(size) + std::move(dst) + std::move(src) + OP_RETURNDATACOPY;
```
gumb0, line 322: it's not the right order
```suggestion
    return std::move(size) + std::move(src) + std::move(dst) + OP_RETURNDATACOPY;
```

In tests you pass dst as first argument, so it works.

Similar problem with `calldatacopy(_` above.

chfast: Ah, ok. I don't want to swap order artificially so I will just rename arguments.

gumb0 (approved):

PR #658, opened 2023-05-30
statetest: Add changes required to be executed in goevmlab

hugo-dc:
- Add `--trace-summary` CLI flag to output state root.
- Make `_info` JSON section optional (goevmlab is not providing this field).
- Implement convention for `Host::get_block_hash()` (done in #698).

### DIFF
--- a/test/statetest/statetest.cpp
+++ b/test/statetest/statetest.cpp
@@ -14,27 +14,28 @@ class StateTest : public testing::Test
 {
     fs::path m_json_test_file;
     evmc::VM& m_vm;
+    bool m_trace = false;
 
 public:
-    explicit StateTest(fs::path json_test_file, evmc::VM& vm) noexcept
-      : m_json_test_file{std::move(json_test_file)}, m_vm{vm}
+    explicit StateTest(fs::path json_test_file, evmc::VM& vm, bool trace) noexcept
+      : m_json_test_file{std::move(json_test_file)}, m_vm{vm}, m_trace{trace}
     {}
 
     void TestBody() final
     {
         std::ifstream f{m_json_test_file};
-        evmone::test::run_state_test(evmone::test::load_state_test(f), m_vm);
+        evmone::test::run_state_test(evmone::test::load_state_test(f), m_vm, m_trace);
     }
 };
 
-void register_test(const std::string& suite_name, const fs::path& file, evmc::VM& vm)
+void register_test(const std::string& suite_name, const fs::path& file, evmc::VM& vm, bool trace)
 {
     testing::RegisterTest(suite_name.c_str(), file.stem().string().c_str(), nullptr, nullptr,
         file.string().c_str(), 0,
-        [file, &vm]() -> testing::Test* { return new StateTest(file, vm); });
+        [file, &vm, trace]() -> testing::Test* { return new StateTest(file, vm, trace); });
 }
 
-void register_test_files(const fs::path& root, evmc::VM& vm)
+void register_test_files(const fs::path& root, evmc::VM& vm, bool trace)
 {
     if (is_directory(root))
     {
@@ -46,11 +47,11 @@ void register_test_files(const fs::path& root, evmc::VM& vm)
         std::sort(test_files.begin(), test_files.end());
 
         for (const auto& p : test_files)
-            register_test(fs::relative(p, root).parent_path().string(), p, vm);
+            register_test(fs::relative(p, root).parent_path().string(), p, vm, trace);
     }
     else  // Treat as a file.
     {
-        register_test(root.parent_path().string(), root, vm);
+        register_test(root.parent_path().string(), root, vm, trace);
     }
 }
 }  // namespace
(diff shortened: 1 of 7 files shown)


test/statetest/statetest_runner.cpp
```diff
@@ -33,11 +34,23 @@ void run_state_test(const StateTransitionTest& test, evmc::VM& vm)
...
                 EXPECT_TRUE(expected.exception);
 
-            EXPECT_EQ(state::mpt_hash(state.get_accounts()), expected.state_hash);
+            if (!ignore_state_root)
+                EXPECT_EQ(state::mpt_hash(state.get_accounts()), expected.state_hash);
+            else
+            {
+                std::cerr << "{";
+                std::cerr << R"("stateRoot": ")" << state::mpt_hash(state.get_accounts()) << "\"";
+                std::cerr << "}\n";
```
holiman, line 51: If you move these lines up before the clause, so they are emitted regardless .. then I don't see why `ignore_state_root` is even needed. The `jsonl` outputs are emitted on `stderr`, and the complaints about stateroot are emitted on `stdout`, so the caller can just ignore `stdout`. No need for evmone to do anything about it.

test/statetest/statetest_runner.cpp
```diff
@@ -33,11 +34,23 @@ void run_state_test(const StateTransitionTest& test, evmc::VM& vm)
             state::finalize(state, rev, test.block.coinbase, 0, {});
 
             if (holds_alternative<state::TransactionReceipt>(res))
-                EXPECT_EQ(logs_hash(get<state::TransactionReceipt>(res).logs), expected.logs_hash);
+                if (!ignore_logs)
```
holiman, line 37: I think `ignore_logs` is not needed

test/statetest/statetest.cpp
```diff
@@ -80,18 +81,21 @@ int main(int argc, char* argv[])
             ->required()
             ->check(CLI::ExistingPath);
 
-        bool trace_flag = false;
-        app.add_flag("--trace", trace_flag, "Enable EVM tracing");
+        bool trace = false;
```
rodiazet, line 84: I think the name `tracing_enabled` would be better here.

rodiazet: Same applies to `trace_summary`

test/statetest/statetest.cpp
```diff
@@ -80,18 +81,21 @@ int main(int argc, char* argv[])
...
 
         evmc::VM vm{evmc_create_evmone(), {{"O", "0"}}};
 
-        if (trace_flag)
+        if (trace)
             vm.set_option("trace", "1");
 
         for (const auto& p : paths)
-            register_test_files(p, vm);
+            register_test_files(p, vm, trace || trace_summary);
```
rodiazet, line 98: I'm not sure I get this logic. What does `trace` boolean flag mean in the `register_test_files`? Maybe some comment would clarify this.

test/state/host.cpp
```diff
@@ -323,7 +324,11 @@ bytes32 Host::get_block_hash(int64_t block_number) const noexcept
     (void)block_number;
     // TODO: This is not properly implemented, but only single state test requires BLOCKHASH
     //       and is fine with any value.
-    return {};
+
```
rodiazet, line 327: Is it temporary change until block hash PR will be merged. Right?

test/statetest/statetest_loader.cpp
```diff
@@ -395,11 +395,6 @@ void validate_deployed_code(const state::State& state, evmc_revision rev)
                         " is invalid: " + std::string(get_error_message(result)));
                 }
             }
-            else
```
rodiazet, line 394: Why do we need this change?

test/state/host.cpp
chfast: Seems we still need the fallback "fake" mode for raw state tests (used by goevmlab). The `t8n` should receive block hashes in `"blockHashes"`. @holiman ? @winsvega ?

test/statetest/statetest.cpp
chfast: I'm not sure this is good idea to have variable name different than CLI option.

test/statetest/statetest_loader.cpp
chfast: This was committed as part of https://github.com/ethereum/evmone/pull/696.

test/state/host.cpp
chfast: Moved to #698.

gumb0:
Looks good, needs rebase.

test/statetest/statetest.cpp
gumb0: Actually I'm confused too, this logic means `--trace-summary` is just an alias for `--trace` ? why do we need it?

test/statetest/statetest_runner.cpp
```diff
@@ -9,7 +9,10 @@
 
 namespace evmone::test
 {
-void run_state_test(const StateTransitionTest& test, evmc::VM& vm)
+/// Execute the state @p test using the @p vm.
+///
+/// @param trace  Output execution summary to the default trace stream.
+void run_state_test(const StateTransitionTest& test, evmc::VM& vm, bool trace)
```
gumb0, line 15: this would be much less confusing
```suggestion
void run_state_test(const StateTransitionTest& test, evmc::VM& vm, bool trace_summary)
```

test/statetest/statetest.cpp
```diff
@@ -80,18 +81,21 @@ int main(int argc, char* argv[])
             ->required()
             ->check(CLI::ExistingPath);
 
-        bool trace_flag = false;
-        app.add_flag("--trace", trace_flag, "Enable EVM tracing");
+        bool trace = false;
+        bool trace_summary = false;
+        const auto trace_opt = app.add_flag("--trace", trace, "Enable EVM tracing");
+        app.add_flag("--trace-summary", trace_summary, "Output trace summary only")
+            ->excludes(trace_opt);
```
gumb0, line 88: Does `excludes()` disallow having both options at once? I would allow it, given that only `--trace` is actually the same as `--trace --trace-summary`

chfast: I'm not sure. See `--trace --trace-summary` you can infer both are needed.

test/statetest/statetest_runner.cpp
chfast: Done.

test/statetest/statetest.cpp
chfast: I prefer more restrictive version. We can change it when there is a complain.

gumb0 (approved):

PR #667, opened 2023-06-25
Refactor transaction validation and JSON loading

chfast:


test/statetest/statetest_loader.cpp
```diff
@@ -287,13 +287,12 @@ state::Transaction from_json<state::Transaction>(const json::json& j)
     if (const auto ac_it = j.find("accessList"); ac_it != j.end())
     {
         o.access_list = from_json<state::AccessList>(*ac_it);
-        if (o.kind == state::Transaction::Kind::legacy)  // Upgrade tx type if tx has "accessList"
-            o.kind = state::Transaction::Kind::eip2930;
+        o.type = std::max(o.type, state::Transaction::Type::access_list);  // Upgrade tx type.
```
gumb0, line 290: Not sure why this change is needed, previous version made more sense to me, as I don't quite see type as an arithmetic value for which you find max...

test/state/state.cpp
```diff
@@ -53,24 +53,34 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
...
 /// Validates transaction and computes its execution gas limit (the amount of gas provided to EVM).
-/// @return  Non-negative execution gas limit for valid transaction
-///          or negative value for invalid transaction.
+/// @return  Execution gas limit or transaction validation error.
 std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
     const BlockInfo& block, const Transaction& tx, evmc_revision rev) noexcept
 {
-    if (rev < EVMC_LONDON && tx.kind == Transaction::Kind::eip1559)
-        return make_error_code(TX_TYPE_NOT_SUPPORTED);
+    switch (tx.kind)
```
gumb0: Is the goal of using switch here to hope for compiler warnings when we get non-handled types at some point?
(otherwise it's quite peculiar switch with two fallthroughs and one empty case)

test/state/state.cpp
```diff
@@ -53,24 +53,34 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
...
+        if (rev < EVMC_BERLIN)
+            return make_error_code(TX_TYPE_NOT_SUPPORTED);
+        [[fallthrough]];
+
+    case Transaction::Kind::legacy:;
+    }
 
-    if (tx.max_priority_gas_price > tx.max_gas_price)
-        return make_error_code(TIP_GT_FEE_CAP);  // Priority gas price is too high.
+    assert(tx.max_priority_gas_price <= tx.max_gas_price);
```
gumb0, line 78: So for `access_list` and `legacy` this can fail assertion now for some invalid input? I think we should still return an error.

chfast: The `access_list` and `legacy` only have "gas price" and we set `max_priority_gas_price` and `max_gas_price` to the same value. That's why the assert.

test/statetest/statetest_loader.cpp
chfast: ok

test/state/state.cpp
chfast: The goal was to group checks by transaction type so it is more visible which transaction type introduce a particular check.
The blob transactions will extend it by 4 more checks. https://github.com/ethereum/evmone/pull/668/commits/336e0f9fc08f1f27c9bb26e5330c8202afefcd44#diff-51a631fbac5b8dec0cfc6a2d9dcb58a01eedc369ac3730150d40f4eaa4f9c1e3R88-R98

gumb0 (approved):

test/state/state.hpp
```diff
@@ -125,7 +136,7 @@ struct Log
 
 struct TransactionReceipt
 {
-    Transaction::Kind kind = Transaction::Kind::legacy;
+    Transaction::Type kind = Transaction::Type::legacy;
```
gumb0, line 139: should this be `type`?
```suggestion
    Transaction::Type type = Transaction::Type::legacy;
```

chfast: Likely.

test/statetest/statetest_loader.cpp
chfast: I don't need this directly for blob transactions, but previous version only works for `legacy→access_list` upgrade. E.g. for (not needed now) `legacy→eip1559` upgrade the condition is `o.kind == legacy || o.kind==access_list`.

test/state/state.hpp
chfast: Done.

test/statetest/statetest_loader.cpp
chfast: Are we good here or should I charge back?

gumb0: I don't think it's worth to try to accomodate some non-existing future upgrades, I prefer version with `if`, but fine either way.

PR #668, opened 2023-06-26
Add BLOBHASH instruction

chfast:
Implement `BLOBHASH` instruction and add support for blob transactions in state tests.

### DIFF
--- a/test/unittests/evm_eip4844_blobhash_test.cpp
+++ b/test/unittests/evm_eip4844_blobhash_test.cpp
@@ -1,0 +1,108 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+/// This file contains EVM unit tests for the BLOBHASH instruction from EIP-4844
+/// https://eips.ethereum.org/EIPS/eip-4844
+
+#include "evm_fixture.hpp"
+
+using namespace evmc::literals;
+using evmone::test::evm;
+
+TEST_P(evm, blobhash_undefined)
+{
+    rev = EVMC_SHANGHAI;
+    execute(blobhash(0));
+    EXPECT_STATUS(EVMC_UNDEFINED_INSTRUCTION);
+}
+
+TEST_P(evm, blobhash_empty)
+{
+    rev = EVMC_CANCUN;
+    execute(blobhash(0) + ret_top());
+    EXPECT_OUTPUT_INT(0);
+
+    execute(blobhash(1) + ret_top());
+    EXPECT_OUTPUT_INT(0);
+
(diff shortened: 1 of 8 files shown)


test/state/state.cpp
```diff
@@ -48,8 +63,19 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
...
+    auto c = tx_cost + compute_tx_data_cost(rev, tx.data) +
+             compute_access_list_cost(tx.access_list) + initcode_cost;
+
+    if (tx.type == Transaction::Type::blob)
+    {
+        static constexpr auto DATA_GAS_PER_BLOB = 0x20000;
+        const auto data_gas_price = fake_exponential(1, excess_data_gas, 3338477);
+        const auto total_data_gas = DATA_GAS_PER_BLOB * tx.blob_hashes.size();
+        const auto data_fee = total_data_gas * data_gas_price;
+        c += static_cast<int64_t>(data_fee);
```
chfast, line 75: This is seems to be horribly wrong because we are adding a "fee" to the "gas". Maybe `data_gas_price` is always 1 in tests?

test/state/state.hpp
```diff
@@ -120,6 +129,7 @@ struct Transaction
     std::optional<address> to;
     intx::uint256 value;
     AccessList access_list;
+    std::vector<bytes32> blob_hashes;
```
chfast, line 132: I think we are missing `max_fee_per_data_gas` but non of the tests seems to exercise it.

test/unittests/evm_eip4844_blobhash_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+TEST_P(evm, blobhash_empty)
+{
+    rev = EVMC_CANCUN;
+    execute(blobhash(0) + ret_top());
+    EXPECT_OUTPUT_INT(0);
+
+    execute(blobhash(1) + ret_top());
+    EXPECT_OUTPUT_INT(0);
+}
```
gumb0, line 28: You could also test it with `0xff...ff`, here and in other tests

gumb0:
Looks good to me, but Silkworm EVMC mismatch error on CI

test/unittests/evm_eip4844_blobhash_test.cpp
chfast: Added.

gumb0: What about other tests?

chfast: Added to all.

chfast:
> Looks good to me, but Silkworm EVMC mismatch error on CI

Fixed now. Starting from #669 we use Silkworm with EVMC 11. It does not support `BLOBHASH` yet but is ABI-compatible and we make sure the old tests have not been broken.

gumb0 (approved):

PR #669, opened 2023-06-29
Implement EIP-1153: Transient storage

chfast:
> EIP-1153: Transient storage opcodes
> Add opcodes for manipulating state that behaves identically to storage but is discarded after every transaction

https://eips.ethereum.org/EIPS/eip-1153

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -764,6 +764,25 @@ inline Result gas(StackTop stack, int64_t gas_left, ExecutionState& /*state*/) n
     return {EVMC_SUCCESS, gas_left};
 }
 
+inline void tload(StackTop stack, ExecutionState& state) noexcept
+{
+    auto& x = stack.top();
+    const auto key = intx::be::store<evmc::bytes32>(x);
+    const auto value = state.host.get_transient_storage(state.msg->recipient, key);
+    x = intx::be::load<uint256>(value);
+}
+
+inline Result tstore(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    if (state.in_static_mode())
+        return {EVMC_STATIC_MODE_VIOLATION, 0};
+
+    const auto key = intx::be::store<evmc::bytes32>(stack.pop());
+    const auto value = intx::be::store<evmc::bytes32>(stack.pop());
+    state.host.set_transient_storage(state.msg->recipient, key, value);
+    return {EVMC_SUCCESS, gas_left};
+}
+
 inline void push0(StackTop stack) noexcept
 {
     stack.push({});
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -276,7 +276,7 @@ evmc::Result Host::call(const evmc_message& orig_msg) noexcept
         return evmc::Result{EVMC_FAILURE, orig_msg.gas};  // Light exception.
 
     auto state_snapshot = m_state;
-    auto logs_snapshot = m_logs.size();
+    const auto logs_snapshot = m_logs.size();
 
     auto result = execute_message(*msg);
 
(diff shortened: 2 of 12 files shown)


test/state/host.cpp
```diff
@@ -276,7 +276,7 @@ evmc::Result Host::call(const evmc_message& orig_msg) noexcept
         return evmc::Result{EVMC_FAILURE, orig_msg.gas};  // Light exception.
 
     auto state_snapshot = m_state;
-    auto logs_snapshot = m_logs.size();
+    const auto logs_snapshot = m_logs.size();
```
rodiazet, line 279: It look like refactoring not related to the commit.

lib/evmone/instructions.hpp
```diff
@@ -764,6 +764,24 @@ inline Result gas(StackTop stack, int64_t gas_left, ExecutionState& /*state*/) n
...
+{
+    const auto key = intx::be::store<evmc::bytes32>(stack.pop());
+    const auto value = state.host.get_transient_storage(state.msg->recipient, key);
+    stack.push(intx::be::load<uint256>(value));
+}
+
+inline Result tstore(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    if (state.in_static_mode())
+        return {EVMC_STATIC_MODE_VIOLATION, 0};
```
rodiazet, line 777: Would be good to have this like covered with unit test.

test/state/host.cpp
chfast: True, but I don't think this is worth a separate commit. The change is mostly stylistic.

lib/evmone/instructions.hpp
```diff
@@ -764,6 +764,24 @@ inline Result gas(StackTop stack, int64_t gas_left, ExecutionState& /*state*/) n
     return {EVMC_SUCCESS, gas_left};
 }
 
+inline void tload(StackTop stack, ExecutionState& state) noexcept
+{
+    const auto key = intx::be::store<evmc::bytes32>(stack.pop());
```
gumb0, line 769: `sload` gets a reference to `stack.top()` and assigns to it, without doing `push()`/`pop()`

test/unittests/instructions_test.cpp
```diff
@@ -119,6 +119,8 @@ constexpr bool instruction_only_in_evmone(evmc_revision rev, Opcode op) noexcept
     case OP_DATALOADN:
     case OP_DATASIZE:
     case OP_DATACOPY:
+    case OP_TLOAD:
```
gumb0, line 122: You could add them to EVMC already

test/utils/bytecode.hpp
```diff
@@ -333,6 +338,21 @@ inline bytecode sload(bytecode index)
     return index + OP_SLOAD;
 }
 
+inline bytecode tstore(bytecode index, bytecode value)
+{
+    return value + index + OP_TSTORE;
+}
+
+inline bytecode tstore(bytecode index)
```
gumb0, line 346: Why do you need this?

lib/evmone/instructions.hpp
chfast: Added `transient_storage_static` test.

test/unittests/instructions_test.cpp
chfast: I stopped adding these to EVMC (e.g. `MCOPY` is missing). This seems double work with little benefits. Also EVMC instruction information is not directly usable by evmone because not `constexpr`.

What do you think about removing instructions from EVMC? This is a bit of work, but can be done.

test/utils/bytecode.hpp
chfast: I reworked a unit test and this was a leftover.

test/unittests/instructions_test.cpp
gumb0: Probably fine, if nobody uses them.

gumb0:
Looks good to me, needs rebase.

gumb0 (approved):

PR #670, opened 2023-07-07
state: Fix processing withdrawals of 0

chfast:
The withdrawals should be processed before state clearing so the empty accounts touched with 0 value withdrawals are also cleared from the state.

Fixes issue reported by https://github.com/ethereum/execution-spec-tests/pull/142.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -120,8 +120,12 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 void finalize(State& state, evmc_revision rev, const address& coinbase,
     std::optional<uint64_t> block_reward, std::span<Withdrawal> withdrawals)
 {
+    // TODO: The block reward can be represented as a withdrawal.
     if (block_reward.has_value())
         state.touch(coinbase).balance += *block_reward;
+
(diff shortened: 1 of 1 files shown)


test/state/state.cpp
```diff
@@ -123,6 +123,9 @@ void finalize(State& state, evmc_revision rev, const address& coinbase,
     if (block_reward.has_value())
         state.touch(coinbase).balance += *block_reward;
 
+    for (const auto& withdrawal : withdrawals)
+        state.touch(withdrawal.recipient).balance += withdrawal.get_amount();
```
chfast, line 127: I don't remember why we implemented this in different order originally. The use of `.touch()` indicates we want the account to be cleared.

danceratopz:
I could confirm locally that this (`evmone-t8n 0.11.0-dev+commit.436f5f18`) fixes the withdrawals errors when running against `fill` as:
```
fill --evm-bin=~/code/github/ethereum/evmone/build/bin/evmone-t8n 
```
from https://github.com/ethereum/execution-spec-tests/commit/ace89a1e4af4394f1d1e3346fa5d5f43cb6c6edf.

axic (approved):

PR #673, opened 2023-07-14
EVMMAX: Base C++ API for prototyping

chfast:


include/evmmax/evmmax.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace evmmax
+{
+
+/// The modular arithmetic operations for EVMMAX (EVM Modular Arithmetic Extensions).
+template <typename UintT>
+class ModArith
+{
+public:
+    using uint = UintT;
```
axic, line 16: Ah, this naming is so confusing.

include/evmmax/evmmax.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+    const uint mod;  ///< The modulus.
+
+private:
+    const uint m_r_squared;    ///< R² % mod.
+    const uint64_t m_mod_inv;  ///< The modulus inversion.
+
+public:
+    explicit ModArith(const UintT& modulus) noexcept;
```
axic, line 25: Why `UintT` here but the new type in others?

include/evmmax/evmmax.hpp
```diff
@@ -0,0 +1,42 @@
...
+    /// Converts a value to Montgomery form.
+    uint to_mont(const uint& x) const noexcept;
+
+    /// Converts a value in Montgomery form back to normal value.
+    uint from_mont(const uint& x) const noexcept;
+
+    /// Performs a Montgomery modular multiplication. Inputs must be in Montgomery form.
+    uint mul(const uint& x, const uint& y) const noexcept;
+
+    /// Performs a modular addition.
```
axic, line 36: Are these in non-montgomery form? Or for the addition/subtraction implementation it won't matter?

axic: I wonder if we should have another type for the montgomery-form version, like `uintmont` or `montuint`.

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,113 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <evmmax/evmmax.hpp>
+
+using namespace intx;
```
axic, line 7: Doesn't this conflict with the `uint` definition we have in `evmmax`?

include/evmmax/evmmax.hpp
chfast: It doesn't matter.

lib/evmmax/evmmax.cpp
chfast: No. In evmmax.hpp we have `ModArith::uint`.

include/evmmax/evmmax.hpp
chfast: No reason. The methods were added at different times.

include/evmmax/evmmax.hpp
chfast: Like what?

chfast: This alias is not very needed for the implementation.

chfast: Removed.

include/evmmax/evmmax.hpp
chfast: Fixed.

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace evmmax
+{
+namespace
+{
+inline constexpr uint64_t mul_inv64(uint64_t base) noexcept
+{
+    uint64_t result = 1;
+    for (auto _ = 0; _ < 64; ++_)
```
gumb0, line 16:
```suggestion
    for (auto i = 0; i < 64; ++i)
```

include/evmmax/evmmax.hpp
chfast: I will add a comment.

rodiazet (approved):

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+        result *= base;
+        base *= base;
+    }
+    return result;
+}
+
+template <typename UintT>
+inline UintT compute_r_squared(const UintT& mod) noexcept
+{
+    static constexpr auto r2 = intx::uint<UintT::num_bits * 2 + 64>{1} << UintT::num_bits * 2;
```
gumb0, line 27: Where does this number come from?

chfast: Not really. This currently only works with intx although some template parameters are more generic. So in theory this works for any size although it may be impractical.

gumb0, line 27: hard to remember precedence of `<<` vs `*`
```suggestion
    static constexpr auto r2 = intx::uint<UintT::num_bits * 2 + 64>{1} << (UintT::num_bits * 2);
```

lib/evmmax/evmmax.cpp
chfast: Changed. But for the reference, C++ wants to standardize `_` as the "discard" variable name (can't find the paper right now).

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+        auto tmp = addc(t[S], c);
+        t[S] = tmp.value;
+        auto d = tmp.carry;
+
+        c = 0;
+        auto m = t[0] * m_mod_inv;
+        std::tie(c, t[0]) = addmul(t[0], m, mod[0], c);
+        for (size_t j = 1; j != S; ++j)
+        {
+            std::tie(c, t[j - 1]) = addmul(t[j], m, mod[j], c);
```
gumb0, line 71: Confused about this part, in the referenced pdf, they loop `0..s-1` and assign 
```suggestion
            std::tie(c, t[j]) = addmul(t[j], m, mod[j], c);
```

chfast: I think you need to follow the later part starting with "A slight improvement is to integrate the shifting into
the reduction as follows:"

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+    if (t >= mod)  // TODO: cannot overflow if modulus is sparse (e.g. 255 bits).
+        t -= mod;
+
+    return static_cast<UintT>(t);
+}
+
+template <typename UintT>
+UintT ModArith<UintT>::to_mont(const UintT& x) const noexcept
+{
+    return mul(x, m_r_squared);
```
gumb0, line 87: Isn't Montgomery form `aR mod N` and not `aR² mod N` ?

chfast: True, but `mul` is actually `mont_mul` which is `abR⁻¹ mod N`. Therefore: `mul(a, R²)` is `aR²R⁻¹ mod N` is `aR mod N`.

I will add a comment.

jwasinger: Montgomery reduction of an input `a` yields `aR**-1 mod N`.  So to convert `a` to Montgomery form, `R²` is passed and the extra factor of `R` cancels out.

gumb0: oh I see! missed that it calls Montgoemry multiplication. Why not call it `mont_mul()` ?

chfast: The internal representation (e.g. Montgomery form) is abstracted away in EVMMAX. The multiplication is just `MULMODX`. But `to_mont()` and `from_mont()` maps to `LOADX` and `STOREX`. Here we started with add, sub and mul, but we don't have proper load, store yet.

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <evmmax/evmmax.hpp>
+
+using namespace intx;
+
+namespace evmmax
+{
+namespace
+{
+inline constexpr uint64_t mul_inv64(uint64_t base) noexcept
```
gumb0, line 13: Can you explain this too? what does it calculate and what's the forumla?

Also, why is used only with 1 word of modulus.

chfast: I don't know exactly. I copied it from somewhere. But this is some Montgomery multiplication magic. I'm guessing this is `R⁻¹ mod N` but only most significant bits are needed.

chfast: Ok, this is used to compute "Integer N′ in [0, B − 1] such that NN′ ≡ −1 (mod B)" where B is `2^64`. This loop looks like exponentiation with exponent being `−1 (mod B)` i.e. `2^64-1`.

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+    const auto d = subc(s.value, mod);
+    return (!s.carry && d.carry) ? s.value : d.value;
+}
+
+template <typename UintT>
+UintT ModArith<UintT>::sub(const UintT& x, const UintT& y) const noexcept
+{
+    const auto d = subc(x, y);
+    const auto s = d.value + mod;
+    return (d.carry) ? s : d.value;
```
gumb0, line 109: This could calculate `s` only when needed
```suggestion
    return d.carry ? (d.value + mod) : d.value;
```

test/unittests/evmmax_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+template <typename UintT, const UintT& Mod>
+struct ModA : ModArith<UintT>
+{
+    using uint = UintT;
+    ModA() : ModArith<UintT>{Mod} {}
+};
+
+template <typename>
+class evmmax_test : public testing::Test
+{};
```
gumb0, line 36: Suggestion if possible: alias inside fixture class `TypeParam` as `ModA` or something close to `ModArith`.

gumb0: Maybe also could have an instance of it as class member.

lib/evmmax/evmmax.cpp
```diff
@@ -0,0 +1,120 @@
...
+        base *= base;
+    }
+    return result;
+}
+
+/// Compute R² % mod.
+template <typename UintT>
+inline UintT compute_r_squared(const UintT& mod) noexcept
+{
+    static constexpr auto r2 = intx::uint<UintT::num_bits * 2 + 64>{1} << (UintT::num_bits * 2);
```
gumb0, line 33:
```suggestion
    // R is 2^num_bits, R² is 2^(2*num_bits) and needs 2^(2*num_bits + 1) bits to represent, rounded to 2^(2*num_bits + 64) for intx requirements.
    static constexpr auto r2 = intx::uint<UintT::num_bits * 2 + 64>{1} << (UintT::num_bits * 2);
```

gumb0 (approved):

test/unittests/evmmax_test.cpp
chfast: There are some limitations to this. Either you need to use `this->` or `TestFixture::`. http://google.github.io/googletest/reference/testing.html#TYPED_TEST

lib/evmmax/evmmax.cpp
chfast: This is done this way on purpose to follow branchless implementation.

PR #675, opened 2023-07-20
Bump minimal Xcode version to 14.3

rodiazet:
This is to enable access to important C++20 features:
- `std::bit_cast`,
- ranges.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -70,7 +70,7 @@ executors:
   macos-xcode-min:
     resource_class: macos.x86.medium.gen2
     macos:
-      xcode: 13.4.1
+      xcode: 14.3.1
     environment:
       CMAKE_BUILD_PARALLEL_LEVEL: 4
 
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -70,7 +70,7 @@ executors:
   macos-xcode-min:
     resource_class: macos.x86.medium.gen2
     macos:
-      xcode: 13.4.1
+      xcode: 14.3.0
```
chfast, line 73: I don't think it make sense to have `14.3.0` as minimal and `14.3.1` as default. Let's stick to `14.3.1` here. But we can disable `macos-xcode-min` build for now (just add TODO to re-enable it in case we upgrade Xcode again.

chfast (changes requested):

circle.yml
rodiazet: done

chfast (approved):

PR #676, opened 2023-07-26
EOF: less restricted stack validation

gumb0:
https://notes.ethereum.org/Fk1TksgjTxOotvQNL07mEQ#Less-restricted-validation-algorithm

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -360,21 +360,16 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
     std::vector<int32_t> stack_heights(code.size(), LOC_UNVISITED);
     stack_heights[0] = code_types[func_index].inputs;
 
-    std::stack<size_t> worklist;
-    worklist.push(0);
-
-    while (!worklist.empty())
+    for (size_t i = 0; i < code.size();)
     {
-        const auto i = worklist.top();
-        worklist.pop();
-
         const auto opcode = static_cast<Opcode>(code[i]);
 
         int stack_height_required = instr::traits[opcode].stack_height_required;
         auto stack_height_change = instr::traits[opcode].stack_height_change;
 
-        auto stack_height = stack_heights[i];
-        assert(stack_height != LOC_UNVISITED);
+        const auto stack_height = stack_heights[i];
+        if (stack_height == LOC_UNVISITED)
+            return EOFValidationError::unreachable_instructions;
 
         if (opcode == OP_CALLF)
         {
@@ -428,7 +423,7 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         if (stack_height < stack_height_required)
             return EOFValidationError::stack_underflow;
 
-        stack_height += stack_height_change;
+        const auto next_stack_height = stack_height + stack_height_change;
 
         // Determine size of immediate, including the special case of RJUMPV.
         const size_t imm_size = (opcode == OP_RJUMPV) ?
@@ -439,13 +434,12 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         std::fill_n(&stack_heights[i + 1], imm_size, LOC_IMMEDIATE);
 
         // Validates the successor instruction and updates its stack height.
-        const auto validate_successor = [&stack_heights, &worklist](size_t successor_offset,
-                                            int32_t expected_stack_height) {
+        const auto visit_successor = [&stack_heights](
+                                         size_t successor_offset, int32_t expected_stack_height) {
             auto& successor_stack_height = stack_heights[successor_offset];
             if (successor_stack_height == LOC_UNVISITED)
             {
                 successor_stack_height = expected_stack_height;
-                worklist.push(successor_offset);
                 return true;
             }
             else
@@ -459,7 +453,7 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         {
             if (next >= code.size())
                 return EOFValidationError::no_terminating_instruction;
-            if (!validate_successor(next, stack_height))
+            if (!visit_successor(next, next_stack_height))
                 return EOFValidationError::stack_height_mismatch;
         }
 
@@ -468,7 +462,7 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         {
             const auto target_rel_offset = read_int16_be(&code[i + 1]);
             const auto target = static_cast<int32_t>(i) + target_rel_offset + 3;
-            if (!validate_successor(static_cast<size_t>(target), stack_height))
+            if (!visit_successor(static_cast<size_t>(target), next_stack_height))
                 return EOFValidationError::stack_height_mismatch;
         }
         else if (opcode == OP_RJUMPV)
@@ -480,17 +474,15 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
             {
                 const auto target_rel_offset = read_int16_be(&code[i + k * REL_OFFSET_SIZE + 2]);
                 const auto target = static_cast<int32_t>(next) + target_rel_offset;
-                if (!validate_successor(static_cast<size_t>(target), stack_height))
+                if (!visit_successor(static_cast<size_t>(target), next_stack_height))
                     return EOFValidationError::stack_height_mismatch;
             }
         }
+
+        i = next;
     }
 
     const auto max_stack_height = *std::max_element(stack_heights.begin(), stack_heights.end());
-
-    if (std::find(stack_heights.begin(), stack_heights.end(), LOC_UNVISITED) != stack_heights.end())
-        return EOFValidationError::unreachable_instructions;
-
     return max_stack_height;
 }
 
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -349,16 +349,23 @@ bool validate_rjump_destinations(bytes_view code) noexcept
 std::variant<EOFValidationError, int32_t> validate_max_stack_height(
     bytes_view code, size_t func_index, const std::vector<EOFCodeType>& code_types)
 {
-    assert(!code.empty());
-
-    // Special values used for detecting errors.
+    // Special value used for detecting errors.
     static constexpr int32_t LOC_UNVISITED = -1;  // Unvisited byte.
-    static constexpr int32_t LOC_IMMEDIATE = -2;  // Immediate byte.
 
     // Stack height in the header is limited to uint16_t,
     // but keeping larger size for ease of calculation.
-    std::vector<int32_t> stack_heights(code.size(), LOC_UNVISITED);
-    stack_heights[0] = code_types[func_index].inputs;
+    struct StackHeightRange
+    {
+        int32_t min = LOC_UNVISITED;
+        int32_t max = LOC_UNVISITED;
+
+        [[nodiscard]] bool visited() const noexcept { return min != LOC_UNVISITED; }
+    };
+
+    assert(!code.empty());
+
+    std::vector<StackHeightRange> stack_heights(code.size());
+    stack_heights[0] = {code_types[func_index].inputs, code_types[func_index].inputs};
 
     for (size_t i = 0; i < code.size();)
     {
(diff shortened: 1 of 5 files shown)


lib/evmone/eof.cpp
```diff
@@ -410,46 +411,58 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
+                                            size_t successor_offset,
+                                            StackHeightRange expected_stack_height) {
             auto& successor_stack_height = stack_heights[successor_offset];
-            if (successor_stack_height == LOC_UNVISITED)
+            if (successor_offset <= current_offset)  // backwards jump
             {
+                assert(successor_stack_height.min != LOC_UNVISITED);
+                assert(successor_stack_height.max != LOC_UNVISITED);
+                return expected_stack_height.min == successor_stack_height.min &&
+                       expected_stack_height.max == successor_stack_height.max;
```
gumb0, line 454: The spec could have been relaxed to
```suggestion
                return expected_stack_height.min >= successor_stack_height.min &&
                       expected_stack_height.max <= successor_stack_height.max;
```
but it was decided against this for simlicity.

pdobacz (approved):
I thought for a while what other test cases we could have. Some ideas, choose which make sense in your opinion:

1. RJUMP(IV) sequence of maximum possible length, doing hops from one to the other
2. RJUMPI(V) sequence of maximum possible length, all targeting same instruction
3. stack_height.min/max range maximally broad
4. maximum number of code sections
5. CALLF sequence of maximum possible length
6. [RJUMPI, RETF] sequence of maximum ...
7. [RJUMPI, JUMPF] sequence ...

For each of these situations test if stack validation behaves - one valid code, one invalid.

lib/evmone/eof.cpp
```diff
@@ -349,40 +349,41 @@ bool validate_rjump_destinations(bytes_view code) noexcept
...
 {
-    assert(!code.empty());
-
     // Special values used for detecting errors.
     static constexpr int32_t LOC_UNVISITED = -1;  // Unvisited byte.
-    static constexpr int32_t LOC_IMMEDIATE = -2;  // Immediate byte.
+
+    struct StackHeightRange
+    {
+        int32_t min = 0;
```
chfast, line 357: It looks it is better to init them with "unvisited". You can also move the constant inside the struct.

```suggestion
        int32_t min = LOC_UNVISITED;
```

test/unittests/eof_validation.cpp
```diff
@@ -2,17 +2,156 @@
...
+
 void eof_validation::TearDown()
 {
-    for (const auto& test_case : test_cases)
+    for (size_t i = 0; i < test_cases.size(); ++i)
     {
+        const auto& test_case = test_cases[i];
         EXPECT_EQ(evmone::validate_eof(rev, test_case.container), test_case.error)
-            << test_case.name << "\n"
+            << "test case " << i << " " << test_case.name << "\n"
```
chfast, line 119: Maybe use Go convention `test_name/index` but too late now.

lib/evmone/eof.cpp
```diff
@@ -349,40 +349,41 @@ bool validate_rjump_destinations(bytes_view code) noexcept
...
     {
-        const auto i = worklist.top();
-        worklist.pop();
-
         const auto opcode = static_cast<Opcode>(code[i]);
 
         int stack_height_required = instr::traits[opcode].stack_height_required;
         auto stack_height_change = instr::traits[opcode].stack_height_change;
 
         auto stack_height = stack_heights[i];
```
chfast, line 375:
```suggestion
        const auto stack_height = stack_heights[i];
```

gumb0: It was update on line 436, but I decided to make a separate variable `next_stack_height` for that, maybe it's cleaner.

lib/evmone/eof.cpp
gumb0: Changed the default. Decided not to move the constant, to avoid `StackHeightRange::LOC_UNVISITED` below.

lib/evmone/eof.cpp
chfast: Leave this comment in the code.

Also, I'd swap the arguments: `successor_stack_height.min <= required_stack_height`.

lib/evmone/eof.cpp
```diff
@@ -349,40 +349,41 @@ bool validate_rjump_destinations(bytes_view code) noexcept
...
         const auto opcode = static_cast<Opcode>(code[i]);
 
         int stack_height_required = instr::traits[opcode].stack_height_required;
         auto stack_height_change = instr::traits[opcode].stack_height_change;
 
-        auto stack_height = stack_heights[i];
-        assert(stack_height != LOC_UNVISITED);
+        const auto stack_height = stack_heights[i];
+        if (stack_height.min == LOC_UNVISITED)
+            return EOFValidationError::unreachable_instructions;
```
chfast, line 377: This is tricky. Please explain in a comment.

lib/evmone/eof.cpp
```diff
@@ -349,40 +349,41 @@ bool validate_rjump_destinations(bytes_view code) noexcept
...
-    static constexpr int32_t LOC_IMMEDIATE = -2;  // Immediate byte.
 
     // Stack height in the header is limited to uint16_t,
     // but keeping larger size for ease of calculation.
-    std::vector<int32_t> stack_heights(code.size(), LOC_UNVISITED);
-    stack_heights[0] = code_types[func_index].inputs;
+    struct StackHeightRange
+    {
+        int32_t min = LOC_UNVISITED;
+        int32_t max = LOC_UNVISITED;
```
chfast, line 360:
```suggestion
        int32_t max = LOC_UNVISITED;
        
        bool visited() const {...}
```

This should simplify some `LOC_VISITED` handling later.

test/unittests/eof_stack_validation_test.cpp
```diff
@@ -12,145 +12,1214 @@
...
+{
+    add_test_case(
+        eof_bytecode(bytecode{OP_STOP} + OP_STOP), EOFValidationError::unreachable_instructions);
+
+    add_test_case(
+        eof_bytecode(rjump(1) + OP_STOP + OP_STOP), EOFValidationError::unreachable_instructions);
+
+    // STOP reachable only via backwards jump - invalid
+    add_test_case(
+        eof_bytecode(rjump(1) + OP_STOP + rjump(-4)), EOFValidationError::unreachable_instructions);
```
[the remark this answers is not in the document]
gumb0: I think we wouldn't be able to distinguish really unreachable instructions from those reachable only via backwards jump (because we validate in a linear forwards pass)

So we could only make this error message more general, something like `instruction_not_reachabe_via_forward_control_flow` or even more general `invalid_instruction_order`. I don't have better ideas yet.

test/unittests/eof_stack_validation_test.cpp
```diff
@@ -12,145 +12,1214 @@
...
+}
+
+TEST_F(eof_validation, no_terminating_instruction)
```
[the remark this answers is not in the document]
gumb0: No, it's zero size section error, detected during header validation.

test/unittests/eof_stack_validation_test.cpp
```diff
@@ -12,145 +12,1214 @@
...
+{
+    add_test_case(eof_bytecode(push0()), EOFValidationError::no_terminating_instruction);
+
+    add_test_case(eof_bytecode(add(1, 2)), EOFValidationError::no_terminating_instruction);
+
+    add_test_case(eof_bytecode(rjumpi(-5, 1)), EOFValidationError::no_terminating_instruction);
+}
+
+TEST_F(eof_validation, non_constant_stack_height)
```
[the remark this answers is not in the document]
gumb0: Yeah it will break many things, including many relative jumps tests that assume `rjump(offset, 0)` is 5 bytes.

`push0()` was added I think to make possible things like `1023 * push0()`

(this particular test was written even before `push0()`)

lib/evmone/eof.cpp
gumb0: Related to https://github.com/ethereum/evmone/pull/676/files/f424a769d99d8fd96b0f5d908a8194ace20d3b39#r1496139888, maybe error name is not ideal, but not quite sure how to improve.

I'll add some comment.

lib/evmone/eof.cpp
```diff
@@ -410,46 +411,58 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
-
         // Validates the successor instruction and updates its stack height.
-        const auto validate_successor = [&stack_heights, &worklist](size_t successor_offset,
-                                            int32_t expected_stack_height) {
+        const auto validate_successor = [&stack_heights](size_t current_offset,
+                                            size_t successor_offset,
+                                            StackHeightRange expected_stack_height) {
             auto& successor_stack_height = stack_heights[successor_offset];
-            if (successor_stack_height == LOC_UNVISITED)
+            if (successor_offset <= current_offset)  // backwards jump
```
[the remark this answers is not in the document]
gumb0: It's not a no-op (good that tests detected this), I've added a comment why.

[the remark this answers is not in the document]
chfast: By no-op I mean that the validity condition is always true because in this case `successor_stack_height` is `required_stack_height`. Or I'm missing something.

We could handle this separately for clarity and test coverage visibility. As follows:
```cpp
if (successor_offset == current_offset) // self-referencing jump
    return true;
```

But I'm insisting on this. I'm mostly verifying my understanding.

lib/evmone/eof.cpp
```diff
@@ -410,46 +416,63 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
             {
-                successor_stack_height = expected_stack_height;
-                worklist.push(successor_offset);
-                return true;
+                // successor_offset == current_offset case is possible only with jump into the same
+                // jump instruction, e.g. RJUMP(-3), so it is technically a backwards jump, too.
+                assert(successor_stack_height.visited());
+                // The spec could have been relaxed to
+                // return successor_stack_height.min <= required_stack_height.min &&
+                //        required_stack_height.max <= successor_stack_height.max;
```
chfast, line 461:
```suggestion
                //        successor_stack_height.max <= required_stack_height.max;
```

lib/evmone/eof.cpp
```diff
@@ -410,46 +416,63 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
+            if (successor_offset <= current_offset)  // backwards jump
             {
-                successor_stack_height = expected_stack_height;
-                worklist.push(successor_offset);
-                return true;
+                // successor_offset == current_offset case is possible only with jump into the same
+                // jump instruction, e.g. RJUMP(-3), so it is technically a backwards jump, too.
+                assert(successor_stack_height.visited());
+                // The spec could have been relaxed to
+                // return successor_stack_height.min <= required_stack_height.min &&
```
chfast, line 460:
```suggestion
                // return successor_stack_height.min >= required_stack_height.min &&
```
I believe the computed stack height (successor's) must fit into the required one.

test/unittests/eof_stack_validation_test.cpp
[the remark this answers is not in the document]
chfast: Let's leave it for later.

lib/evmone/eof.cpp
chfast: Now you can hide the constant in the `StackHeightRange::UNVISITED` :)

lib/evmone/eof.cpp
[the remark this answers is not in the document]
gumb0: Yes, you're missing something as I tried to explain in the comment. The stack height still needs to be checked in this case.

Here's the code that would be broken: `PUSH0 RJUMPI(-3) STOP` - RJUMPI(-3) generates equality case, and without the check it would be valid, but it is invalid because stack after RJUMPI is not equal to stack before RJUMPI (requires 1 item)

lib/evmone/eof.cpp
gumb0: Look like static members not allowed in local structs.
```
/home/andrei/dev/evmone/lib/evmone/eof.cpp:357:30: error: static data member 'LOC_UNVISITED' not allowed in local struct 'StackHeightRange'
        static const int32_t LOC_UNVISITED = -1;
                             ^
1 error generated.
```

chfast (approved):

chfast (approved):

PR #677, opened 2023-07-28
eof: Fix CALLF runtime stack overflow check

gumb0:


gumb0:
State tests have incorrect expectations https://github.com/ethereum/tests/pull/1238/files#r1279089894

lib/evmone/instructions.hpp
```diff
@@ -1025,7 +1025,8 @@ inline code_iterator callf(StackTop stack, ExecutionState& state, code_iterator
     const auto index = read_uint16_be(&pos[1]);
     const auto& header = state.analysis.baseline->eof_header;
     const auto stack_size = &stack.top() - state.stack_space.bottom();
-    if (stack_size + header.types[index].max_stack_height > StackSpace::limit)
+    if (stack_size + header.types[index].max_stack_height - header.types[index].inputs >
```
chfast, line 1028: Maybe add variable `const auto callee_required_stack_size = header.types[index].max_stack_height - header.types[index].inputs;`

chfast (commented):
The fix description should also go to the commit message.

lib/evmone/instructions.hpp
gumb0: Added.

PR #678, opened 2023-08-08
EOFTests

gzanitti:
This PR adds the necessary code to run the EOF tests.

Note: The value of the exception field is not validated, following what we discussed in our first meeting.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -493,6 +493,11 @@ jobs:
           working_directory: ~/build
           command: |
             bin/evmone-statetest ~/tests/EIPTests/StateTests/stEOF
+      - run:
+          name: "EOF validation tests"
+          working_directory: ~/build
+          command: |
+            bin/evmone-eoftest ~/tests/EOFTests
       - collect_coverage_gcc
       - upload_coverage:
           flags: statetests
--- a/circle.yml
+++ b/circle.yml
@@ -486,7 +486,7 @@ jobs:
             ~/tests/EIPTests/StateTests/stEIP1153-transientStorage
       - download_execution_tests:
           repo: ipsilon/tests
-          rev: eof-rjumpv-20230803
+          rev: eof-update-20230828
           legacy: false
       - run:
           name: "State tests (EOF)"
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -20,10 +20,11 @@ add_subdirectory(integration)
 add_subdirectory(internal_benchmarks)
 add_subdirectory(state)
 add_subdirectory(statetest)
+add_subdirectory(eoftest)
 add_subdirectory(t8n)
 add_subdirectory(unittests)
 
-set(targets evmone-bench evmone-bench-internal evmone-eofparse evmone-state evmone-statetest evmone-t8n evmone-unittests)
+set(targets evmone-bench evmone-bench-internal evmone-eofparse evmone-state evmone-statetest evmone-eoftest evmone-t8n evmone-unittests)
 
 if(EVMONE_FUZZING)
     add_subdirectory(eofparsefuzz)
--- a/test/eoftest/eoftest.cpp
+++ b/test/eoftest/eoftest.cpp
@@ -1,0 +1,87 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+
+#include "eoftest.hpp"
+#include <CLI/CLI.hpp>
+#include <evmone/evmone.h>
+#include <gtest/gtest.h>
+#include <filesystem>
+
+namespace fs = std::filesystem;
+
+namespace
+{
+
+class EOFTest : public testing::Test
+{
+    fs::path m_json_test_file;
+
+public:
+    explicit EOFTest(fs::path json_test_file) noexcept : m_json_test_file{std::move(json_test_file)}
+    {}
+
+    void TestBody() final
+    {
+        std::ifstream f{m_json_test_file};
+        evmone::test::run_eof_test(f);
+    }
+};
+
+void register_test(const std::string& suite_name, const fs::path& file)
+{
+    testing::RegisterTest(suite_name.c_str(), file.stem().string().c_str(), nullptr, nullptr,
+        file.string().c_str(), 0, [file]() -> testing::Test* { return new EOFTest(file); });
+}
+
+void register_test_files(const fs::path& root)
+{
+    if (is_directory(root))
+    {
+        std::vector<fs::path> test_files;
+        std::copy_if(fs::recursive_directory_iterator{root}, fs::recursive_directory_iterator{},
+            std::back_inserter(test_files), [](const fs::directory_entry& entry) {
+                return entry.is_regular_file() && entry.path().extension() == ".json";
+            });
+        std::sort(test_files.begin(), test_files.end());
+
+        for (const auto& p : test_files)
+            register_test(fs::relative(p, root).parent_path().string(), p);
+    }
+    else  // Treat as a file.
+    {
+        register_test(root.parent_path().string(), root);
+    }
+}
+
+}  // namespace
+
+
+int main(int argc, char* argv[])
+{
+    try
+    {
+        testing::InitGoogleTest(&argc, argv);
+        CLI::App app{"evmone eof test runner"};
+
+        std::vector<std::string> paths;
+        app.add_option("path", paths, "Path to test file or directory")
+            ->required()
+            ->check(CLI::ExistingPath);
+
+        CLI11_PARSE(app, argc, argv);
+
+        for (const auto& p : paths)
+        {
+            register_test_files(p);
+        }
+
+        return RUN_ALL_TESTS();
+    }
+    catch (const std::exception& ex)
+    {
+        std::cerr << ex.what() << "\n";
+        return -1;
+    }
+}
--- a/test/eoftest/eoftest.hpp
+++ b/test/eoftest/eoftest.hpp
@@ -1,0 +1,13 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <iosfwd>
+
+namespace evmone::test
+{
+
+void run_eof_test(std::istream& input);
+
+}  // namespace evmone::test
--- a/test/eoftest/eoftest_runner.cpp
+++ b/test/eoftest/eoftest_runner.cpp
@@ -1,0 +1,78 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "../utils/utils.hpp"
+#include "eoftest.hpp"
+#include <evmc/evmc.hpp>
+#include <evmone/eof.hpp>
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+
+namespace json = nlohmann;
+
+namespace evmone::test
+{
+
+namespace
+{
+struct EOFValidationTest
+{
+    struct Case
+    {
+        struct Expectation
+        {
+            evmc_revision rev;
+            bool result;
+        };
+        std::string name;
+        evmc::bytes code;
+        std::vector<Expectation> expectations;
+    };
+    std::unordered_map<std::string, Case> cases;
+};
+
+void from_json(const json::json& j, EOFValidationTest::Case& o)
+{
+    const auto op_code = evmc::from_hex(j.at("code").get<std::string>());
+    if (!op_code)
+        throw std::invalid_argument{"code is invalid hex string"};
+    o.code = *op_code;
+
+    for (const auto& [rev, result] : j.at("results").items())
+    {
+        o.expectations.push_back({to_rev(rev), result.at("result").get<bool>()});
+    }
+}
+
+void from_json(const json::json& j, EOFValidationTest& o)
+{
+    if (!j.is_object() || j.empty())
+        throw std::invalid_argument{"JSON test must be an object with single key of the test name"};
+
+    const auto& j_t = *j.begin();  // Content is in a dict with the test name.
+
+    for (const auto& [name, test] : j_t.at("vectors").items())
+    {
+        o.cases.emplace(name, test.get<EOFValidationTest::Case>());
+    }
+}
+
+}  // namespace
+
+void run_eof_test(std::istream& input)
+{
+    const auto test = json::json::parse(input).get<EOFValidationTest>();
+    for (const auto& [name, cases] : test.cases)
+    {
+        for (const auto& expectation : cases.expectations)
+        {
+            const auto result = evmone::validate_eof(expectation.rev, cases.code);
+            const bool b_result = (result == EOFValidationError::success);
+            EXPECT_EQ(b_result, expectation.result)
+                << name << " " << expectation.rev << " " << hex(cases.code);
+        }
+    }
+}
+
+}  // namespace evmone::test
--- a/test/utils/utils.cpp
+++ b/test/utils/utils.cpp
@@ -1,0 +1,45 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "utils.hpp"
+
+namespace evmone::test
+{
+
+evmc_revision to_rev(std::string_view s)
+{
+    if (s == "Frontier")
+        return EVMC_FRONTIER;
+    if (s == "Homestead")
+        return EVMC_HOMESTEAD;
+    if (s == "EIP150")
+        return EVMC_TANGERINE_WHISTLE;
+    if (s == "EIP158")
+        return EVMC_SPURIOUS_DRAGON;
+    if (s == "Byzantium")
+        return EVMC_BYZANTIUM;
+    if (s == "Constantinople")
+        return EVMC_CONSTANTINOPLE;
+    if (s == "ConstantinopleFix")
+        return EVMC_PETERSBURG;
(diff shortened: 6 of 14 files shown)


test/eoftests/eoftests_runner.cpp
```diff
@@ -88,20 +87,16 @@ void from_json(const json::json& j, EOFTests& o)
...
 void from_json(const json::json& j, EOFTests::Case& o)
 {
-    std::string s{j.at("code").get<std::string>()};
+    std::string hex_str{j.at("code").get<std::string>()};
 
-    std::vector<uint8_t> vec;
-    for (std::size_t i = 2; i < s.length(); i += 2)
+    std::vector<uint8_t> bytes;
+
+    for (std::size_t i = 2; i < hex_str.size(); i += 2)
```
gumb0, line 94: Use `evmc::from_hex()` similar to https://github.com/ethereum/evmone/blob/095f83a3931c012af43e6580d188d4bb88af5d45/test/statetest/statetest_loader.cpp#L68

gumb0:
Please squash the commits

test/eoftests/eoftests_runner.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        for (const auto& expectation : cases.expectations)
+        {
+            const EOFValidationError result = evmone::validate_eof(expectation.rev, cases.code);
+            bool b_result = true;
+            if (result != EOFValidationError::success)
+            {
+                b_result = false;
+            }
+            // else
```
gumb0, line 27: Please delete commented out code

test/eoftests/eoftests_runner.cpp
```diff
@@ -0,0 +1,107 @@
...
+    for (const auto& cases : test.cases)
+    {
+        for (const auto& expectation : cases.expectations)
+        {
+            const EOFValidationError result = evmone::validate_eof(expectation.rev, cases.code);
+            bool b_result = true;
+            if (result != EOFValidationError::success)
+            {
+                b_result = false;
+            }
```
gumb0, line 26:
```suggestion
            const bool b_result = (result == EOFValidationError::success);
```

test/eoftests/eoftests_runner.cpp
```diff
@@ -0,0 +1,107 @@
...
+            // else
+            //{
+            //     EXPECT_EQ(get_error_message(result), expectation.exception);
+            // }
+            EXPECT_EQ(b_result, expectation.result);
+        }
+    }
+}
+
+evmc_revision to_rev(std::string_view s)
```
gumb0, line 36: We shouldn't duplicate this function. Please move it to `test/utils` and make it shared between this tool and statetest

test/CMakeLists.txt
```diff
@@ -20,10 +20,11 @@ add_subdirectory(integration)
 add_subdirectory(internal_benchmarks)
 add_subdirectory(state)
 add_subdirectory(statetest)
+add_subdirectory(eoftests)
 add_subdirectory(t8n)
 add_subdirectory(unittests)
 
-set(targets evmone-bench evmone-bench-internal evmone-eofparse evmone-state evmone-statetest evmone-t8n evmone-unittests)
+set(targets evmone-bench evmone-bench-internal evmone-eofparse evmone-state evmone-statetest evmone-eoftests evmone-t8n evmone-unittests)
```
gumb0, line 27: Please keep these in alphabetical order

test/eoftests/eoftests_runner.cpp
gzanitti: Sorry, I left them to remind myself that the code wasn't finished yet haha. Removed.

test/eoftests/eoftests_runner.cpp
```diff
@@ -0,0 +1,61 @@
...
+void run_eof_test(std::istream& input)
+{
+    const EOFTests test = json::json::parse(input).get<EOFTests>();
+    for (const auto& [name, cases] : test.cases)
+    {
+        for (const auto& expectation : cases.expectations)
+        {
+            const EOFValidationError result = evmone::validate_eof(expectation.rev, cases.code);
+            const bool b_result = (result == EOFValidationError::success);
+            EXPECT_EQ(b_result, expectation.result);
```
gumb0, line 24: use `operator<<` to output code and revision in case of failure, something like:
```suggestion
            EXPECT_EQ(b_result, expectation.result) << evmc::to_string(expectation.rev) << " " << cases.code;
```

gzanitti: Done :)

gumb0:
Please squash the commits

circle.yml
```diff
@@ -481,6 +481,11 @@ jobs:
           working_directory: ~/build
           command: |
             bin/evmone-statetest ~/tests/EIPTests/StateTests/stEOF
+      - run:
```
gumb0, line 484: Please change test revision for now to `update-tests` (on line 477), that's the latest version

test/eoftests/eoftests.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <evmone/eof.hpp>
+#include <nlohmann/json.hpp>
+#include <filesystem>
+
+namespace fs = std::filesystem;
+namespace json = nlohmann;
+
+namespace evmone::test
+{
+struct EOFTests
```
gumb0, line 17: This struct represents what test framework calls a test (with multiple cases inside), so better to rename
```suggestion
struct EOFValidationTest
```

gumb0: Also would be good rename files and directory to `eoftest` for consistency with `statetest`

test/eoftests/eoftests_runner.cpp
```diff
@@ -0,0 +1,61 @@
...
+void run_eof_test(std::istream& input)
+{
+    const EOFTests test = json::json::parse(input).get<EOFTests>();
+    for (const auto& [name, cases] : test.cases)
+    {
+        for (const auto& expectation : cases.expectations)
+        {
+            const EOFValidationError result = evmone::validate_eof(expectation.rev, cases.code);
+            const bool b_result = (result == EOFValidationError::success);
+            EXPECT_EQ(b_result, expectation.result) << expectation.rev << " " << hex(cases.code);
```
gumb0, line 24: Also useful to output case name, since you happen to have it
```suggestion
            EXPECT_EQ(b_result, expectation.result) << name << " " << expectation.rev << " " << hex(cases.code);
```

gumb0 (changes requested):

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,83 @@
...
+    {
+        o.cases.insert({name, test.get<EOFValidationTest::Case>()});
+    }
+}
+}  // namespace
+
+namespace evmone::test
+{
+
+namespace json = nlohmann;
```
gumb0, line 67: Remove this, you have it at the top of the file.

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,83 @@
...
+        std::string name;
+        evmc::bytes code;
+        std::vector<Expectation> expectations;
+    };
+    std::unordered_map<std::string, Case> cases;
+};
+
+void from_json(const json::json& j, EOFValidationTest::Case& o)
+{
+    std::optional<evmc::bytes> hex_str{evmc::from_hex(j.at("code").get<std::string>())};
```
gumb0, line 35:
```suggestion
    const std::optional<evmc::bytes> hex_str{evmc::from_hex(j.at("code").get<std::string>())};
```

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,83 @@
...
+void from_json(const json::json& j, EOFValidationTest& o)
+{
+    if (!j.is_object() || j.empty())
+        throw std::invalid_argument{"JSON test must be an object with single key of the test name"};
+
+    const auto& j_t = *j.begin();  // Content is in a dict with the test name.
+
+    for (const auto& [name, test] : j_t.at("vectors").items())
+    {
+        o.cases.insert({name, test.get<EOFValidationTest::Case>()});
```
gumb0, line 59:
```suggestion
        o.cases.emplace(name, test.get<EOFValidationTest::Case>());
```

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+struct EOFValidationTest
+{
+    struct Case
+    {
+        struct Expectation
+        {
+            evmc_revision rev;
+            bool result;
+            std::string exception;
```
gumb0, line 24: You don't fill it when parsing, so it can be removed.

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,83 @@
...
+        throw std::invalid_argument{"invalid hex string"};
+    const evmc::bytes code{hex_str.value()};
+    o.code = code;
+
+    for (const auto& [rev, result] : j.at("results").items())
+    {
+        EOFValidationTest::Case::Expectation expectation{};
+        expectation.rev = to_rev(rev);
+        expectation.result = result.at("result").get<bool>();
+        o.expectations.push_back(expectation);
```
gumb0, line 46: After you remove `Expectation::exception` this probably can be just
```suggestion
        o.expectations.emplace_back(to_rev(rev), result.at("result").get<bool>());
```

gumb0 (changes requested):
Some minor nitpicks left.

test/eoftest/eoftest_runner.cpp
gzanitti: Oh, sorry, I thought the exception field was going to be used but you still had to define how

gzanitti:
@gumb0 some of the tests fail because of the emplace_back (probably a version problem), should I roll back the change?

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+{
+    const std::optional<evmc::bytes> hex_str{evmc::from_hex(j.at("code").get<std::string>())};
+    if (!hex_str)
+        throw std::invalid_argument{"invalid hex string"};
+    const evmc::bytes& code{hex_str.value()};
+    o.code = code;
+
+    for (const auto& [rev, result] : j.at("results").items())
+    {
+        o.expectations.emplace_back(to_rev(rev), result.at("result").get<bool>());
```
gumb0, line 42: Looks like not all compilers are happy with this, this should help I think
```suggestion
        o.expectations.emplace_back({to_rev(rev), result.at("result").get<bool>()});
```

gzanitti: That doesn't seem to work either, so I decided to follow this code: https://github.com/ethereum/evmone/blob/095f83a3931c012af43e6580d188d4bb88af5d45/test/statetest/statetest_loader.cpp#L367

Looks like someone was faced with the same issue haha.

gumb0:
Looks good to me now, please squash once again

gumb0:
I rebased and changed commit message.

gumb0 (approved):
One failing test should be fixed when https://github.com/ethereum/evmone/pull/619 is merged.

test/eoftest/eoftest.cpp
```diff
@@ -0,0 +1,87 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2022 The evmone Authors.
```
chfast, line 2:
```suggestion
// Copyright 2023 The evmone Authors.
```

test/eoftest/eoftest.hpp
```diff
@@ -0,0 +1,13 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2022 The evmone Authors.
```
chfast, line 2:
```suggestion
// Copyright 2023 The evmone Authors.
```

test/utils/utils.hpp
```diff
@@ -27,3 +28,39 @@ inline bytes operator""_hex(const char* s, size_t size)
 {
     return from_spaced_hex({s, size}).value();
 }
+
+/// Translates tests fork name to EVM revision
+inline evmc_revision to_rev(std::string_view s)
```
chfast, line 33: i think this still should be inside `evmone::test` namespace (although other things in this files are not).
Moreover, we should move the implementation to `utils.cpp` because we often need to drop new entry here for testing. This will save us rebuilding most of the files.
Finally, it would be nice to split this change into separate commit. I can do this in the end.

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+    const auto& j_t = *j.begin();  // Content is in a dict with the test name.
+
+    for (const auto& [name, test] : j_t.at("vectors").items())
+    {
+        o.cases.emplace(name, test.get<EOFValidationTest::Case>());
+    }
+}
+}  // namespace
+
+namespace evmone::test
```
chfast, line 60: Wrap whole file with `evmone::test` namespace. I.e. move the anonymous namespace into the `evmone::test`.

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+        std::vector<Expectation> expectations;
+    };
+    std::unordered_map<std::string, Case> cases;
+};
+
+void from_json(const json::json& j, EOFValidationTest::Case& o)
+{
+    const std::optional<evmc::bytes> hex_str{evmc::from_hex(j.at("code").get<std::string>())};
+    if (!hex_str)
+        throw std::invalid_argument{"invalid hex string"};
```
chfast, line 36:
```suggestion
        throw std::invalid_argument{"code is invalid hex string"};
```

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+        std::string name;
+        evmc::bytes code;
+        std::vector<Expectation> expectations;
+    };
+    std::unordered_map<std::string, Case> cases;
+};
+
+void from_json(const json::json& j, EOFValidationTest::Case& o)
+{
+    const std::optional<evmc::bytes> hex_str{evmc::from_hex(j.at("code").get<std::string>())};
```
chfast, line 34:
```suggestion
    const auto opt_code = evmc::from_hex(j.at("code").get<std::string>());
```
We rather use auto for returned types and classic assignment syntax. Also the value is not in hex any more so call it `opt_code`.

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+    if (!hex_str)
+        throw std::invalid_argument{"invalid hex string"};
+    const evmc::bytes& code{hex_str.value()};
```
chfast, line 37:
```suggestion
    o.code = *opt_code;
```
1. Taking returned value by reference is always risky. This sometimes works because of the lifetime extension but there are some gotchas leading to dangling pointers. So it should be `const evmc::bytes code{hex_str.value()};`.
2. At this point we know that the optional has a value so we can use check-less `*opt_code` operator.
3. Just avoid going through a temporary `code` variable, assign directly to `o.code`.

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+}  // namespace
+
+namespace evmone::test
+{
+
+void run_eof_test(std::istream& input)
+{
+    const EOFValidationTest test = json::json::parse(input).get<EOFValidationTest>();
```
chfast, line 65:
```suggestion
    const auto test = json::json::parse(input).get<EOFValidationTest>();
```

test/eoftest/eoftest_runner.cpp
```diff
@@ -0,0 +1,78 @@
...
+    for (const auto& [name, cases] : test.cases)
+    {
+        for (const auto& expectation : cases.expectations)
+        {
+            const EOFValidationError result = evmone::validate_eof(expectation.rev, cases.code);
```
chfast, line 70:
```suggestion
            const auto result = evmone::validate_eof(expectation.rev, cases.code);
```

chfast (changes requested):

test/utils/utils.hpp
gzanitti: Hi @chfast, thanks for all your comments and explanations, they were very helpful. I tried to implement the separation you mention in this comment. However, I always get an 'undefined reference to `evmone::test::to_rev(...)', so clearly I'm missing something and I don't see what it could be.

Sorry, my knowledge of C++ is not something that stands out haha.

gumb0: You're not compiling `utils.cpp` currently, add it to `test/utils/CMakeLists.txt`
(also you'll probably need to add library dependency `eoftest` - > `testutils`)

gzanitti: > You're not compiling `utils.cpp` currently, add it to `test/utils/CMakeLists.txt`
> (also you'll probably need to add library dependency `eoftest` - > `testutils`)

Oh, right, the CMakeList! I always forget them 😞. Thanks!

test/utils/utils.cpp
```diff
@@ -0,0 +1,46 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "utils.hpp"
+
+namespace evmone::test
+{
+
+/// Translates tests fork name to EVM revision
```
chfast, line 10: This doc comment should be placed in the function declaration in the header file.

test/eoftest/eoftest.hpp
```diff
@@ -0,0 +1,13 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <iostream>
```
chfast, line 6:
```suggestion
#include <iosfwd>
```

chfast (changes requested):

gumb0 (approved):

PR #682, opened 2023-08-09
Blockchain tests fixing: Introduce difficulty calculation in t8n

rodiazet:
- Difficulty calculation added
- Use it in t8n interface implementation


test/t8n/t8n.cpp
```diff
@@ -19,6 +19,64 @@ using namespace evmone;
...
+
+    const auto H_i_prime = kappa >= block_num ? 0 : block_num - kappa;
+
+    const auto p = (H_i_prime / 100000) - 2;
+    assert(p < 63);
+
+    const auto epsilon = p < 0 ? 0 : int64_t{1} << p;
+
+    static const hash256 empty_uncle_hash =
+        0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347_bytes32;
```
chfast, line 68: I think this is a hash of empty RLP list. See `EmptyLogsHash`.

rodiazet: Created `EmptyListHash` in `state.hpp`. Later it can be used instead `EmptyLogsHash`

test/state/state.cpp
```diff
@@ -136,11 +136,26 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
...
+    std::optional<uint64_t> block_reward, std::span<Withdrawal> withdrawals,
+    std::span<Ommer> ommers)
 {
     // TODO: The block reward can be represented as a withdrawal.
     if (block_reward.has_value())
+    {
         state.touch(coinbase).balance += *block_reward;
+        if (!ommers.empty())
+        {
+            state.touch(coinbase).balance += *block_reward * ommers.size() / 32;
```
gumb0, line 148: I think no need to `touch()` because it's always touched above, can be just `get()`.

test/state/state.cpp
```diff
@@ -136,11 +136,26 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
...
+
+            for (const auto& ommer : ommers)
+            {
+                uint256 block_reward_256 = *block_reward;
+                block_reward_256 = block_reward_256 * (8 - ommer.delta) / 8;
```
gumb0, line 153: Just initialize once to make it const
```suggestion
                const auto block_reward_256 = u256{*block_reward} * (8 - ommer.delta) / 8;
```

test/statetest/statetest_loader.cpp
```diff
@@ -181,10 +195,25 @@ state::BlockInfo from_json<state::BlockInfo>(const json::json& j)
         }
     }
 
+    std::vector<state::Ommer> ommers;
+    if (const auto ommers_it = j.find("ommers"); ommers_it != j.end())
+    {
+        for (const auto& ommer : *ommers_it)
+        {
+            ommers.push_back(
+                {from_json<evmc::address>(ommer.at("address")), ommer.at("delta").get<uint8_t>()});
```
gumb0, line 204: This will also check if the test has valid number in `delta`
```suggestion
                {from_json<evmc::address>(ommer.at("address")), from_json<uint8_t>(ommer.at("delta"))});
```

test/t8n/t8n.cpp
chfast: We should get #699 first.

test/state/state.hpp
```diff
@@ -82,11 +90,16 @@ struct BlockInfo
 {
     int64_t number = 0;
     int64_t timestamp = 0;
+    int64_t parent_timestamp = 0;
     int64_t gas_limit = 0;
     address coinbase;
+    int64_t current_difficulty = 0;
+    int64_t parent_difficulty = 0;
+    hash256 parent_uncle_hash = {};
```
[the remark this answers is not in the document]
chfast: It has to be like this because of some compiler warning.s

test/state/ethash_difficulty.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    // TODO: Implement for older revisions
+    if (rev < EVMC_BYZANTIUM)
+        return 0x020000;
+
+    static constexpr auto min_difficulty = int64_t{1} << 17;
+
+    const auto kappa = get_bomb_delay(rev);
+
+    const auto H_i_prime = kappa >= block_number ? 0 : block_number - kappa;
```
chfast, line 48:
```suggestion
    const auto H_i_prime = std::max(int64_t{0}, block_number - kappa);
```

test/state/state.cpp
chfast: Outdated.

test/statetest/statetest_loader.cpp
chfast: The `from_json<uint8_t>()` is not implemented, but by convention it would expect values in hex string. I also changed it to `uint32_t` but I'm thinking about using `int` because all difficulty calculations are done with signed integers.

test/state/state.hpp
```diff
@@ -84,11 +90,16 @@ struct BlockInfo
 {
     int64_t number = 0;
     int64_t timestamp = 0;
+    int64_t parent_timestamp = 0;
     int64_t gas_limit = 0;
     address coinbase;
+    int64_t difficulty = 0;
+    int64_t parent_difficulty = 0;
+    hash256 parent_uncle_hash = {};
```
chfast, line 98:
```suggestion
    hash256 parent_ommers_hash = {};
```

test/t8n/t8n.cpp
```diff
@@ -84,8 +85,36 @@ int main(int argc, const char* argv[])
...
-        j_result["currentDifficulty"] = "0x20000";
+
+        // Difficulty was received from upstream. No need to calc
+        // TODO: Check is it's needed by the blockchain test. If not remove if statement true branch
+        if (block.difficulty != 0)
+            j_result["currentDifficulty"] = hex0x(block.difficulty);
+        else
+        {
+            if (rev >= EVMC_SHANGHAI)
+                j_result["currentDifficulty"] = nullptr;
```
chfast, line 96: Does this mean to delete the key in JSON?

test/unittests/state_transition_block_test.cpp
```diff
@@ -34,3 +34,19 @@ TEST_F(state_transition, known_block_hash)
...
         0x0000000000000000000000000000000000000000000000000000000000000111_bytes32;
 }
+
+TEST_F(state_transition, block_apply_ommers_reward)
+{
+    static constexpr auto o1 = Ommer{0x0eeee1_address, 1};
+    static constexpr auto o2 = Ommer{0x0eeee2_address, 3};
+
+    rev = EVMC_LONDON;
+    block_reward = 9'700'000;
```
chfast, line 44: Make it 5 ETH (`5*10**16`) to check for potential overflows in calculations.

test/statetest/statetest_loader.cpp
chfast: Keeping it as `uint32_t` because the reward is unsigned also.

test/state/state.cpp
```diff
@@ -136,11 +136,24 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
 }
 
 void finalize(State& state, evmc_revision rev, const address& coinbase,
-    std::optional<uint64_t> block_reward, std::span<Withdrawal> withdrawals)
+    std::optional<uint64_t> block_reward, std::span<Withdrawal> withdrawals,
```
gumb0, line 139: Nit: maybe `withdrawals` the last parameter, because it's the newest invention and they are applied last in the code.

test/state/state.cpp
```diff
@@ -136,11 +136,24 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
...
+    {
+        const auto reward = *block_reward;
+        assert(reward % 32 == 0);  // Assume block reward is divisible by 32.
+        const auto reward_by_32 = reward / 32;
+        const auto reward_by_8 = reward / 8;
+
+        state.touch(coinbase).balance += reward + reward_by_32 * ommers.size();
+        for (const auto& ommer : ommers)
+        {
+            assert(ommer.delta > 0 && ommer.delta <= 8);
```
gumb0, line 153: maybe
```suggestion
            assert(ommer.delta > 0 && ommer.delta < 8);
```
becaue if it's equal 8 and reward is 0, it would be in correct to touch

test/state/state.cpp
chfast: Done.

chfast (approved):

PR #684, opened 2023-08-09
Blockchain tests support: Do not expect tx sender always exist

rodiazet:
Properly handle the case when the account of transaction sender doesn't
exist. In pre-Byzantium revisions such transaction could be valid with
gas price 0. In any case the transaction validation should not crash.

Fixes https://github.com/ethereum/evmone/issues/692.

### DIFF
--- a/test/unittests/state_transition.cpp
+++ b/test/unittests/state_transition.cpp
@@ -17,8 +17,27 @@ void state_transition::SetUp()
 
 void state_transition::TearDown()
 {
-    auto& state = pre;
+    auto state = pre;
     const auto res = evmone::state::transition(state, block, tx, rev, vm, block.gas_limit);
+
+    if (const auto expected_error = make_error_code(expect.tx_error))
+    {
+        ASSERT_TRUE(holds_alternative<std::error_code>(res))
+            << "tx expected to be invalid with error: " << expected_error.message();
+        const auto tx_error = std::get<std::error_code>(res);
+        EXPECT_EQ(tx_error, expected_error)
+            << tx_error.message() << " vs " << expected_error.message();
+
+        // TODO: Compare states carefully, they should be identical.
+        EXPECT_EQ(state.get_accounts().size(), pre.get_accounts().size());
+        for (const auto& [addr, acc] : state.get_accounts())
+        {
+            EXPECT_TRUE(pre.get_accounts().contains(addr)) << "unexpected account " << addr;
+        }
+
+        return;  // Do not check anything else.
+    }
+
     ASSERT_TRUE(holds_alternative<TransactionReceipt>(res))
         << std::get<std::error_code>(res).message();
     const auto& receipt = std::get<TransactionReceipt>(res);
--- a/test/unittests/state_transition.hpp
+++ b/test/unittests/state_transition.hpp
@@ -5,6 +5,7 @@
 
 #include <evmone/evmone.h>
 #include <gtest/gtest.h>
+#include <test/state/errors.hpp>
 #include <test/state/host.hpp>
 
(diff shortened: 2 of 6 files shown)


test/t8n/t8n.cpp
```diff
@@ -111,6 +111,26 @@ int main(int argc, const char* argv[])
 
                     const auto computed_tx_hash = keccak256(rlp::encode(tx));
 
+                    auto sender_acc_ptr = state.find(tx.sender);
+                    if (sender_acc_ptr == nullptr)
+                    {
+                        // It's possible to send tx from non-existent account before London fork.
+                        // TODO: This should be handled in `validate_transaction` but it requires
```
chfast, line 118: I'm not sure. We should not assume the sender account exist.

Try changing in `transition()`:
```cpp
auto& sender_acc = state.get(tx.sender);
```
to
```cpp
auto& sender_acc = state.get_or_insert(tx.sender);
```

chfast:
Added request for more tests: https://github.com/ethereum/tests/issues/1286.

test/unittests/state_transition.cpp
```diff
@@ -17,8 +17,26 @@ void state_transition::SetUp()
 
 void state_transition::TearDown()
 {
-    auto& state = pre;
+    auto state = pre;
     const auto res = evmone::state::transition(state, block, tx, rev, vm, block.gas_limit);
+
+    if (expect.tx_error != SUCCESS)
+    {
+        ASSERT_TRUE(holds_alternative<std::error_code>(res));
```
gumb0, line 25: Why is this assertion and not proper check, what if implementation mistakenly does successful transition for invalid transaction?

test/unittests/state_transition.hpp
```diff
@@ -43,6 +44,10 @@ class state_transition : public testing::Test
 
     struct Expectation
     {
+        /// The transaction is invalid because of the given error.
+        /// The rest of Expectation is ignored if the error is expected.
```
gumb0, line 48: It could be expressed with `std::variant<Expectation, ErrorCode>`

test/unittests/state_transition.cpp
chfast: There is another old check after this block which checks the other part. But probably make sense to rework this.

test/unittests/state_transition.hpp
chfast: Good idea, but `expect.post[tx.sender].nonce` stops working because `expect` is variant now and does not have `.post` any more.

test/unittests/state_transition.cpp
chfast: Hm... I'm not sure how to do this better. Any suggestions?

gumb0: Actually I confused this with normal assertion. It think this is fine then, maybe just add a better message
```cpp
ASSERT_TRUE(holds_alternative<std::error_code>(res))) << "tx expected to be invalid with error: " << expected.message();
```

test/unittests/state_transition.hpp
gumb0: Yeah the only solution that I can think of is adding `expect()` and `tx_error()` helpers that would set a variant when called, but perhaps this is too awkward.

test/unittests/state_transition.cpp
```diff
@@ -17,8 +17,26 @@ void state_transition::SetUp()
 
 void state_transition::TearDown()
 {
-    auto& state = pre;
+    auto state = pre;
```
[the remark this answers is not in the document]
chfast: To compare if the state is unchanged in case of invalid transaction.

gumb0 (approved):

chfast (approved):

PR #685, opened 2023-08-09
Blockchain tests support: Add nonce value verification

rodiazet:


chfast (commented):
All of this looks to be done in a wrong place. We should check it in `validate_transaction()` where we inspect sender's nonce already.

test/state/state.cpp
```diff
@@ -89,6 +89,12 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     if (sender_acc.nonce == Account::NonceMax)
         return make_error_code(NONCE_HAS_MAX_VALUE);
 
+    if (sender_acc.nonce < tx.nonce)
```
chfast, line 92: Do we need "to high"/"to low" variants? How about `NONCE_INVALID`?

rodiazet: We don't. But blockchain tests required to distinct between them.
`src/BlockchainTestsFiller/ValidBlocks/bcStateTests/TransactionNonceCheckFiller.json`
`src/BlockchainTestsFiller/ValidBlocks/bcStateTests/TransactionNonceCheck2Filler.json`

chfast: Ok, let it be.

chfast (changes requested):
Why are many state tests failing?

rodiazet:
> Why are many state tests failing?

Because tx which is build by `statetest_runner` has `nonce == 0` when the `sender` `nonce == 1`

test/statetest/statetest_loader.cpp
```diff
@@ -251,6 +251,9 @@ static void from_json_tx_common(const json::json& j, state::Transaction& o)
 {
     o.sender = from_json<evmc::address>(j.at("sender"));
 
+    if (const auto nonce_it = j.find("nonce"); nonce_it != j.end())
```
chfast, line 254: Isn't the nonce always expected there?

rodiazet: Our `load_minimal_test` test does not define nonce. So we can add that it's mandatory or we load it only when it's defined.

test/unittests/state_validate_transaction_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+#include <test/state/errors.hpp>
+#include <test/state/state.hpp>
+
+using namespace evmone::state;
+
+TEST(transaction, validate_transaction)
```
chfast, line 11:
```suggestion
TEST(state_tx, validate_nonce)
```
Also change file name to `state_tx_test.cpp`

test/unittests/state_validate_transaction_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+    Transaction tx{
+        .data = {},
+        .gas_limit = 60000,
+        .max_gas_price = bi.base_fee,
+        .nonce = 1,
+    };
+
+    {
+        const auto res = validate_transaction(acc, bi, tx, EVMC_BERLIN);
+        ASSERT_TRUE(holds_alternative<int64_t>(res)) << std::get<std::error_code>(res).message();
```
chfast, line 24:
```suggestion
        EXPECT_FALSE(holds_alternative<std::error_code>(res)) << std::get<std::error_code>(res).message();
```

test/unittests/state_validate_transaction_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    tx.nonce = 0;
+
+    {
+        const auto res = validate_transaction(acc, bi, tx, EVMC_BERLIN);
+        ASSERT_TRUE(holds_alternative<std::error_code>(res))
+            << make_error_code(ErrorCode::NONCE_TOO_LOW).message();
```
chfast, line 32: Why do you need this additional message?

chfast (changes requested):

test/unittests/state_validate_transaction_test.cpp
rodiazet: No we don't. I will add also checking the error code returned.

test/unittests/state_tx_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+using namespace evmone::state;
+
+TEST(state_tx, validate_nonce)
+{
+    using evmc::operator""_address;
+
+    const BlockInfo bi{.gas_limit = 0x989680,
+        .coinbase = 0x01_address,
+        .prev_randao = {},
```
chfast, line 17:
```suggestion
```
I think you can skip default initialized fields.

chfast (approved):

test/unittests/state_tx_test.cpp
rodiazet: No I cannot. CI complains during build.

PR #686, opened 2023-08-15
MPT: Add support for "short" nodes

chfast:
For the branch and extended nodes in Merkle Patricia Trie, when the length of the encoded child is 31 or smaller the hashing of the child node is omitted.

### DIFF
--- a/test/state/mpt.cpp
+++ b/test/state/mpt.cpp
@@ -219,6 +219,16 @@ void MPTNode::insert(const Path& path, bytes&& value)  // NOLINT(misc-no-recursi
     }
 }
 
+/// Encodes a node and optionally hashes the encoded bytes
+/// if their length exceeds the specified threshold.
+static bytes encode_child(const MPTNode& child) noexcept  // NOLINT(misc-no-recursion)
+{
+    if (auto e = child.encode(); e.size() < 32)
+        return e;  // "short" node
+    else
+        return rlp::encode(keccak256(e));
+}
+
 bytes MPTNode::encode() const  // NOLINT(misc-no-recursion)
 {
     bytes encoded;
@@ -237,7 +247,7 @@ bytes MPTNode::encode() const  // NOLINT(misc-no-recursion)
         for (const auto& child : m_children)
         {
             if (child)
-                encoded += rlp::encode(keccak256(child->encode()));
+                encoded += encode_child(*child);
             else
                 encoded += empty;
         }
@@ -246,8 +256,7 @@ bytes MPTNode::encode() const  // NOLINT(misc-no-recursion)
     }
     case Kind::ext:
     {
-        encoded =
-            rlp::encode(m_path.encode(true)) + rlp::encode(keccak256(m_children[0]->encode()));
+        encoded = rlp::encode(m_path.encode(true)) + encode_child(*m_children[0]);
         break;
     }
     }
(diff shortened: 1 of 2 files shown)


rodiazet (approved):

test/state/mpt.cpp
```diff
@@ -237,7 +244,7 @@ bytes MPTNode::encode() const  // NOLINT(misc-no-recursion)
         for (const auto& child : m_children)
         {
             if (child)
-                encoded += rlp::encode(keccak256(child->encode()));
+                encoded += encode_child(*child);
```
gumb0, line 247: Noy sure why lambda is needed, just to avoid +1 nesting level?

test/state/mpt.cpp
```diff
@@ -219,46 +219,46 @@ void MPTNode::insert(const Path& path, bytes&& value)  // NOLINT(misc-no-recursi
     }
 }
 
-hash256 MPTNode::hash() const  // NOLINT(misc-no-recursion)
+bytes MPTNode::encode() const  // NOLINT(misc-no-recursion)
 {
+    static constexpr auto encode_child = [](const MPTNode& child) {  // NOLINT(misc-no-recursion)
+        if (auto e = child.encode(); e.size() < 32)
```
gumb0, line 225:
```suggestion
        if (const auto e = child.encode(); e.size() < 32)
```

test/unittests/state_mpt_test.cpp
```diff
@@ -164,6 +164,9 @@ TEST(state_mpt, trie_topologies)
         const char* hash_hex;
     };
 
+    // The test cases are cross-checked with go-ethereum implementation.
+    // https://github.com/ethereum/go-ethereum/blob/master/trie/stacktrie_test.go#L35
```
gumb0, line 168: permalink
```suggestion
    // https://github.com/ethereum/go-ethereum/blob/7dea9c10cdb42e8c9f71b8b324cbe9222ab105cf/trie/stacktrie_test.go#L35
```

test/state/mpt.cpp
gumb0: Okay if you wish, but I kinda jump up and down the function to understand it :upside_down_face:

test/state/mpt.cpp
chfast: This prevents move at `return e;`

test/state/mpt.cpp
chfast: I needed a helper function because this is used for `branch` and `ext` nodes. It can be a regular static function. I was just curious how lambda works in this case.

test/unittests/state_mpt_test.cpp
chfast: It could be, but I'm not sure. Maybe some other test are going to be added there later. On the other hand fixed commit hash will be information when these were in sync...

test/state/mpt.cpp
gumb0: That's pretty weird, I think in a normal function copy/move would be elided even if returned variable is const. What if you define lamda's return type?

test/state/mpt.cpp
gumb0: I ah see, didn't noticed it's reused in ext.

test/state/mpt.cpp
chfast: Seems being under the `if` statement make a difference. https://godbolt.org/z/rq7W7fsxM

PR #688, opened 2023-08-18
precompiles: Implement secp256k1 ECDSA recovery using EVMMAX

chfast:
Implement Elliptic Curve Digital Signature Algorithm (ECDSA)
Public Key Recovery algorithm for secp256k1 curve
using [EVMMAX](https://eips.ethereum.org/EIPS/eip-6690) primitives.

This can be used to provide ecrecovery EVM precompile
but also to verify signatures in Ethereum transactions.

This work has been done at ETHPrague Hackathon
https://devfolio.co/projects/evmmax-ecrecovery-bd49


lib/evmone_precompiles/ecc.hpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+template <typename IntT, int A = 0>
+ProjPoint<IntT> add(const evmmax::ModArith<IntT>& s, const ProjPoint<IntT>& p,
+    const ProjPoint<IntT>& q, const IntT& b3) noexcept
+{
+    static_assert(A == 0, "point addition procedure is simplified for a = 0");
+
+    if (p.is_inf())
+        return q;
```
rodiazet, line 73: Shouldn't we return `infinity` here and for other cases when input is at infinity?

lib/evmone_precompiles/secp256k1.hpp
```diff
@@ -28,4 +32,6 @@ uint256 field_inv(const ModArith<uint256>& m, const uint256& x) noexcept;
 /// @return Square root of x if it exists, std::nullopt otherwise.
 std::optional<uint256> field_sqrt(const ModArith<uint256>& m, const uint256& x) noexcept;
 
+uint256 scalar_inv(const ModArith<uint256>& m, const uint256& x) noexcept;
```
rodiazet, line 40: Doc comment missing

lib/evmone_precompiles/secp256k1.hpp
```diff
@@ -42,7 +43,16 @@ uint256 scalar_inv(const ModArith<uint256>& m, const uint256& x) noexcept;
 std::optional<uint256> calculate_y(
     const ModArith<uint256>& m, const uint256& x, bool y_parity) noexcept;
 
+Point add(const Point& p, const Point& q) noexcept;
```
rodiazet, line 46: Doc comment missing. Should we move it to separated commit?

lib/evmone_precompiles/ecc.hpp
chfast: The "infinity" work as identity element (0) :shrug:

rodiazet: Make sense then 👍

lib/evmone_precompiles/secp256k1.cpp
```diff
@@ -0,0 +1,764 @@
...
+    for (int i = 0; i < 2; ++i)
+        z = m.mul(z, z);
+
+    if (m.mul(z, z) != x)
+        return std::nullopt;  // Computed value is not the square root.
+
+    return z;
+}
+
+uint256 scalar_inv(const ModArith<uint256>& m, const uint256& x) noexcept
```
axic, line 433: Can we add a comment (or extend the one below) to include a reference to the `expmod.tmpl` file?

chfast: Yes. The commit messages have all the information how to generate this code, but I will add this information here too.

chfast: Added detailed instruction how this was generated.

lib/evmone_precompiles/ecc.hpp
```diff
@@ -0,0 +1,229 @@
...
+    if (q.is_inf())
+        return p;
+
+    // https://eprint.iacr.org/2015/1060 algorithm 1.
+    // Simplified with a == 0
```
pdobacz, line 78: Why not use algorithms 7 & 9 catering for the a==0 case instead? They seem to be marginally cheaper than our versions (Alg9 quite a bit actually) and also we adhere more closely to the paper we're based on.

chfast: Looks like my comments were not sent. Anyway, I was just clumsy and didn't notice there were variants with `a == 0` in the paper.

pdobacz: I can take a stab at this and use alg 7 & 9 here as a warm-up PR, WDYT?

lib/evmone_precompiles/secp256k1.hpp
```diff
@@ -19,4 +20,12 @@ inline constexpr auto FieldPrime =
 /// where P is ::FieldPrime.
 uint256 field_inv(const ModArith<uint256>& m, const uint256& x) noexcept;
 
+/// Square root for secp256k1 prime field.
+///
+/// Computes √x mod P modular inversion by computing modular exponentiation x^(P/4),
```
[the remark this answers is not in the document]
chfast: I guess. This is little Fermat?

test/unittests/evmmax_secp256k1_test.cpp
```diff
@@ -150,3 +150,220 @@ TEST(secp256k1, point_to_address)
     // https://etherscan.io/address/0x3f17f1962b36e491b30a40b2405849e597ba5fb5
     EXPECT_EQ(to_address(Point{}), 0x3f17f1962B36e491b30A40b2405849e597Ba5FB5_address);
 }
+
+TEST(evmmax, secp256k1_calculate_u1)
+{
+    // u1 = -zr^(-1)
+    const auto z = 0x31d6fb860f6d12cee6e5b640646089bd5883d586e43de3dedc75695c11ac2da9_u256;
```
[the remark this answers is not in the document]
chfast: I think we just dumped these intermediate value from some other implementation or used PyEC to compute this example. At this point we may just remove these tests.

test/unittests/evmmax_secp256k1_test.cpp
```diff
@@ -0,0 +1,369 @@
...
+    }
+}
+
+TEST(secp256k1, scalar_inv)
+{
+    const evmmax::ModArith n{Order};
+
+    for (const auto& t : {
+             1_u256,
+             // FIXME: 0x6e140df17432311190232a91a38daed3ee9ed7f038645dd0278da7ca6e497de_u256,
```
chfast, line 74: This test case is broken so we need to find out why.

lib/evmone_precompiles/secp256k1.hpp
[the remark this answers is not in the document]
pdobacz: yup!

test/unittests/evmmax_secp256k1_test.cpp
pdobacz: It uses `field_inv`, not `scalar_inv`. Fixed now and test passes.

test/unittests/evmmax_secp256k1_test.cpp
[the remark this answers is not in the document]
pdobacz: ok, will remove the intermediate result tests, but we still could document the values from the `add` and `mul` tests (which are more like end-result tests). If you don't have it handy, I can try to recreate, unless you think this is overdoing.

test/state/precompiles.cpp
```diff
@@ -142,6 +143,37 @@ PrecompileAnalysis expmod_analyze(bytes_view input, evmc_revision rev) noexcept
         static_cast<size_t>(mod_len)};
 }
 
+ExecutionResult ecrecover_execute(const uint8_t* input, size_t input_size, uint8_t* output,
+    [[maybe_unused]] size_t output_size) noexcept
+{
+    assert(output_size == 32);
```
pdobacz, line 149: I think my comment got lost somewhere, re-adding:

@chfast this assertion fails, because the caller sends `4096` as `output_size`. I'm new to this so I'm not sure which part gets it wrong.

Shall we do `assert(output_size >= 32)` and `output_size = 32` instead?

chfast: Yes, you can change it to `>= 32`. We changed logic a bit recently.

pdobacz: pls double check if this makes sense, I adhered to `identity_execute` [a82f0cb](https://github.com/ethereum/evmone/pull/688/commits/a82f0cbf90146a4bf608eecb577b3cfc0ac91a09)

pdobacz (approved):

PR #690, opened 2023-08-22
Blockchain tests support: Withdrawals hashing

rodiazet:
- Support withdrawals root hash calculation
- Add `ArrowGlacier` fork name mapping to `evmc_revision` (cast to London fork).


test/unittests/state_mpt_hash_test.cpp
```diff
@@ -10,6 +10,7 @@
 #include <test/state/mpt_hash.hpp>
 #include <test/state/rlp.hpp>
 #include <test/state/state.hpp>
+#include <test/statetest/statetest.hpp>
```
chfast, line 13: You should not mix JSON with these tests.

chfast (approved):

PR #696, opened 2023-08-24
t8n: validate deployed code

hugo-dc:
When running State Tests, there is a validation that checks any EOF code in pre-state is valid. If it is not valid retesteth shows the validation error.  However when filling state tests, such validation is not performed, leaving invalid code in pre-state and allowing to be executed.


chfast (commented):
I think this is fine for t8n but the other mentioned check on loading breaks fuzz testing in goevmlab. What do you think about removing it? Or move to t8n?

rodiazet (approved):

rodiazet:
Fix linter please

hugo-dc:
> I think this is fine for t8n but the other mentioned check on loading breaks fuzz testing in goevmlab. What do you think about removing it? Or move to t8n?

I think it is enough to have this validation only when filling the tests (t8n tool), this way it prevents creating tests with invalid EOF in pre-state.

If you agree I can remove this validation, and only let it in t8n tool.

chfast:
> > I think this is fine for t8n but the other mentioned check on loading breaks fuzz testing in goevmlab. What do you think about removing it? Or move to t8n?
> 
> I think it is enough to have this validation only when filling the tests (t8n tool), this way it prevents creating tests with invalid EOF in pre-state.
> 
> If you agree I can remove this validation, and only let it in t8n tool.

Let's do this. I already pushed the commit we had for goevmlab that removes this check.

PR #698, opened 2023-08-30
state: Implement fake `BLOCKHASH` resolution for testing

chfast:
If explict block hashes are not provided, return predefined block hashes from `Host::get_block_hash()` as `keccak256(str(block_number))`. This is badly documented convention used in state tests.

Note: the readability of block hash values from `BLOCKHASH` instruction still applies. I.e. you will get 0 if you ask for a block from the future or more than 256 blocks older than the current one.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -326,7 +326,10 @@ bytes32 Host::get_block_hash(int64_t block_number) const noexcept
         it != m_block.known_block_hashes.end())
         return it->second;
 
-    return {};
+    // Convention for testing: if the block hash in unknown return the predefined "fake" value.
+    // https://github.com/ethereum/go-ethereum/blob/v1.12.2/tests/state_test_util.go#L432
+    const auto s = std::to_string(block_number);
+    return keccak256({reinterpret_cast<const uint8_t*>(s.data()), s.size()});
 }
 
 void Host::emit_log(const address& addr, const uint8_t* data, size_t data_size,
--- a/test/unittests/evm_state_test.cpp
+++ b/test/unittests/evm_state_test.cpp
@@ -432,7 +432,7 @@ TEST_P(evm, blockhash)
     host.block_hash.bytes[13] = 0x13;
 
     host.tx_context.block_number = 0;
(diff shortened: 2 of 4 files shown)


winsvega:
Yes, since we decided that makes no sense. There should be no state tests using blockhash instruction as it is a blockchain logic. Otherwise it just tests a test logic.

holiman:
I can't meaningfully review this, but it sounds good. A better description is here: https://github.com/NethermindEth/nethermind/issues/4967#issue-1478779347

> if we execute on BLOCKNUMBER 1 , in which case 0 is the only one "eligible" for blockhash. If we're on a later number, there are other blocks which are 'eligible'.

goevmlab only generates tests with blocknumber `1`, but it is possible for a statetest to specify a later number. So if we execute on block `1000`, and `BLOCKHASH(1)` is requested, then it should, of course, put zero on the stack, and not "look up (== keccak)" that number.

chfast:
> So if we execute on block `1000`, and `BLOCKHASH(1)` is requested, then it should, of course, put zero on the stack, and not "look up (== keccak)" that number.

Yes, this is handled already closer to the instruction implementation (i.e. EVM will not ask anything the Host in this situation).

I noted this is not needed for regular state tests because they have a discipline to avoid `BLOCKHASH` entirely.

However, for goevmlab this is _required_. But as mentioned goevmlab generated tests are always running at block 1 therefore the Host will only be asked for `BLOCKHASH(0)`. So my question to @holiman is mostly: should be keep the generic formula for the "fake" block hashes (as currently in this PR) or maybe we should limit ourselves only to

```cpp
if (block_number == 0)
    return 0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d_bytes32;
```

chfast:
I think this is the geth implementation so I will link to it. https://github.com/ethereum/go-ethereum/blob/v1.12.2/tests/state_test_util.go#L432

What is the preferred naming for this?
- fake
- cheat
- test

gumb0 (approved):

PR #699, opened 2023-08-30
test: Add test fixture for statetest tracing

chfast:
- Add `state::EmptyListHash` constant (also needed in #682).
- Make `run_state_test()` function available to unit tests.
- Introduce the unit test fixture for inspecting state test based traces.

### DIFF
--- a/test/statetest/CMakeLists.txt
+++ b/test/statetest/CMakeLists.txt
@@ -15,6 +15,7 @@ target_sources(
     statetest.hpp
     statetest_loader.cpp
     statetest_logs_hash.cpp
+    statetest_runner.cpp
 )
 
 add_executable(evmone-statetest)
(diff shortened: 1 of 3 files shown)


gumb0:
Hm the entire idea of this seems to me closer to state transition tests using `state_transition` fixture, rather than state tests that load `StateTransitionTest` struct from json tests. It maybe could have been an option of or a descendant of `state_transition`, did you consider it?

chfast:
> Hm the entire idea of this seems to me closer to state transition tests using `state_transition` fixture, rather than state tests that load `StateTransitionTest` struct from json tests. It maybe could have been an option of or a descendant of `state_transition`, did you consider it?

Yes, but close to the point this was finished. I decided to leave it like this because it had more unit-test like separation. But I will try the other way now because setting up `StateTransitionTest` feels weird.

gumb0:
It would help with debugging state transition tests, too.

chfast:
> It would help with debugging state transition tests, too.

Good point. Done in #704.

chfast:
Replaced by #704.

PR #700, opened 2023-09-02
Build and test riscv32

chfast:
This uses a pre-built RISC-V toolchain to build and test evmone.

This compiles evmone and dependencies to `rv32imafdc` with `ilp32d` ABI. I.e. this target has hardware support for floating point.

Our preference would be to compile for `rv32imac` with `ilp32` ABI. However, there are some obstacles to get there:

1. The pre-built toolchain is built for `ilp32d`. We would need to build the toolchain ourselves.
2. We can hack it by setting `-march=rv32imac -mabi=ilp32d`. The compiler is not very happy about using architecture with incompatible ABI but it will proceed anyway. Unfortunately, the dependencies have some floating-point code so we also need to provide soft-float library. I was not able to make it build so far.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -86,6 +86,14 @@ commands:
           command: |
             curl -L https://github.com/Kitware/CMake/releases/download/v<<parameters.version>>/cmake-<<parameters.version>>-linux-x86_64.tar.gz | sudo tar -xz --strip=1
 
+  install_riscv_toolchain:
+    steps:
+      - run:
+          name: "Install RISC-V Toolchain"
+          working_directory: /usr/local
+          command: |
+            curl -L https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/2023.07.07/riscv32-glibc-ubuntu-22.04-llvm-nightly-2023.07.07-nightly.tar.gz | sudo tar -xz
+
   checkout_submodules:
     steps:
       - run:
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -655,50 +663,20 @@ jobs:
...
-      - gcc-latest-memcheck
-      - clang-latest-sanitizers
-      - clang-latest-coverage
-      - macos-asan
-      # TODO: Re-enable on next Xcode upgrade
-      # - xcode-min
-      - gcc-32bit
-      - x86-64-v1
-      - fuzzing
+      - riscv32
```
chfast, line 682: TODO: Reenable other jobs.

rodiazet (approved):

cmake/toolchains/riscv32.cmake
```diff
@@ -0,0 +1,19 @@
+# evmone: Ethereum Virtual Machine
+# Copyright 2023 Pawel Bylica.
+# Licensed under the Apache License, Version 2.0. See the LICENSE file.
+
+set(RISCV /usr/local/riscv)
+
+set(CMAKE_SYSTEM_PROCESSOR riscv32)
```
axic, line 7: Can we specify the riscv32imac triple?

chfast: This value is only informational, but you can modify the compiler flags.

By default it builds as `rv32imafdc`. If we switch to `rv32imac` it will not build because the included glibc is not compatible (the ABI for floating-point values is different). We can configure and build the toolchain ourselves but seems to be a lot of work for the start.

PR #701, opened 2023-09-04
state: Delete empty accounts after each tx

rodiazet:
Delete empty accounts after each transaction
and after applying withdrawals and block rewards.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -135,6 +135,18 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     return execution_gas_limit;
 }
 
+namespace
+{
+/// Deletes "touched" (marked as erasable) empty accounts in the state.
+void delete_empty_accounts(State& state)
+{
+    std::erase_if(state.get_accounts(), [](const std::pair<const address, Account>& p) noexcept {
+        const auto& acc = p.second;
+        return acc.erasable && acc.is_empty();
+    });
+}
+}  // namespace
+
 void finalize(State& state, evmc_revision rev, const address& coinbase,
     std::optional<uint64_t> block_reward, std::span<Ommer> ommers,
     std::span<Withdrawal> withdrawals)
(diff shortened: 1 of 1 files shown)


test/state/state.cpp
```diff
@@ -158,14 +169,9 @@ void finalize(State& state, evmc_revision rev, const address& coinbase,
     for (const auto& withdrawal : withdrawals)
         state.touch(withdrawal.recipient).balance += withdrawal.get_amount();
 
+    // Empty accounts need to be clear because is 'block_reward' coinbase should be removed.
```
chfast, line 172:
```suggestion
    // Clear potentially empty block reward recipients. 
```

test/state/state.cpp
```diff
@@ -241,6 +247,11 @@ std::variant<TransactionReceipt, std::error_code> transition(State& state, const
     // Cannot put it into constructor call because logs are std::moved from host instance.
     receipt.logs_bloom_filter = compute_bloom_filter(receipt.logs);
 
+    // Empty accounts need to be cleared here to properly calculate post state hash which is a part
```
chfast, line 250:
```suggestion
    // Clear empty accounts after every transaction. This is strictly required until Byzantium where intermediate state root hashes are part of the transaction receipt.
    // TODO: Consider limiting this only to Spurious Dragon.
```

test/state/state.cpp
```diff
@@ -135,6 +135,17 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     return execution_gas_limit;
 }
 
+namespace
+{
+void clear_empty_accounts(State& state)
```
chfast, line 140: I think "clear" is not very good name. Let's call it "delete" or "erase". We can consult Yellow Paper.

chfast: Let's use "delete".

chfast (approved):

PR #702, opened 2023-09-06
EOF: TXCREATE instruction

gumb0:
Based on https://github.com/ethereum/evmone/pull/553

### DIFF
--- a/lib/evmone/execution_state.hpp
+++ b/lib/evmone/execution_state.hpp
@@ -147,6 +147,7 @@ public:
 
 private:
     evmc_tx_context m_tx = {};
+    std::optional<std::unordered_map<evmc::bytes32, bytes_view>> m_initcodes;
 
 public:
     /// Pointer to code analysis.
@@ -202,6 +203,26 @@ public:
         if (INTX_UNLIKELY(m_tx.block_timestamp == 0))
             m_tx = host.get_tx_context();
         return m_tx;
+    }
+
+    /// Get initcode by its hash from transaction initcodes.
+    ///
+    /// Returns empty bytes_view if no such initcode was found.
+    [[nodiscard]] bytes_view get_tx_initcode_by_hash(const evmc_bytes32& hash) noexcept
+    {
+        if (!m_initcodes.has_value())
+        {
+            m_initcodes.emplace();
+            const auto& tx_context = get_tx_context();
+            for (size_t i = 0; i < tx_context.initcodes_count; ++i)
+            {
+                const auto& initcode = tx_context.initcodes[i];
+                m_initcodes->insert({initcode.hash, {initcode.code, initcode.code_size}});
+            }
+        }
+
+        const auto it = m_initcodes->find(hash);
+        return it != m_initcodes->end() ? it->second : bytes_view{};
     }
 };
 }  // namespace evmone
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -300,19 +300,17 @@ Result create_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noex
     return {EVMC_SUCCESS, gas_left};
 }
 
-Result eofcreate(
+template <Opcode Op>
+Result create_eof_impl(
     StackTop stack, int64_t gas_left, ExecutionState& state, code_iterator& pos) noexcept
 {
+    static_assert(Op == OP_EOFCREATE || Op == OP_TXCREATE);
+
     if (state.in_static_mode())
         return {EVMC_STATIC_MODE_VIOLATION, gas_left};
 
-    const auto initcontainer_index = uint8_t{pos[1]};
-    pos += 2;
-
-    const auto& container = state.original_code;
-    const auto eof_header = read_valid_eof1_header(state.original_code);
-    const auto initcontainer = eof_header.get_container(container, initcontainer_index);
-
+    const auto initcode_hash =
+        (Op == OP_TXCREATE) ? intx::be::store<evmc::bytes32>(stack.pop()) : evmc::bytes32{};
     const auto endowment = stack.pop();
     const auto salt = stack.pop();
     const auto input_offset_u256 = stack.pop();
@@ -323,6 +321,33 @@ Result eofcreate(
 
     if (!check_memory(gas_left, state.memory, input_offset_u256, input_size_u256))
         return {EVMC_OUT_OF_GAS, gas_left};
+
+    bytes_view initcontainer;
+    if constexpr (Op == OP_EOFCREATE)
+    {
+        const auto initcontainer_index = pos[1];
+        pos += 2;
+        const auto& container = state.original_code;
+        const auto& eof_header = state.analysis.baseline->eof_header;
+        initcontainer = eof_header.get_container(container, initcontainer_index);
+    }
+    else
+    {
+        pos += 1;
+
+        initcontainer = state.get_tx_initcode_by_hash(initcode_hash);
+        // In case initcode was not found, empty bytes_view was returned.
+        // Transaction initcodes are not allowed to be empty.
+        if (initcontainer.empty())
+            return {EVMC_SUCCESS, gas_left};  // "Light" failure
+
+        // Charge for initcode validation.
+        constexpr auto initcode_word_cost_validation = 2;
+        const auto initcode_cost_validation =
+            num_words(initcontainer.size()) * initcode_word_cost_validation;
+        if ((gas_left -= initcode_cost_validation) < 0)
+            return {EVMC_OUT_OF_GAS, gas_left};
+    }
 
     // Charge for initcode hashing.
     constexpr auto initcode_word_cost_hashing = 6;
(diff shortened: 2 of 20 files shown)


gumb0:
Rebased. RETURNCONTRACT is back to having two arguments.

gumb0:
Rebased and squashed.

gumb0:
Rebased.

lib/evmone/instructions_calls.cpp
```diff
@@ -226,6 +224,43 @@ Result create3(StackTop stack, int64_t gas_left, ExecutionState& state, code_ite
...
+        initcontainer = {initcode.data, initcode.size};
+
+        // Charge for initcode validation.
+        constexpr auto initcode_word_cost_validation = 2;
+        const auto initcode_cost_validation =
+            num_words(initcontainer.size()) * initcode_word_cost_validation;
+        if ((gas_left -= initcode_cost_validation) < 0)
+            return {EVMC_OUT_OF_GAS, gas_left};
+
+        const auto error_subcont = validate_eof(state.rev, initcontainer);
```
gumb0, line 254: It might make sense to move this validation after endowment check below, but not sure.

gumb0: Note that moving it would mean in case of invalid container, we still pay for hashing that never happens.
(But it is also true for balance check even in CREATE2)

rakita: Would make sense to move gas spending for hash before validation. My thinking is if the contract can't be deployed bcs of gas I would like to know that before starting validation.

pdobacz: I don't have a strong opinion on this. Charging gas on each logical step performed seemed like commons sense, but I'm inclined to agree that the consistency reasons should take priority over this.

gumb0: I moved validation to later and made a change to the spec: https://github.com/ipsilon/eof/pull/66

lib/evmone/instructions_calls.cpp
```diff
@@ -324,6 +321,32 @@ Result eofcreate(
...
+        const auto& container = state.original_code;
+        const auto eof_header = read_valid_eof1_header(state.original_code);
+        initcontainer = eof_header.get_container(container, initcontainer_index);
+    }
+    else
+    {
+        pos += 1;
+
+        initcontainer =
+            state.get_tx_initcode_by_hash(intx::be::store<evmc::bytes32>(initcode_hash));
```
chfast, line 338: Maybe move it up to `(Op == OP_TXCREATE) ? intx::be::store<evmc::bytes32>(stack.pop()) : evmc::bytes32{}`?

lib/evmone/instructions_calls.cpp
```diff
@@ -324,6 +321,32 @@ Result eofcreate(
...
+        if (initcontainer.empty())
```
chfast, line 339: Can you actually provide an empty initcontainer in the transaction?

lib/evmone/instructions_calls.cpp
```diff
@@ -324,6 +321,32 @@ Result eofcreate(
...
+            return {EVMC_SUCCESS, gas_left};  // "Light" failure
+
+        // Charge for initcode validation.
+        constexpr auto initcode_word_cost_validation = 2;
+        const auto initcode_cost_validation =
+            num_words(initcontainer.size()) * initcode_word_cost_validation;
+        if ((gas_left -= initcode_cost_validation) < 0)
+            return {EVMC_OUT_OF_GAS, gas_left};
```
chfast, line 347: Not covered by tests?

test/state/host.cpp
```diff
@@ -410,20 +410,13 @@ evmc_tx_context Host::get_tx_context() const noexcept
...
         m_block.prev_randao,
         0x01_bytes32,  // Chain ID is expected to be 1.
-        uint256be{m_block.base_fee},
-        intx::be::store<uint256be>(m_block.blob_base_fee),
-        m_tx.blob_hashes.data(),
-        m_tx.blob_hashes.size(),
-    };
+        uint256be{m_block.base_fee}, intx::be::store<uint256be>(m_block.blob_base_fee),
+        m_tx.blob_hashes.data(), m_tx.blob_hashes.size(), m_tx_initcodes.data(),
+        m_tx_initcodes.size()};
```
chfast, line 419: This will keep the field initialization unrolled.

```suggestion
        m_tx_initcodes.size(),};
```

test/unittests/state_transition.cpp
```diff
@@ -197,6 +197,9 @@ void state_transition::export_state_test(const TransactionReceipt& receipt, cons
         jtx["maxPriorityFeePerGas"] = hex0x(tx.max_priority_gas_price);
     }
 
+    for (size_t i = 0; i < tx.initcodes.size(); ++i)
```
chfast, line 200: Maybe something like:
```cpp
auto&  jinitcodes = jtx["initcodes"] = json::json::array();
for (const auto& initcode : tx.initcodes)
    jinitcodes.push_back(hex0x(initcode));

File: test/utils/bytecode.hpp [hunk 7]
```diff
@@ -630,18 +632,30 @@ struct create_instruction
         return *this;
     }
 
+    template <Opcode k = kind>
+    typename std::enable_if<k == OP_TXCREATE, create_instruction&>::type initcode(bytecode hash)
```
chfast, line 636: C++20: `requires(k == OP_TXCREATE)`.

lib/evmone/execution_state.hpp
```diff
@@ -203,5 +204,22 @@ class ExecutionState
             m_tx = host.get_tx_context();
         return m_tx;
     }
+
+    [[nodiscard]] bytes_view get_tx_initcode_by_hash(const evmc_bytes32& hash) noexcept
+    {
+        if (INTX_UNLIKELY(!m_initcodes.has_value()))
```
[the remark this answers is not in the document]
gumb0: I'll remove it

lib/evmone/instructions_calls.cpp
```diff
@@ -324,6 +321,32 @@ Result eofcreate(
     if (!check_memory(gas_left, state.memory, input_offset_u256, input_size_u256))
         return {EVMC_OUT_OF_GAS, gas_left};
 
+    bytes_view initcontainer;
+    if constexpr (Op == OP_EOFCREATE)
+    {
+        const auto initcontainer_index = uint8_t{pos[1]};
+        pos += 2;
+        const auto& container = state.original_code;
+        const auto eof_header = read_valid_eof1_header(state.original_code);
```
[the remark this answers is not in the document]
gumb0: You're right, by this point we've already read it, and it's accessible, but only in `state.analysis.baseline`.

Maybe it's better to move it `ExecutionState`, otherwise it's like an easy way to crash Advanced.

lib/evmone/instructions_calls.cpp
gumb0: Actually you cannot, we forgot to remove this with the introduction of this rule.

test/unittests/state_transition.cpp
gumb0: I don't like this double assignment, especially with a reference, but this is similar to other parts of this code, so I'll change it.

lib/evmone/instructions_calls.cpp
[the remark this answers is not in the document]
gumb0: Another idea is to put array of `bytes_view`s to all containers into `ExecutionState`, similar to what we do now with data section.

lib/evmone/instructions_calls.cpp
```diff
@@ -300,19 +300,17 @@ Result create_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noex
...
         return {EVMC_STATIC_MODE_VIOLATION, gas_left};
 
-    const auto initcontainer_index = uint8_t{pos[1]};
-    pos += 2;
-
-    const auto& container = state.original_code;
-    const auto eof_header = read_valid_eof1_header(state.original_code);
-    const auto initcontainer = eof_header.get_container(container, initcontainer_index);
-
+    const auto initcode_hash =
```
rodiazet, line 312: It's only used for `OP_TXCREATE`. Can we move it to `else` below?

lib/evmone/instructions_calls.cpp
chfast: Please document this.

lib/evmone/instructions_calls.cpp
```diff
@@ -324,6 +322,31 @@ Result eofcreate(
...
+        constexpr auto initcode_word_cost_validation = 2;
+        const auto initcode_cost_validation =
+            num_words(initcontainer.size()) * initcode_word_cost_validation;
+        if ((gas_left -= initcode_cost_validation) < 0)
+            return {EVMC_OUT_OF_GAS, gas_left};
+    }
+
     // Charge for initcode hashing.
     constexpr auto initcode_word_cost_hashing = 6;
     const auto initcode_cost_hashing = num_words(initcontainer.size()) * initcode_word_cost_hashing;
```
rodiazet, line 352: Should we deduct charge to hashing earlier for `TXCREATE`?

lib/evmone/instructions_calls.cpp
rodiazet: I would change `get_tx_initcode_by_hash` returning `std::optional`

lib/evmone/instructions_calls.cpp
chfast: We need to pop the stack in the right order...

test/utils/bytecode.hpp
```diff
@@ -616,32 +617,46 @@ struct create_instruction
     }
 
     template <Opcode k = kind>
-    typename std::enable_if<k == OP_CREATE2 || k == OP_EOFCREATE, create_instruction&>::type salt(
-        bytecode salt)
+        requires(k == OP_CREATE2 || k == OP_EOFCREATE || k == OP_TXCREATE)
```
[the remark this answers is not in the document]
rodiazet: Resolved when https://github.com/ethereum/evmone/pull/870 merged.

lib/evmone/instructions_calls.cpp
rodiazet: Thanks. Explained also in discord.

chfast (approved):

rodiazet (approved):

PR #703, opened 2023-09-13
MTP: Extension node is a lie!

chfast:
MPT: Remove explicit extended node kind

Notice that the formally specified extended node in
the Merkle Patricia Trie always leads to a branch node.
Therefore, we can treat branch nodes as having potentially non-empty
"extended path" and remove the explicit extended node kind.
This significantly simplifies the implementation.


chfast:
After extracting refactoring to #751, this is finally ready for review.

test/state/mpt.cpp
```diff
@@ -90,39 +87,19 @@ class MPTNode
...
-    /// Optionally wraps the child node with newly created extended node in case
-    /// the provided path is not empty.
-    static std::unique_ptr<MPTNode> optional_ext(
-        const Path& path, std::unique_ptr<MPTNode> child) noexcept
-    {
-        return (!path.empty()) ? std::make_unique<MPTNode>(ext(path, std::move(child))) :
-                                 std::move(child);
-    }
-
     /// Creates a branch node out of two children and optionally extends it with an extended
```
gumb0, line 90: Comment needs to be updated.

chfast: Fixed, thanks.

test/state/mpt.cpp
```diff
@@ -143,108 +119,83 @@ void MPTNode::insert(const Path& path, bytes&& value)  // NOLINT(misc-no-recursi
...
-        auto this_leaf = leaf({this_idx + 1, m_path.end()}, std::move(m_value));
-        auto new_leaf = leaf(insert_tail, std::move(value));
-        *this =
-            ext_branch(common, *this_idx, std::move(this_leaf), *insert_idx, std::move(new_leaf));
-        break;
-    }
-
-    default:
-        assert(false);
+        const Path extended_path{m_path.begin(), this_idx_it};
```
gumb0, line 162: Can `m_kind` be `leaf` in this else branch? If not, add assertion.

chfast: This can be a leaf or an extension node (i.e. the branch node with non-empty path).

test/state/mpt.cpp
```diff
@@ -143,108 +119,83 @@ void MPTNode::insert(const Path& path, bytes&& value)  // NOLINT(misc-no-recursi
...
+        const auto this_idx = *this_idx_it;
+        m_path = Path{this_idx_it + 1, m_path.end()};  // shorten this path, invalidates this_idx_it
+        *this = branch(extended_path, this_idx, std::make_unique<MPTNode>(std::move(*this)),
+            *insert_idx_it, leaf(insert_tail, std::move(value)));
```
gumb0, line 166: What if we have a method to copy node with children, and try to simplify like:
```suggestion
        // The original node must be pushed down.
        const Path common_path{m_path.begin(), this_idx_it};
        const Path this_node_tail{this_idx_it + 1, m_path.end()};
        auto this_node = copy_node(m_kind, this_node_tail, std::move(m_children), std::move(m_value)); // this_idx_it is not invalidated
        auto new_leaf = leaf(insert_tail, std::move(value));
        
        *this = branch(common_path, *this_idx_it, std::move(this_node), *insert_idx_it, std::move(new_leaf));
```

chfast: I don't think we need special method. The move constructor can do this. I believe I tried this but the effect was more code similarly confusing. I'm still thinking about some refactoring here (e.g. removing named constructors `leaf()` and `branch()`) but wanted to minimize the diff this time.

PR #704, opened 2023-09-13
test: Capture the trace of state_transition

chfast:
Optionally (when `expect.trace = "<trace string>"`) is used capture the transaction execution trace in state_transition and compare it with the expected one.

Replaces #699.

### DIFF
--- a/test/unittests/state_transition.cpp
+++ b/test/unittests/state_transition.cpp
@@ -15,10 +15,27 @@ void state_transition::SetUp()
     expect.post[tx.sender].exists = true;
 }
 
+class TraceCapture
+{
+    std::streambuf* m_orig_rdbuf = nullptr;
+    std::ostringstream m_trace_stream;
+
+public:
+    TraceCapture() { m_orig_rdbuf = std::clog.rdbuf(m_trace_stream.rdbuf()); }
+    ~TraceCapture() { std::clog.rdbuf(m_orig_rdbuf); }
+    [[maybe_unused]] std::string get_capture() const { return m_trace_stream.str(); }
+};
+
 void state_transition::TearDown()
 {
     auto state = pre;
-    const auto res = evmone::state::transition(state, block, tx, rev, vm, block.gas_limit);
+    const auto trace = !expect.trace.empty();
+    auto& selected_vm = trace ? tracing_vm : vm;
+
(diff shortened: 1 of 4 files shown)


test/unittests/state_transition.cpp
```diff
@@ -15,10 +15,29 @@ void state_transition::SetUp()
...
 {
     auto state = pre;
-    const auto res = evmone::state::transition(state, block, tx, rev, vm, block.gas_limit);
+    const auto trace = !expect.trace.empty();
+    auto& selected_vm = trace ? tracing_vm : vm;
+
+    /// Optionally enable trace capturing but keep the RAII object in scope.
+    std::optional<const TraceCapture> trace_capture;
+    if (trace)
+        trace_capture.emplace();
```
gumb0, line 38: could be something like
```suggestion
    const std::optional<const TraceCapture> trace_capture(trace ? TraceCapture{} : std::nullopt);
```

test/unittests/state_transition_trace_test.cpp
```diff
@@ -0,0 +1,27 @@
...
+using namespace evmc::literals;
+using namespace evmone::test;
+
+TEST_F(state_transition, trace_example)
+{
+    rev = EVMC_SHANGHAI;
+    tx.to = To;
+    pre.insert(To, {.code = revert(0, 1)});
+
+    rev = EVMC_BERLIN;
```
gumb0, line 17: Why do you assign `rev` twice?

test/unittests/state_transition.cpp
chfast: `TraceCapture` is not movable because `std::ostringstream` is not (?).

test/unittests/state_transition_trace_test.cpp
chfast: Copy-paste bug.

test/unittests/state_transition.cpp
chfast: Ok, it seems it is movable. So maybe I need to enable move constructor in `TraceCapture`...

chfast: I found a way, but it looks it relies on copy elision. It is nice but may backfire.

gumb0 (approved):

PR #706, opened 2023-09-14
Fix MSVC compilation

chfast:


test/unittests/state_transition.cpp
```diff
@@ -33,7 +33,11 @@ void state_transition::TearDown()
     auto& selected_vm = trace ? tracing_vm : vm;
 
     /// Optionally enable trace capturing in form of a RAII object.
-    const auto trace_capture = trace ? std::optional<TraceCapture>{std::in_place} : std::nullopt;
+    const auto trace_capture = [trace]() {
```
chfast, line 36: This version uses proper copy elision (mandatory by C++ standard now?). But I guess the version with `.emplace()` should look better now.

gumb0: It kinda looks excessive to use lambda for this simple case. I would prefer to declare `trace_capture` non-const as originally, if no other way.

gumb0 (approved):

test/unittests/state_transition.cpp
chfast: I also prefer simpler version. Although it won't be "functional style".

gumb0 (approved):

PR #709, opened 2023-09-24
Implement EIP-4788: Beacon block root in the EVM

chfast:
Implement the system call from the "Block processing" of the EIP-4788.
https://eips.ethereum.org/EIPS/eip-4788#block-processing

At the beginning of each block a gas-free call is invoked from
the _system_ account (`0xff..fe`) to the specific _system_ contract.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -147,6 +147,41 @@ void delete_empty_accounts(State& state)
 }
 }  // namespace
 
+void system_call(State& state, const BlockInfo& block, evmc_revision rev, evmc::VM& vm)
+{
+    static constexpr auto SystemAddress = 0xfffffffffffffffffffffffffffffffffffffffe_address;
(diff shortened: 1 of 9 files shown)


test/state/state.cpp
```diff
@@ -147,6 +147,44 @@ void delete_empty_accounts(State& state)
 }
 }  // namespace
 
+void system_call(State& state, const BlockInfo& block, evmc_revision rev, evmc::VM& vm)
```
rodiazet, line 150: Can we easily add unit tests for this function?

gumb0 (approved):

gumb0 (changes requested):
Unit test fails on sanitizer build

chfast:
> Looks good, but unit test fails on sanitizer build

Noticed, but I don't know why yet.

rodiazet (approved):

gumb0 (approved):

PR #713, opened 2023-09-27
Support blob transaction type (EIP-4844)

rodiazet:
- Add blob tx type support in state. 
- Modify `transition` and `validate_transaction` functions to pass `blob_gas_left` parameter.
- Update unit tests
- Support blob tx in t8n
- Compute `excess_blob_gas` and `blob_gas_price`.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -72,6 +72,29 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 }
 }  // namespace
 
+intx::uint256 compute_blob_gas_price(uint64_t excess_blob_gas) noexcept
+{
+    /// A helper function approximating `factor * e ** (numerator / denominator)`.
+    /// https://eips.ethereum.org/EIPS/eip-4844#helpers
+    static constexpr auto fake_exponential = [](uint64_t factor, uint64_t numerator,
+                                                 uint64_t denominator) noexcept {
+        intx::uint256 i = 1;
+        intx::uint256 output = 0;
+        intx::uint256 numerator_accum = factor * denominator;
+        while (numerator_accum > 0)
+        {
+            output += numerator_accum;
+            numerator_accum = (numerator_accum * numerator) / (denominator * i);
+            i += 1;
+        }
+        return output / denominator;
+    };
+
+    static constexpr auto MIN_BLOB_GASPRICE = 1;
+    static constexpr auto BLOB_GASPRICE_UPDATE_FRACTION = 3338477;
+    return fake_exponential(MIN_BLOB_GASPRICE, excess_blob_gas, BLOB_GASPRICE_UPDATE_FRACTION);
+}
+
 /// Validates transaction and computes its execution gas limit (the amount of gas provided to EVM).
 /// @return  Execution gas limit or transaction validation error.
 std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -98,11 +98,28 @@ intx::uint256 compute_blob_gas_price(uint64_t excess_blob_gas) noexcept
 /// Validates transaction and computes its execution gas limit (the amount of gas provided to EVM).
 /// @return  Execution gas limit or transaction validation error.
 std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
-    const BlockInfo& block, const Transaction& tx, evmc_revision rev,
-    int64_t block_gas_left) noexcept
+    const BlockInfo& block, const Transaction& tx, evmc_revision rev, int64_t block_gas_left,
+    int64_t blob_gas_left) noexcept
 {
     switch (tx.type)
     {
+    case Transaction::Type::blob:
+        if (rev < EVMC_CANCUN)
+            return make_error_code(TX_TYPE_NOT_SUPPORTED);
+        if (!tx.to.has_value())
+            return make_error_code(CREATE_BLOB_TX);
+        if (tx.blob_hashes.empty())
+            return make_error_code(EMPTY_BLOB_HASHES_LIST);
+
+        if (tx.max_blob_gas_price < block.blob_base_fee)
+            return make_error_code(FEE_CAP_LESS_THEN_BLOCKS);
+
+        if (std::ranges::any_of(tx.blob_hashes, [](const auto& h) { return h.bytes[0] != 0x01; }))
+            return make_error_code(INVALID_BLOB_HASH_VERSION);
+        if (std::cmp_greater(tx.blob_gas_used(), blob_gas_left))
+            return make_error_code(BLOB_GAS_LIMIT_EXCEEDED);
+        [[fallthrough]];
+
     case Transaction::Type::eip1559:
         if (rev < EVMC_LONDON)
             return make_error_code(TX_TYPE_NOT_SUPPORTED);
@@ -146,9 +163,16 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     // Compute and check if sender has enough balance for the theoretical maximum transaction cost.
     // Note this is different from tx_max_cost computed with effective gas price later.
     // The computation cannot overflow if done with 512-bit precision.
-    if (const auto tx_cost_limit_512 =
-            umul(intx::uint256{tx.gas_limit}, tx.max_gas_price) + tx.value;
-        sender_acc.balance < tx_cost_limit_512)
+    auto max_total_fee = umul(intx::uint256{tx.gas_limit}, tx.max_gas_price);
+    max_total_fee += tx.value;
+
+    if (tx.type == Transaction::Type::blob)
+    {
+        const auto total_blob_gas = tx.blob_gas_used();
+        // FIXME: Can overflow uint256.
+        max_total_fee += total_blob_gas * tx.max_blob_gas_price;
+    }
+    if (sender_acc.balance < max_total_fee)
         return make_error_code(INSUFFICIENT_FUNDS);
 
     const auto execution_gas_limit = tx.gas_limit - compute_tx_intrinsic_cost(rev, tx);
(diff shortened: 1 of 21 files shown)


test/state/state.hpp
```diff
@@ -184,11 +200,11 @@ void finalize(State& state, evmc_revision rev, const address& coinbase,
 
 [[nodiscard]] std::variant<TransactionReceipt, std::error_code> transition(State& state,
     const BlockInfo& block, const Transaction& tx, evmc_revision rev, evmc::VM& vm,
-    int64_t block_gas_left);
+    int64_t block_gas_left, uint64_t blob_gas_left);
```
gumb0, line 203: Why did you make `blob_gas_left` unsigned, when `block_gas_left` is signed?

rodiazet: fixed. No idea actually.

chfast (commented):
How was this tested? I think we need to enable at least some external JSON tests on CI.

rodiazet:
> How was this tested? I think we need to enable at least some external JSON tests on CI.

Done by https://github.com/ethereum/evmone/pull/713/commits/3dfcf0de9ea4acbe686c826f5575698e096ab08c

test/state/state.hpp
```diff
@@ -165,6 +178,9 @@ struct TransactionReceipt
     /// Amount of gas used by this transaction.
     int64_t gas_used = 0;
 
+    /// Amount of blob gas used by this transaction.
+    int64_t blob_gas_used = 0;
```
chfast, line 182: This is always constant times number of blob hashes. Maybe replace this with a member `blob_gas_used() { GAS_PER_BLOB * blob_hashes.size() }`?

test/state/state.hpp
```diff
@@ -99,6 +99,13 @@ struct BlockInfo
     bytes32 prev_randao;
     uint64_t base_fee = 0;
     std::vector<Ommer> ommers = {};
+
+    /// The "excess blob gas" parameter from EIP-4844
+    /// for computing the blob gas price in the current block.
+    int64_t excess_blob_gas = 0;
+    /// Max amount of blob gas allowed in block. It's static now but can be dynamic in the future.
+    static constexpr int64_t max_blob_gas_per_block = 786432;
```
chfast, line 107:
```suggestion
    static constexpr int64_t MaxBlobGasPerBlock = 786432;
```

test/state/state.cpp
```diff
@@ -307,6 +368,20 @@ std::variant<TransactionReceipt, std::error_code> transition(State& state, const
                    tx.to.has_value() ? tx.to.value() : bytes_view(), tx.value, tx.data,
                    tx.access_list, static_cast<bool>(tx.v), tx.r, tx.s);
     }
+    else
+    {
+        if (tx.v > 1)
+            throw std::invalid_argument("`v` value for eip1559 transaction must be 0 or 1");
```
chfast, line 374:
```suggestion
            throw std::invalid_argument("`v` value for blob transaction must be 0 or 1");
```

test/state/state.hpp
rodiazet: We don't have access to `blob_hashes.size()` in the context of `TransactionReceipt`  struct.

test/state/state.cpp
```diff
@@ -49,8 +65,10 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
     const auto initcode_cost =
         is_create && rev >= EVMC_SHANGHAI ? initcode_word_cost * num_words(tx.data.size()) : 0;
     const auto tx_cost = is_create && rev >= EVMC_HOMESTEAD ? create_tx_cost : call_tx_cost;
-    return tx_cost + compute_tx_data_cost(rev, tx.data) + compute_access_list_cost(tx.access_list) +
-           initcode_cost;
+    auto c = tx_cost + compute_tx_data_cost(rev, tx.data) +
```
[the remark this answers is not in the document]
rodiazet: fixed

test/state/state.cpp
```diff
@@ -75,11 +93,32 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
...
+            return make_error_code(EMPTY_BLOB_HASHES_LIST);
+        if (tx.blob_hashes.size() > 6)
+            return make_error_code(BLOB_HASHES_LIST_SIZE_LIMIT_EXCEEDED);
+
+        if (tx.max_blob_gas_price < blob_gas_price)
+            return make_error_code(FEE_CAP_LESS_THEN_BLOCKS);
+
+        if (std::ranges::any_of(tx.blob_hashes, [](const auto& h) { return h.bytes[0] != 0x01; }))
+            return make_error_code(INVALID_BLOB_HASH_VERSION);
+        if (GAS_PER_BLOB * tx.blob_hashes.size() > static_cast<uint64_t>(blob_gas_left))
```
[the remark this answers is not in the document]
rodiazet: done

test/state/state.cpp
```diff
@@ -307,6 +368,20 @@ std::variant<TransactionReceipt, std::error_code> transition(State& state, const
...
+    {
+        if (tx.v > 1)
+            throw std::invalid_argument("`v` value for eip1559 transaction must be 0 or 1");
+        // tx_type +
+        // rlp [chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, to, value,
+        // data, access_list, max_fee_per_blob_gas, blob_versioned_hashes, sig_parity, r, s];
+        return bytes{stdx::to_underlying(Transaction::Type::blob)} +
+               rlp::encode_tuple(tx.chain_id, tx.nonce, tx.max_priority_gas_price, tx.max_gas_price,
+                   static_cast<uint64_t>(tx.gas_limit),
+                   tx.to.has_value() ? tx.to.value() : bytes_view(), tx.value, tx.data,
```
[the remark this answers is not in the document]
rodiazet: done

test/state/state.hpp
gumb0: You should probably get this value form a transaction, not from receipt. This is not part of receipt as defined in EIP.

test/state/state.hpp
```diff
@@ -123,17 +130,23 @@ struct Transaction
...
+        /// Introduced by EIP-4844 https://eips.ethereum.org/EIPS/eip-4844.
+        blob = 3,
     };
 
     Type type = Type::legacy;
     bytes data;
     int64_t gas_limit;
     intx::uint256 max_gas_price;
     intx::uint256 max_priority_gas_price;
+    intx::uint256 max_blob_gas_price = {};
```
gumb0, line 144: redundant
```suggestion
    intx::uint256 max_blob_gas_price;
```

test/state/state.hpp
```diff
@@ -123,17 +130,23 @@ struct Transaction
...
     address sender;
     std::optional<address> to;
     intx::uint256 value;
     AccessList access_list;
+    std::vector<bytes32> blob_hashes = {};
```
gumb0, line 149:
```suggestion
    std::vector<bytes32> blob_hashes;
```

test/state/state.cpp
```diff
@@ -75,11 +91,32 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
...
     switch (tx.type)
     {
+    case Transaction::Type::blob:
+        if (rev < EVMC_CANCUN)
+            return make_error_code(TX_TYPE_NOT_SUPPORTED);
+        if (!tx.to.has_value())
+            return make_error_code(CREATE_BLOB_TX);
+        if (tx.blob_hashes.empty())
+            return make_error_code(EMPTY_BLOB_HASHES_LIST);
+        if (tx.blob_hashes.size() > 6)
```
gumb0, line 108: I understand that this is a limit for number of blob hashes included in a block, and transaction validity doesn't include this?

test/state/state.cpp
```diff
@@ -75,11 +91,32 @@ evmc_message build_message(const Transaction& tx, int64_t execution_gas_limit) n
 /// Validates transaction and computes its execution gas limit (the amount of gas provided to EVM).
 /// @return  Execution gas limit or transaction validation error.
 std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
-    const BlockInfo& block, const Transaction& tx, evmc_revision rev,
-    int64_t block_gas_left) noexcept
+    const BlockInfo& block, const Transaction& tx, evmc_revision rev, int64_t block_gas_left,
+    int64_t blob_gas_left) noexcept
 {
+    const auto blob_gas_price = fake_exponential(1, block.excess_blob_gas, 3338477);
```
pdobacz, line 97: let's define these 2 constants somewhere

pdobacz (commented):
I need some more time to mull this over, but two comments from me so far

test/state/state.hpp
rodiazet: OK. I get Pawel's comment now. Thanks.

test/state/state.cpp
rodiazet: Yeah. IMO it's already handled by the condition but maybe @chfast wants to take a look also. I don't remember why this additional `BLOB_HASHES_LIST_SIZE_LIMIT_EXCEEDED` error was added. Maybe it's not needed any longer. I would remove it now.

chfast: If these checks are equivalent we can remove the `tx.blob_hashes.size() > 6`. I only recall discussions about having 3/6 or 2/4 blobs in the block. I'm not sure if blob gas cost handles this exact limit precisely and is enough.

chfast: Anyway, if the EIP does not mention 6 at all we proceed without it and check against tests.

rodiazet: `6*0x20000 == 786432` which is precisely `MAX_BLOB_GAS_PER_BLOCK`

test/state/state.cpp
chfast: Do you think we should go with Python naming convention for constants to be consistent with the spec? (sent from CLion PR review)

rodiazet: BTW. We should have some common place to define constants like those two.

pdobacz (approved):

test/state/state.cpp
```diff
@@ -14,6 +14,10 @@ namespace evmone::state
 {
 namespace
 {
+/// Values specified by EIP-4844
+constexpr auto min_blob_gasprice = 1;
```
chfast, line 18: Bring the constants to the single place they are use, which is `if (tx.type == Transaction::Type::blob)`.
Use naming `MIN_BLOB_GASPRICE` or `MinBlobGasPrice`.
They will be still searchable but no need to extend their scope unnecessarily.

test/state/state.cpp
```diff
@@ -307,6 +365,21 @@ std::variant<TransactionReceipt, std::error_code> transition(State& state, const
                    tx.to.has_value() ? tx.to.value() : bytes_view(), tx.value, tx.data,
                    tx.access_list, static_cast<bool>(tx.v), tx.r, tx.s);
     }
+    else
+    {
+        if (tx.v > 1)
+            throw std::invalid_argument("`v` value for blob transaction must be 0 or 1");
+        if (!tx.to.has_value())  // Blob tx has to have `to` address
+            throw std::invalid_argument("`to` value for blob transaction must not be null");
```
chfast, line 373: Codecov says this is uncovered by tests. Please add a test if true.

chfast (changes requested):

test/state/state.cpp
rodiazet: Test addded

chfast:
These tests from execution-spec-tests are failing: `cancun/eip4844_blobs/excess_blob_gas`.

test/statetest/statetest_loader.cpp
```diff
@@ -147,6 +147,29 @@ inline uint64_t calculate_current_base_fee_eip1559(
     return base_fee;
 }
 
+intx::uint256 compute_blob_gas_price(uint64_t excess_blob_gas) noexcept
```
chfast, line 150: This should stay in `state.cpp` because we want to have unit tests for it.

chfast (changes requested):

chfast (approved):

PR #721, opened 2023-10-23
Disallow unreachable code sections

hugo-dc:
This PR disallows EOF unreachable code sections. 

Pending:
- [x] Add new unit tests specific for unreachable code sections


lib/evmone/eof.cpp
```diff
@@ -430,6 +431,27 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
+    for (size_t i = 0; i < code.size(); ++i)
+    {
+        const auto opcode = code[i];
+        if (opcode == OP_CALLF)
+        {
+          const auto fid = read_uint16_be(&code[i + 1]);
+          accessed_code_sections.insert(fid);
+          i += 2;
+        } else
+          i += instr::traits[opcode].immediate_size;
```
pdobacz, line 450: should have a special case for `RJUMPV` like in https://github.com/ethereum/evmone/blob/master/lib/evmone/eof.cpp#L368

gumb0: Maybe better even to do it in `validate_instructions()` not  to repeat immediates handling. E.g. pass `accessed_code_sections` by reference there.

hugo-dc: Thanks for reviewing, I moved it to the `validate_instructions()`, it looks simpler.

lib/evmone/eof.cpp
```diff
@@ -212,7 +213,7 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
 }
 
 EOFValidationError validate_instructions(
-    evmc_revision rev, const EOF1Header& header, size_t code_idx, bytes_view container) noexcept
+    evmc_revision rev, const EOF1Header& header, size_t code_idx, bytes_view container, std::set<size_t> &accessed_code_sections) noexcept
```
gumb0, line 216:
```suggestion
    evmc_revision rev, const EOF1Header& header, size_t code_idx, bytes_view container, std::set<size_t>& accessed_code_sections) noexcept
```

gumb0:
Please rebase and take into account sections targeted by JUMPF.

gumb0:
Also please add a validation unit test with unreachable sections error.

test/unittests/eof_test.cpp
```diff
@@ -38,9 +40,16 @@ TEST(eof, read_valid_eof1_header)
...
-    std::string section_size_1_256;
+    std::string section_size_256;
     for (int i = 0; i < 256; ++i)
-        section_size_1_256 += "0001";
+        section_size_256 += "0003";
+
+    std::string code_sections_256;
+    for (int i = 0; i < 256; ++i)
+        if (i < 255)
+            code_sections_256 += "E5" + std::format("{:04X}", i + 1);
```
gumb0, line 50: I would probably do something like
```suggestion
            code_sections_256 += "E5" + hex(big_endian(static_cast<uint16>(i + 1)));
```

gumb0:
Note also you can now build an EOF bytecode with multiple section using `bytecode.hpp` utility, like here:
https://github.com/ethereum/evmone/blob/3d0242bcb0b95d33d39b2c078d597191fffbf52b/test/unittests/evm_eof_function_test.cpp#L97-L100

test/unittests/eof_test.cpp
gumb0: It looks also some CI builds with old compilers fail because `std::format` is not available.

hugo-dc:
> Note also you can now build an EOF bytecode with multiple section using `bytecode.hpp` utility, like here:
> 
>...

In new tests I'm using that approach: https://github.com/ethereum/evmone/blob/7f1487c33e3f8974e7013140100665465a3c8a24/test/unittests/eof_validation_test.cpp#L1202-L1204

However, in some other parts, I'm generating the code "manually" in order to generate many code sections where each section calls the next one: https://github.com/ethereum/evmone/blob/7f1487c33e3f8974e7013140100665465a3c8a24/test/unittests/eof_validation_test.cpp#L1278-L1281 

Not sure if there is a way to do something like that by using the `bytecode.hpp` utility.

lib/evmone/eof.cpp
```diff
@@ -514,11 +519,13 @@ std::variant<EOF1Header, EOFValidationError> validate_eof1(
         offset += code_size;
     }
 
+    std::set<uint16_t> accessed_code_sections = {0};
```
chfast, line 522: You probably should use `std::unordered_set` as we don't care if the inserted values are in order.

test/unittests/eof_validation_test.cpp
```diff
@@ -8,6 +8,7 @@
 #include <test/utils/bytecode.hpp>
 #include <test/utils/utils.hpp>
 
+using namespace std;
```
chfast, line 11: Not sure where this is used, but we are rather explicit about using `std::`.

test/unittests/eof_test.cpp
```diff
@@ -34,33 +35,40 @@ TEST(eof, read_valid_eof1_header)
...
-    for (int i = 0; i < 256; ++i)
-        section_size_1_256 += "0001";
 
+    std::string nops_255;
+    std::string section_size_256;
+    std::string code_sections_256;
     std::string section_types_256;
     for (int i = 0; i < 256; ++i)
+    {
+        section_size_256 += "0003";
```
gumb0, line 45: I found a more concise way of making these sequences, I think I'll make a separate PR with it.

test/unittests/eof_validation_test.cpp
```diff
@@ -310,8 +312,15 @@ TEST(eof_validation, EOF1_invalid_section_0_type)
...
+        if (i < 1023)
+            code_sections_1024 += "E5" + hex(big_endian(static_cast<uint16_t>(i + 1)));
+        else
+            code_sections_1024 += "5B5B00";
+
+    const auto valid = "EF0001 011000" + bytecode{"020400"} + 0x400 * bytecode{"0003"} +
+                       "040000 00" + 0x400 * bytecode{"00800000"} + code_sections_1024;
     EXPECT_EQ(validate_eof(valid), EOFValidationError::success);
 
     const auto invalid = "EF0001 011002" + bytecode{"020401"} + 0x401 * bytecode{"0001"} +
```
gumb0, line 326: It would make ssense to use `code_sections_1024` in invalid case, too, just adding one more after them.

test/unittests/eof_validation_test.cpp
```diff
@@ -1065,7 +1093,7 @@ TEST(eof_validation, jumpf_into_returning_stack_validation)
     // JUMPF into a function with the same number of outputs as current one
 
     // Exactly required inputs on stack at JUMPF
-    EXPECT_EQ(validate_eof(eof_bytecode(bytecode{OP_STOP})
+    EXPECT_EQ(validate_eof(eof_bytecode(bytecode{OP_CALLF} + "0001" + OP_STOP, 2)
```
gumb0, line 1096: Confused why this works, this section has `max_stack_height` 0, not 2, doesn't it?

lib/evmone/eof.cpp
hugo-dc: It's using `unordered_set` now.

test/unittests/eof_validation_test.cpp
hugo-dc: I added another variable `code_sections_1025` in order that section 1024 is calling section 1025 (so we don't have unreachable code sections).

chfast (approved):
Issues I reported were fixed.

test/unittests/eof_test.cpp
gumb0: I did it in https://github.com/ethereum/evmone/pull/796, please rebase now

test/unittests/eof_validation_test.cpp
```diff
@@ -1161,3 +1184,78 @@ TEST(eof_validation, jumpf_with_inputs_stack_overflow)
...
+
+        EXPECT_EQ(validate_eof(code), EOFValidationError::unreachable_code_sections);
+    }
+
+    {
+        std::string code_sections_256_err_001 = "E50001E50001";
+        std::string code_sections_256_err_254 = "E50001E50002";
+        std::string cs_calling_next;
+        for (int i = 0; i < 252; ++i)
+            cs_calling_next += "E5" + hex(big_endian(static_cast<uint16_t>(i + 1)));
```
gumb0, line 1241: Not sure this does what you intended, maybe loop index is wrong, e.g. `code_sections_256_err_254` seems to end up with a sequence `jumpf{1} jumpf{2} jumpf{1} jumpf{2} jumpf{3} ...`
I think intention was to have `jumpf{1} jumpf{2} jumpf{3} jumpf{4} ...`

Also, if you want to use `bytecode` here, it should be possible with something like:
```cpp
auto code_sections_256_err_001 = eof_bytecode(bytecode{JUMPF} + "0001").code(bytecode{JUMPF} + "0002", 0, 0x80, 0);
for (int i = 2; i < 254; ++i)
  code_sections_256_err_001.code(bytecode{JUMPF} + big_endian(static_cast<uint16_t>(i + 1)), 0, 0x80, 0); 
```

(Also would be nice to have helpers in `bytecode.hpp` for callf / jumpf to not write `bytecode{JUMPF}` every time, but that's probably better for separate PR)

hugo-dc: There was an error in the loop indices, it is fixed and it is now using `bytecode`.

gumb0:
Looks fine to me now, could you squash the commits?

hugo-dc:
> Looks fine to me now, could you squash the commits?

Commits squashed

gumb0 (approved):

PR #726, opened 2023-11-06
Lookup CREATE4 tx initcodes by hash

pdobacz:
Requires https://github.com/ethereum/evmc/pull/699

### DIFF
--- a/test/state/host.hpp
+++ b/test/state/host.hpp
@@ -36,20 +36,17 @@ class Host : public evmc::Host
     const BlockInfo& m_block;
     const Transaction& m_tx;
     std::vector<Log> m_logs;
-    std::vector<const uint8_t*> m_initcode_ptrs;
-    std::vector<size_t> m_initcode_sizes;
+    std::unordered_map<hash256, bytes_view> m_initcodes;
 
 public:
     Host(evmc_revision rev, evmc::VM& vm, State& state, const BlockInfo& block,
         const Transaction& tx) noexcept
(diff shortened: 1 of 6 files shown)


test/state/host.hpp
```diff
@@ -36,20 +36,17 @@ class Host : public evmc::Host
     const BlockInfo& m_block;
     const Transaction& m_tx;
     std::vector<Log> m_logs;
-    std::vector<const uint8_t*> m_initcode_ptrs;
-    std::vector<size_t> m_initcode_sizes;
+    std::unordered_map<hash256, bytes> m_initcodes;
```
gumb0, line 39: You can store `byte_view`s to data stored in `m_tx` to avoid extra copying.

gumb0:
Looks fine overall

test/utils/bytecode.hpp
```diff
@@ -524,9 +524,9 @@ struct create_instruction
     }
 
     template <Opcode k = kind>
-    typename std::enable_if<k == OP_CREATE4, create_instruction&>::type initcode(uint8_t index)
+    typename std::enable_if<k == OP_CREATE4, create_instruction&>::type initcode(bytecode hash)
```
gumb0, line 527: Argument shouldn't be `bytecode`, because this is just immediate 32 bytes and we don't need here the flexibility of any bytecode producing this argument as for regular opcodes with stack arguments. So I thihk it should be just `bytes32` or `hash256`, same for `m_initcode_hash` member.

gumb0 (approved):

PR #727, opened 2023-11-09
Rename error non_empty_stack_on_terminating_instruction

hugo-dc:
Error `non_empty_stack_on_terminating_instruction` is currently only used when there is a [mismatch between the number of stack items in current function context and the total number of defined outputs](https://github.com/ethereum/evmone/blob/master/lib/evmone/eof.cpp#L421-L422), which is no longer accurate. 

This PR changes the error name to indicate outputs mismatch.

### DIFF
--- a/lib/evmone/eof.hpp
+++ b/lib/evmone/eof.hpp
@@ -91,7 +91,7 @@ enum class EOFValidationError
     invalid_max_stack_height,
     no_terminating_instruction,
     stack_height_mismatch,
-    non_empty_stack_on_terminating_instruction,
+    code_section_outputs_mismatch,
     max_stack_height_above_limit,
     inputs_outputs_num_above_limit,
     unreachable_instructions,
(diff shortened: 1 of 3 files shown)


lib/evmone/eof.hpp
```diff
@@ -91,7 +91,7 @@ enum class EOFValidationError
     invalid_max_stack_height,
     no_terminating_instruction,
     stack_height_mismatch,
-    non_empty_stack_on_terminating_instruction,
+    code_section_outputs_mismatch_on_retf,
```
gumb0, line 94: It is also used for JUMPF in #644 
```suggestion
    code_section_outputs_mismatch,
```

gumb0:
Also please change commit message to `Rename error non_empty_stack_on_terminating_instruction` (we keep the first line capitalized as title and short so that github does not cut it in UI)

gumb0:
Looks good, but can you leave only RETF change in this PR, then we will merge it, and then I'll rebase JUMPF?

gumb0:
> Looks good, but can you leave only RETF change in this PR, then we will merge it, and then I'll rebase JUMPF?

Now JUMPF is merged, you can just rebase this PR instead.

gumb0 (approved):

PR #732, opened 2023-11-22
Refactor precompiles detection

chfast:
Extract a precompile address recognition into new is_precompile()
function and use it in all places where this information is needed.

Improve maintenance of "precompile ids" by adding aliases to the enum.

Add unit tests for is_precompile().

### DIFF
--- a/test/state/precompiles.hpp
+++ b/test/state/precompiles.hpp
@@ -9,9 +9,7 @@
 
 namespace evmone::state
 {
-/// The total number of known precompiles ids, including 0.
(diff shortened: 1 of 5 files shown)


test/state/precompiles.hpp
```diff
@@ -9,9 +9,7 @@
 
 namespace evmone::state
 {
-/// The total number of known precompiles ids, including 0.
-inline constexpr std::size_t NumPrecompiles = 10;
-
+/// The precompiles identifies and their corresponding addresses.
```
rodiazet, line 12: `identifies - > identifiers` ?

PR #733, opened 2023-11-23
test: Improve maintainability of JSON loading

chfast:
- Use designated initializers for big structs like `BlockInfo` and `BlockHeader`. They show how JSON names are mapped to field names and help avoiding complicated merge conflicts.
- Load JSON integers as unsigned values + cast to signed if needed. This workarounds the problem with execution-spec-tests where timestamps can be uint64.
- Omit 0 storage values from state dumps in `evmone-blockchaintest`.

### DIFF
--- a/test/unittests/statetest_loader_test.cpp
+++ b/test/unittests/statetest_loader_test.cpp
@@ -52,9 +52,16 @@ TEST(json_loader, int64_t)
         from_json<int64_t>(basic_json("9223372036854775807")), std::numeric_limits<int64_t>::max());
     EXPECT_EQ(from_json<int64_t>(basic_json("-9223372036854775808")),
         std::numeric_limits<int64_t>::min());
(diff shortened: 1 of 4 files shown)


test/unittests/statetest_loader_test.cpp
```diff
@@ -52,9 +52,13 @@ TEST(json_loader, int64_t)
         from_json<int64_t>(basic_json("9223372036854775807")), std::numeric_limits<int64_t>::max());
     EXPECT_EQ(from_json<int64_t>(basic_json("-9223372036854775808")),
         std::numeric_limits<int64_t>::min());
-    EXPECT_THROW(from_json<int64_t>(basic_json("0xffffffffffffffff")), std::out_of_range);
```
gumb0, line 55: Maybe add another test with `std::out_of_range` result?

gumb0 (approved):

test/unittests/statetest_loader_test.cpp
chfast: Added, thanks.

PR #735, opened 2023-11-24
Implement EIP-6780 "SELFDESTRUCT"

chfast:
Implement the EIP-6780 "SELFDESTRUCT only in same transaction".
https://eips.ethereum.org/EIPS/eip-6780

This EIP changes the functionality of the `SELFDESTRUCT` instruction.
The new functionality will be only to send all Ether in the account
to the beneficiary, except that the current behavior is preserved when
`SELFDESTRUCT` is called in the same transaction a contract was created.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -100,12 +100,27 @@ size_t Host::copy_code(const address& addr, size_t code_offset, uint8_t* buffer_
 
 bool Host::selfdestruct(const address& addr, const address& beneficiary) noexcept
 {
-    // Touch beneficiary and transfer all balance to it.
-    // This may happen multiple times per single account as account's balance
-    // can be increased with a call following previous selfdestruct.
     auto& acc = m_state.get(addr);
-    m_state.touch(beneficiary).balance += acc.balance;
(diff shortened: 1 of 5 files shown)


test/state/host.cpp
```diff
@@ -100,12 +100,27 @@ size_t Host::copy_code(const address& addr, size_t code_offset, uint8_t* buffer_
...
+    const auto balance = acc.balance;
+    auto& beneficiary_acc = m_state.touch(beneficiary);
+
+    if (m_rev >= EVMC_CANCUN && !acc.just_created)
+    {
+        // EIP-6780:
+        // "SELFDESTRUCT is executed in a transaction that is not the same
+        // as the contract calling SELFDESTRUCT was created"
+        acc.balance = 0;
+        beneficiary_acc.balance += balance;
```
pdobacz, line 113: add comment, that it is per the EIP to have different order and behavior of the `acc == beneficiary_acc` case.
```suggestion
        acc.balance = 0;
        beneficiary_acc.balance += balance;  // Keep balance if acc is the beneficiary.
```

pdobacz (approved):

PR #742, opened 2023-12-05
Evmmax instructions in evmone

rodiazet:
- Support new instruction implementation signature. `code_iterator (*instr_fn)(StackTop, ExecutionState&, code_iterator, int64_t&)`
- Add `EVMMAX` state interface
- Add `EVMMAX` modulus interface and implementation for different `UintT` types
- Add simple unit tests.


lib/evmmax/evmmax.cpp
```diff
@@ -8,6 +8,342 @@ using namespace intx;
...
+            values[dst_val_idx + i] =
+                arith.to_mont(load<UintT>(in_ptr + value_size_mult * 8 * i, value_size_mult * 8));
+        }
+
+        return true;
+    }
+
+    [[nodiscard]] bool addmodx(size_t dst_idx, size_t x_idx, size_t y_idx) noexcept override
+    {
+        if (dst_idx < values.size() && x_idx < values.size() && y_idx < values.size())
```
chfast, line 118: I think we should assume the indexes are valid (because they are going to be validated by EOF). Debug asserts are fine obviously.

kmk142789 (approved):

PR #743, opened 2023-12-08
test: Export state transition tests to JSON

chfast:
Add option to export the "state_transition" test suite to JSON State Test format.

The current version of the export feature has limitations:
- tests with invalid transaction are not exported
- not every test property is properly exported
- some exported tests fail in `evmone-statetest`

### DIFF
--- a/test/statetest/statetest_export.cpp
+++ b/test/statetest/statetest_export.cpp
@@ -1,0 +1,28 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "statetest.hpp"
+
+namespace evmone::test
+{
+json::json to_json(const std::unordered_map<address, state::Account>& accounts)
+{
+    json::json j;
+    for (const auto& [addr, acc] : accounts)
+    {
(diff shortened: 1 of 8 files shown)


test/statetest/statetest_export.cpp
```diff
@@ -0,0 +1,28 @@
...
+{
+    json::json j;
+    for (const auto& [addr, acc] : accounts)
+    {
+        auto& j_acc = j[hex0x(addr)];
+        j_acc["nonce"] = hex0x(acc.nonce);
+        j_acc["balance"] = hex0x(acc.balance);
+        j_acc["code"] = hex0x(bytes_view(acc.code.data(), acc.code.size()));
+
+        auto& j_storage = j_acc["storage"] = json::json::object();
```
gumb0, line 19: this clever trick makes me look up what `operator=` returns

gumb0:
Looks good overall, would be nice to fix windows builds.

test/statetest/statetest_export.cpp
chfast: This is to force the type of `"storage"` to an object. Otherwise, for empty storage we get `null` instead of `{}`.

gumb0 (approved):

PR #745, opened 2023-12-11
Fix and update CREATE4 failure scenarios

pdobacz:
- adjusts failure scenarios to new spec after https://github.com/ipsilon/eof/commit/e927c283e881136acb99981190dd4377463e6cd5
- fixes a stack management bug in CREATE4 light failure modes + adds test
- as a bonus it updates the validity rules for `InitcodeTransaction` and adds an `rlp_encode` specialization for it.

Do not merge before [changes to `evmc`](https://github.com/ethereum/evmc/pull/700) are merged (collapsing `EVMC_CREATE3` and `4`)

### DIFF
--- a/lib/evmone/eof.hpp
+++ b/lib/evmone/eof.hpp
@@ -65,6 +65,15 @@ struct EOF1Header
         return container.substr(data_offset);
     }
 
+    /// A helper to check whether the container can be an initcontainer.
+    [[nodiscard]] bool can_init(bytes_view container) const noexcept
+    {
+        // Containers with truncated data section cannot be initcontainers.
+        const auto truncated_data =
+            static_cast<size_t>(data_offset + data_size) != container.size();
+        return !truncated_data;
+    }
+
     /// A helper to extract reference to a specific container section.
     [[nodiscard]] bytes_view get_container(
         bytes_view container, size_t container_idx) const noexcept
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -208,6 +208,15 @@ Result create_eof_impl(
     if (state.in_static_mode())
         return {EVMC_STATIC_MODE_VIOLATION, gas_left};
 
+    const auto initcode_hash = (Op == OP_CREATE4) ? stack.pop() : uint256{};
+    const auto endowment = stack.pop();
+    const auto salt = stack.pop();
+    const auto input_offset_u256 = stack.pop();
+    const auto input_size_u256 = stack.pop();
+
+    stack.push(0);  // Assume failure.
+    state.return_data.clear();
+
     bytes_view initcontainer;
     if constexpr (Op == OP_CREATE3)
     {
@@ -220,14 +229,12 @@ Result create_eof_impl(
     else
     {
         pos += 1;
-        const evmc_bytes32 initcode_hash = intx::be::store<evmc::bytes32>(stack.pop());
 
-        const auto initcode = state.get_tx_initcode_by_hash(initcode_hash);
+        const auto initcode =
+            state.get_tx_initcode_by_hash(intx::be::store<evmc::bytes32>(initcode_hash));
         if (initcode.data == nullptr)
-        {
-            stack.push(0);
             return {EVMC_SUCCESS, gas_left};  // "Light" failure
-        }
+
         initcontainer = {initcode.data, initcode.size};
 
         // Charge for initcode validation.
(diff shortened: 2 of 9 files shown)


lib/evmone/instructions_calls.cpp
```diff
@@ -226,9 +226,33 @@ Result create_eof_impl(
         if (initcode.data == nullptr)
         {
             stack.push(0);
-            return {EVMC_SUCCESS, gas_left};
+            return {EVMC_SUCCESS, gas_left};  // "Light" failure
         }
         initcontainer = {initcode.data, initcode.size};
+
+        // Charge for initcode validation.
+        const auto initcode_word_cost_validation = 2;
```
gumb0, line 241: Make it consexpr

lib/evmone/instructions_calls.cpp
```diff
@@ -226,9 +226,33 @@ Result create_eof_impl(
...
+
+        if (error_subcont != EOFValidationError::success)
+        {
+            stack.push(0);
+            return {EVMC_SUCCESS, gas_left};  // "Light" failure.
+        }
+
+        const auto initcontainer_header = read_valid_eof1_header(initcontainer);
+        const auto size_from_header =
+            static_cast<size_t>(initcontainer_header.data_offset + initcontainer_header.data_size);
```
gumb0, line 254: I would probably make this calculation a helper member of `EOF1Header`

lib/evmone/instructions_calls.cpp
```diff
@@ -259,9 +277,15 @@ Result create_eof_impl(
         intx::be::load<uint256>(state.host.get_balance(state.msg->recipient)) < endowment)
         return {EVMC_SUCCESS, gas_left};  // "Light" failure.
 
+    // Charge for initcode hashing.
+    const auto initcode_word_cost_hashing = 6;
```
gumb0, line 275: constexpr

lib/evmone/instructions_calls.cpp
gumb0: ah I see it was like this in the old code, fine to leave it like this then, but you could also move it

test/state/state.cpp
```diff
@@ -82,8 +83,11 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     case Transaction::Type::initcodes:
         if (rev < EVMC_PRAGUE)
             return make_error_code(TX_TYPE_NOT_SUPPORTED);
-        if (tx.initcodes.size() > 4096)
+        if (tx.initcodes.size() > max_initcode_count)
             return make_error_code(INIT_CODE_SIZE_LIMIT_EXCEEDED);
+        if (rev >= EVMC_PRAGUE && std::any_of(tx.initcodes.begin(), tx.initcodes.end(),
```
gumb0, line 88: `rev` check is not needed, because it was already checked 3 lines above?

test/state/state.cpp
```diff
@@ -82,8 +83,11 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
...
+                                      [](const bytes& v) { return v.size() > max_initcode_size; }))
+            return make_error_code(INIT_CODE_COUNT_LIMIT_EXCEEDED);
```
gumb0, line 90: This and the above check have error codes swapped (this one should be `INIT_CODE_SIZE_LIMIT_EXCEEDED`)

lib/evmone/instructions_calls.cpp
pdobacz: I noticed that the reason and logic of this condition is a bit obscure, so tried moving it together to a helper `EOF1Header::can_init`. Looks much more readable this way

lib/evmone/eof.hpp
```diff
@@ -65,6 +65,14 @@ struct EOF1Header
         return container.substr(data_offset);
     }
 
+    /// A helper to check whether the container can be an initcontainer.
+    [[nodiscard]] bool can_init(bytes_view container) const noexcept
+    {
+        // Containers with truncated data section cannot be initcontainers.
```
gumb0, line 71: I would add at least
```cpp
assert(static_cast<size_t>(data_offset + data_size) >= container.size());
```
or just made the check to use `!=`

gumb0:
Looks good, save for one comment. Feel free to cherry-pick into `eof-create4` if it's easier than merging.

lib/evmone/eof.hpp
pdobacz: yea, I had a 2nd thought to do this with `!=` but didn't get around to fixing this with all the rebase stuff now

gumb0 (approved):

PR #747, opened 2023-12-13
EOF bytecode test helper refactor

gumb0:
New builder-like utility in `bytecode.hpp` to generate EOF containers instead of old `eof1_bytecode()` function.

### DIFF
--- a/test/utils/bytecode.hpp
+++ b/test/utils/bytecode.hpp
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <evmc/evmc.hpp>
+#include <evmone/eof.hpp>
 #include <evmone/instructions_traits.hpp>
 #include <intx/intx.hpp>
 #include <test/utils/utils.hpp>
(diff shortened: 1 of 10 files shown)


pdobacz (approved):

test/utils/bytecode.hpp
```diff
@@ -90,30 +91,79 @@ big_endian(T value)
...
+        bytecode out{eof_header()};
+        for (const auto& code : m_codes)
+            out += code;
+        out += m_data;
+        return out;
+    }
+};
 
-inline bytecode eof1_bytecode(bytecode code, uint16_t max_stack_height = 0, bytecode data = {})
+inline bytecode push0()
```
chfast, line 164: Please move it closer to other pushes...

test/utils/bytecode.hpp
```diff
@@ -90,30 +91,79 @@ big_endian(T value)
...
+        return *this;
+    }
+
+    auto& data(bytecode d)
+    {
+        m_data = std::move(d);
+        return *this;
+    }
+
+    operator bytecode() const
```
chfast, line 154:
```suggestion
    explicit(false) operator bytecode() const
```
Welcome to modern C++.

test/utils/bytecode.hpp
```diff
@@ -90,30 +91,79 @@ big_endian(T value)
...
+private:
+    std::vector<bytecode> m_codes;
+    std::vector<evmone::EOFCodeType> m_types;
+    bytecode m_data;
 
-inline bytecode eof1_header(uint16_t code_size, uint16_t max_stack_height, uint16_t data_size = 0)
-{
-    return eof_header(1, code_size, max_stack_height, data_size);
-}
+    bytecode eof_header() const
```
chfast, line 101: Add a doc comment.

chfast, line 101:
```suggestion
    bytecode header() const
```
Maybe shorter name?

test/utils/bytecode.hpp
```diff
@@ -90,30 +91,79 @@ big_endian(T value)
     return {static_cast<uint8_t>(value >> 8), static_cast<uint8_t>(value)};
 }
 
-inline bytecode eof_header(
-    uint8_t version, uint16_t code_size, uint16_t max_stack_height, uint16_t data_size)
+struct eof_bytecode
```
chfast, line 94: I don't mind leaving it as is because we are not consistent in any way, but you can rename it to CamelCase if you prefer.

chfast (approved):

test/utils/bytecode.hpp
gumb0: It is right above all other pushes?

test/utils/bytecode.hpp
gumb0: This is the same as default without it?

test/utils/bytecode.hpp
gumb0: I would leave it for consistency with `call_instruction`, `create_instruction`.

PR #751, opened 2023-12-19
MPT refactoring

chfast:
Some refactoring and improvements to the MPT implementation:
- Refactor Path to behave like standard container.
- Refactor the `Kind` enum and path encoding.
- Change path mismatch handling by using iterators.

### DIFF
--- a/test/state/mpt.cpp
+++ b/test/state/mpt.cpp
@@ -53,23 +53,9 @@ public:
     }
 
     [[nodiscard]] static constexpr size_t capacity() noexcept { return max_size; }
-    [[nodiscard]] size_t size() const noexcept { return m_size; }
     [[nodiscard]] bool empty() const noexcept { return m_size == 0; }
-    [[nodiscard]] uint8_t operator[](size_t index) const noexcept { return m_nibbles[index]; }
     [[nodiscard]] const uint8_t* begin() const noexcept { return m_nibbles; }
     [[nodiscard]] const uint8_t* end() const noexcept { return m_nibbles + m_size; }
-
-    [[nodiscard]] Path tail(size_t pos) const noexcept
-    {
-        assert(pos > 0 && pos <= m_size);  // MPT never requests whole path copy (pos == 0).
-        return {begin() + pos, end()};
-    }
-
-    [[nodiscard]] Path head(size_t size) const noexcept
-    {
-        assert(size < m_size);  // MPT never requests whole path copy (size == length).
-        return {begin(), begin() + size};
-    }
 
     [[nodiscard]] bytes encode(Kind kind) const
     {
@@ -139,13 +125,6 @@ class MPTNode
                                  std::move(br);
     }
 
-    /// Finds the position at witch two paths differ.
-    static size_t mismatch(const Path& p1, const Path& p2) noexcept
-    {
-        assert(p1.size() <= p2.size());
-        return static_cast<size_t>(std::ranges::mismatch(p1, p2).in1 - p1.begin());
-    }
-
 public:
     MPTNode() = default;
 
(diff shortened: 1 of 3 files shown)


test/state/mpt.cpp
```diff
@@ -132,12 +127,37 @@ class MPTNode
...
+        uint8_t nibble1 = invalid;  ///< Mismatched nibble in path 1. Invalid if paths match.
+        uint8_t nibble2 = invalid;  ///< Mismatched nibble in second path.
+        Path tail1;                 ///< Remaining tail after the nibble in path 1.
+        Path tail2;                 ///< Remaining tail after the nibble in path 2.
+    };
+
+    /// Finds the difference between two paths.
+    static MismatchResult mismatch(const Path& p1, const Path& p2) noexcept
+    {
+        assert(p1.length <= p2.length);  // no support for inserting keys of different lengths
```
gumb0, line 134: If there's no support for different lengths, why doesn't this assert equality of them?

chfast: These are not the full paths (the ones being inserted by users) but sub-paths of the nodes. Once a prefix of a path is processes its tails (produced in this function) is used further down the trie.

chfast: Extended this documentation section as I gathered new information.

gumb0: Still this comment here is confusing. This is not even an inserting function.

test/state/mpt.cpp
```diff
@@ -132,12 +116,37 @@ class MPTNode
...
+        const auto [it1, it2] = std::mismatch(p1.nibbles, p1.nibbles + p1.length, p2.nibbles);
+
+        const Path common{static_cast<size_t>(it1 - p1.nibbles), p1.nibbles};
+        const auto matched = common.length == p1.length;
```
gumb0, line 138: Subjectively easier to understand could be maybe
```suggestion
        const auto matched = it1 == (p1.nibbles + p1.length);
```

test/state/mpt.cpp
```diff
@@ -132,12 +116,37 @@ class MPTNode
...
+        const auto tail_start = common.length + 1;
+        assert(!matched || p1.length != p2.length);  // matched => paths have different lengths
+        assert(matched || tail_start <= p1.length);  // !matched => tail_start is valid
+
+        return {
+            .common = common,
+            .nibble1 = static_cast<uint8_t>(matched ? MismatchResult::invalid : *it1),
+            .nibble2 = *it2,
```
gumb0, line 146: Is this guaranteed to be valid by the assertions above? I don't see it immediately, maybe add explicit assertion for the value of `it2`?

test/state/mpt.cpp
```diff
@@ -132,12 +116,37 @@ class MPTNode
...
+    /// Finds the difference between two paths.
+    static MismatchResult mismatch(const Path& p1, const Path& p2) noexcept
+    {
+        assert(p1.length <= p2.length);  // no support for inserting keys of different lengths
+        const auto [it1, it2] = std::mismatch(p1.nibbles, p1.nibbles + p1.length, p2.nibbles);
+
+        const Path common{static_cast<size_t>(it1 - p1.nibbles), p1.nibbles};
+        const auto matched = common.length == p1.length;
+        const auto tail_start = common.length + 1;
+        assert(!matched || p1.length != p2.length);  // matched => paths have different lengths
```
gumb0, line 140: stronger assertion might also make it easier to understand what's happening 
```suggestion
        assert(!matched || p1.length < p2.length);  // matched => paths have different lengths
```

test/state/mpt.hpp
```diff
@@ -13,6 +13,13 @@ constexpr auto emptyMPTHash =
 
 /// Insert-only Merkle Patricia Trie implementation for getting the root hash
 /// out of (key, value) pairs.
+///
+/// Limitations:
+/// 1. The keys must not be longer than 32 bytes.
```
gumb0, line 18: Is this possible to have these limitations as explicut asserions in `insert()` ?

test/state/mpt.cpp
chfast: This assert was too much. It was replaced with a weaker one with extended explanation (hopefully understandable).

test/state/mpt.cpp
```diff
@@ -128,16 +141,15 @@ class MPTNode
...
     }
 
     /// Finds the position at witch two paths differ.
     static size_t mismatch(const Path& p1, const Path& p2) noexcept
     {
-        assert(p1.length <= p2.length);
-        return static_cast<size_t>(
-            std::mismatch(p1.nibbles, p1.nibbles + p1.length, p2.nibbles).first - p1.nibbles);
+        assert(p1.size() <= p2.size());
+        return static_cast<size_t>(std::ranges::mismatch(p1, p2).in1 - p1.begin());
```
gumb0: why is `std::ranges::mismatch()` better?

test/state/mpt.cpp
```diff
@@ -62,24 +70,13 @@ class Path
         return {begin(), begin() + size};
     }
 
-    [[nodiscard]] bytes encode(bool extended) const
+    [[nodiscard]] bytes encode(Kind kind) const
```
gumb0, line 59: Maybe worth to add assertion that `kind` is either `leaf` or `ext`

test/state/mpt.cpp
```diff
@@ -62,24 +70,13 @@ class Path
...
-            const auto l = m_nibbles[i];
-            assert(h <= 0x0f);
-            assert(l <= 0x0f);
-            bs.push_back(static_cast<uint8_t>((h << 4) | l));
-        }
-        if (!extended)
-            bs[0] |= 0x20;
+        bytes bs{static_cast<uint8_t>(
+            (is_even ? 0x00 : (0x10 | m_nibbles[0])) | (kind == Kind::leaf ? 0x20 : 0x00))};
+        for (size_t i = is_even ? 0 : 1; i < m_size; i += 2)
```
gumb0, line 64: This function got more compressed but harder to read.

test/state/mpt.hpp
chfast: The first 3 are protected by asserts, but not at the beginning of the `insert()` (as you need to traverse the tree to find out). However, I can consistently add messages to these asserts and document these here.

There simply is not method for the 4th restriction.

test/state/mpt.cpp
chfast: Noted.

test/state/mpt.cpp
chfast: You don't have to put a pair of iterators. It handles the case when the second range is shorter. And outputs both iterators.

test/state/mpt.hpp
chfast: We can also add "death" tests but felt like overkill to me.

test/state/mpt.cpp
gumb0: `std::mismatch()` also handles shorter second range in C++20 required by `std::range`, and outputs a pair of iterators, so the only convenience is passing containers instead of iterators, it seems.

test/state/mpt.cpp
```diff
@@ -160,57 +145,54 @@ void MPTNode::insert(const Path& path, bytes&& value)  // NOLINT(misc-no-recursi
     // in an existing branch node. Otherwise, we need to create new branch node
     // (possibly with an adjusted extended node) and transform existing nodes around it.
 
+    const auto [this_idx, insert_idx] = std::ranges::mismatch(m_path, path);
+
+    // insert_idx is always valid if requirements are fulfilled:
+    // - if m_path is not shorter than path they must have mismatched nibbles,
```
gumb0, line 151:
```suggestion
    // - if m_path is not shorter than path they must have mismatched nibbles,  given the requirement that key uniqeness and inserted key not being a prefix if existing key,
```
if I understand correctly

test/state/mpt.hpp
```diff
@@ -13,6 +13,14 @@ constexpr auto emptyMPTHash =
 
 /// Insert-only Merkle Patricia Trie implementation for getting the root hash
 /// out of (key, value) pairs.
+///
+/// Limitations:
+/// 1. A keys must not be longer than 32 bytes. Protected by debug assert.
```
gumb0, line 18:
```suggestion
/// 1. A key must not be longer than 32 bytes. Protected by debug assert.
```

test/unittests/state_mpt_test.cpp
```diff
@@ -85,6 +87,29 @@ TEST(state_mpt, branch_node_example1)
     EXPECT_EQ(hex(trie.hash()), "1aaa6f712413b9a115730852323deb5f5d796c29151a60a1f55f41a25354cd26");
 }
 
+TEST(state_mpt, branch_node_of_3)
+{
+    // A trie of single branch node and three leaf nodes with paths of length 2.
+    // The branch node has leaf nodes at positions [0], [1] and [2]. All leafs have path 0.
```
gumb0, line 93:
```suggestion
    // The branch node has leaf nodes at positions [0], [1] and [2]. All leafs have path 0.
```
```suggestion
    // The branch node has leaf nodes at positions [0], [1] and [2]. All leaves have path 0.
```

test/unittests/state_mpt_test.cpp
```diff
@@ -85,6 +87,29 @@ TEST(state_mpt, branch_node_example1)
...
+    MPT trie;
+    trie.insert("00"_hex, "X"_b);
+    trie.insert("10"_hex, "Y"_b);
+    trie.insert("20"_hex, "Z"_b);
+    EXPECT_EQ(hex(trie.hash()), "5c5154e8d108dcf8b9946c8d33730ec8178345ce9d36e6feed44f0134515482d");
+}
+
+TEST(state_mpt, leaf_node_with_empty_path)
+{
+    // Both inserted leafs have empty path in the end.
```
gumb0, line 105:
```suggestion
    // Both inserted leaves have empty path in the end.
```

gumb0 (approved):

test/state/mpt.cpp
chfast: Yes, correct. The range addition is for convenience and readability.

test/state/mpt.cpp
chfast: Is the new version any good?

PR #753, opened 2023-12-20
EOF stack overflow tests improvements

gumb0:


chfast (commented):
Do we have the change "declared max stack height must be _not smaller_ than the computed one" in? If yes, you can set the default and omit `max_stack_height` parameter.

gumb0:
> Do we have the change "declared max stack height must be _not smaller_ than the computed one" in? If yes, you can set the default and omit `max_stack_height` parameter.

No, we decided not to have it.

pdobacz (approved):

chfast (approved):

chfast:
> No, we decided not to have it.

In the spec or in the implementation for now?

gumb0:
> > No, we decided not to have it.
> 
> In the spec or in the implementation for now?

In the spec.

gumb0:
> > No, we decided not to have it.
> 
> In the spec or in the implementation for now?

Remembered now why: there was an argument that for compiler devs having it strict might be handy as an extra sanity check of max stack value that they calculate. And also that for the tests we could pull in our max stack calculation into the tests to not calculate it manually.

PR #759, opened 2023-12-28
mpt: Refactor Path to behave like standard container

chfast:


chfast:
Changes moved back to #751.

PR #763, opened 2024-01-02
state: Optimize "create" code deployment

chfast:
Optimize code deployment in "create" by reusing the reference to the new account. The reference remains valid because the state reverts are implemented with the journal.


test/state/host.cpp
```diff
@@ -210,6 +210,7 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
         collision_acc != nullptr && (collision_acc->nonce != 0 || !collision_acc->code.empty()))
         return evmc::Result{EVMC_FAILURE};
 
+    // TODO: msg.recipient lookup is done 3x here.
```
chfast, line 213: I added a TODO comment do handle this later. There are more optimizations coming for journal and is not worth to do it now.

pdobacz (approved):

PR #764, opened 2024-01-02
Refactor EOF1Header::can_init

pdobacz:
I noticed we had separate implementations of `EOF1Header::can_init` check (for CREATE3 and 4) so I merged them. Unfortunately, the clean version of this makes me include the container size as a `EOF1Header` field. This smells a little.

The alternative I had was to have `can_init(size_t container_size)` instead, but I didn't like it either. Any thoughts?

### DIFF
--- a/lib/evmone/eof.hpp
+++ b/lib/evmone/eof.hpp
@@ -32,6 +32,9 @@ struct EOF1Header
     /// The EOF version, 0 means legacy code.
     uint8_t version = 0;
 
+    /// Overall size of container. NOTE: Not part of EOFv1 container header.
+    size_t size;
+
     /// Size of every code section.
     std::vector<uint16_t> code_sizes;
 
@@ -66,11 +69,10 @@ struct EOF1Header
     }
 
     /// A helper to check whether the container can be an initcontainer.
-    [[nodiscard]] bool can_init(bytes_view container) const noexcept
+    [[nodiscard]] bool can_init() const noexcept
     {
         // Containers with truncated data section cannot be initcontainers.
-        const auto truncated_data =
-            static_cast<size_t>(data_offset + data_size) != container.size();
+        const auto truncated_data = static_cast<size_t>(data_offset + data_size) != size;
         return !truncated_data;
     }
 
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -314,9 +314,7 @@ EOFValidationError validate_instructions(evmc_revision rev, const EOF1Header& he
                 if (container_idx >= header.container_sizes.size())
                     return EOFValidationError::invalid_container_section_index;
 
-                if (op == OP_CREATE3 && subcontainer_headers[container_idx].data_offset +
-                                                subcontainer_headers[container_idx].data_size !=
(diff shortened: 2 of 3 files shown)


lib/evmone/eof.hpp
```diff
@@ -32,6 +32,9 @@ struct EOF1Header
     /// The EOF version, 0 means legacy code.
     uint8_t version = 0;
 
+    /// Overall size of container. NOTE: Not part of EOFv1 container header.
+    size_t size;
```
chfast, line 36:
```suggestion
    size_t size = 0;
```

chfast (commented):
Good for me, but I leave the final decision to @gumb0.

gumb0:
I think I would prefer `can_init(container_size)`, because so far `EOF1Header` was agnostic to anything outside of the header + type section, and only got container as method argument when it needed something from container body.

Also better to add this to `eof-create3` branch first, then rebase `eof-create4`.

pdobacz:
> I think I would prefer `can_init(container_size)`, because so far `EOF1Header` was agnostic to anything outside of the header + type section, and only got container as method argument when it needed something from container body.
> 
> Also better to add this to `eof-create3` branch first, then rebase `eof-create4`.

I reverted to `can_init(size_t container_size)` to abide by the convention. I would be careful here, as this convention invites the possibility of mismatching the `EOF1Header` with a different container - even more so since our code handles multiple containers sometimes. Just something to keep in mind.

As for adding this to `eof-create3` - that branch doesn't have `can_init` at all, because it only needed that check defined once, inline in `CREATE3`. I am ready to walk the extra mile to reorganize the commits in a way that `can_init()` exists from the very beginning, and the commit history is completely clean. But let me know if you're certain it's worth it :).

As in - what is our end game for `eof-create3` and `eof-create4`? Do we intend for them to merge to `master` with a perfect commit history (without intermediate forms), or squash merge, or merge keeping the current history, to document the curvy path we took to get to the end result?

gumb0:
> I reverted to `can_init(size_t container_size)` to abide by the convention. I would be careful here, as this convention invites the possibility of mismatching the `EOF1Header` with a different container - even more so since our code handles multiple containers sometimes. Just something to keep in mind.

I see this as a reflection of the design, where the same header can be used with various containers after appending or not appending the data.

> As for adding this to `eof-create3` - that branch doesn't have `can_init` at all, because it only needed that check defined once, inline in `CREATE3`. I am ready to walk the extra mile to reorganize the commits in a way that `can_init()` exists from the very beginning, and the commit history is completely clean. But let me know if you're certain it's worth it :).

On a second thought, I'm fine to introduce it in `eof-create4` only.

> As in - what is our end game for `eof-create3` and `eof-create4`? Do we intend for them to merge to `master` with a perfect commit history (without intermediate forms), or squash merge, or merge keeping the current history, to document the curvy path we took to get to the end result?

Clean commit history is my goal at least. I try ro squash most of the intermediate things.

gumb0:
I rebased `eof-create4` and this now needs rebase, too. Hopefully not too complicated, because `eof-create4` commits are simplified now.

gumb0 (approved):

PR #768, opened 2024-01-03
Enable EOF validation before execution in `evmc run`

### LINKED ISSUE #585: EOF validation in tools / tests
When an execution code is taken from external source (e.g. CLI or JSON tests) we should perform EOF validation to keep the EOF guarantee of deployed code being always valid.

Places where it's needed:

- [ ] CLI EVM execution (`evmc run`?)
- [x] JSON state test loader or runner https://github.com/ethereum/evmone/pull/593

pdobacz:
Closes #585 

Not sure about placement and API. I assumed it should be off by default, and not concern `evmc` code, this is what I've got as a result.

### DIFF
--- a/lib/evmone/vm.cpp
+++ b/lib/evmone/vm.cpp
@@ -61,6 +61,11 @@ evmc_set_option_result set_option(evmc_vm* c_vm, char const* c_name, char const*
         vm.add_tracer(create_histogram_tracer(std::clog));
         return EVMC_SET_OPTION_SUCCESS;
     }
+    else if (name == "validate_eof")
+    {
+        vm.validate_eof = true;
+        return EVMC_SET_OPTION_SUCCESS;
+    }
     return EVMC_SET_OPTION_INVALID_NAME;
 }
 
--- a/README.md
+++ b/README.md
@@ -128,6 +128,14 @@ with it.
 docker run --entrypoint evmone-bench ethereum/evmone /src/test/benchmarks
 ```
 
+### EVM Object Format (EOF) support
+
+evmone supports EOFv1. Since EOF validation is done once during deploy-time, evmone does not revalidate during execution of bytecode. To force EOF revalidation, you can use the `validate_eof` option, example:
+
+```
+evmc run --vm libevmone.so,validate_eof --rev 13 "EF00"
+```
+
 ## References
 
 1. [Efficient gas calculation algorithm for EVM](docs/efficient_gas_calculation_algorithm.md)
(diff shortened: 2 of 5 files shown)


pdobacz:
I initially missed the possibilities to cover the new lines with tests, but codecov assisted me. It is now done. I'll squash the 2 commits before merge

chfast:
I think this is relatively good approach. Any other comments how useful it is in practice?

pdobacz:
I wish there was a natural way to have it "on" by default when running from the CLI, but there is no natural way to set it as default in `evmc run`, as it treats options opaquely.

One thought I got today is we can have the option "on" be default, but always **disable** it when running in a controlled environment, when we know revalidation isn't necessary. Setting the option would however be an overhead on each `vm.execute` and all the clients would need to be made aware of whether or not they're using `evmone` (e.g. `host.cpp`). Not good either...

On more thought, the `validate_eof=yes` option isn't any better or safer than requiring users to run `evmone-eofparse` beforehand. So maybe the best way to proceed here is just document EOF support with `evmone-eofparse`?

pdobacz:
Of course, ultimate solution is to add `evmc_validate_eof` to EVMC and fire it in `evmc run` whenever an EOF revision is targeted.

We have a bunch of options, WDYT?

lib/evmone/vm.cpp
```diff
@@ -61,6 +61,15 @@ evmc_set_option_result set_option(evmc_vm* c_vm, char const* c_name, char const*
         vm.add_tracer(create_histogram_tracer(std::clog));
         return EVMC_SET_OPTION_SUCCESS;
     }
+    else if (name == "validate_eof")
```
chfast, line 64: The current convention is to add the option without the `value` (see e.g. `trace` and `histogram`). So the usage is just `evmone.so,validate_eof`.

chfast:
> We have a bunch of options, WDYT?

I think EVMC is unnecessary overhead because there are no EVM implementations other than evmone. At some point we should  create a `evmone run` and drop tooling from EVMC.

lib/evmone/vm.cpp
pdobacz: fixed, I used cgoto as a template for a `validate_eof=off` version, then forgot to simplify.

pdobacz:
> > We have a bunch of options, WDYT?
> 
> I think EVMC is unnecessary overhead because there are no EVM implementations other than evmone. At some point we should create a `evmone run` and drop tooling from EVMC.

hold on then, if this is the long-term plan, it sounds like a case for the "won't fix" option, let's just document the `eofparse`.

chfast:
> hold on then, if this is the long-term plan, it sounds like a case for the "won't fix" option, let's just document the `eofparse`.

I think this is fine for now. Solidity is using EVMC/evmone so this may be handy for them.

PR #779, opened 2024-01-05
bytecode: Decode opcodes with immediate arguments

rodiazet:
- Support decoding all opcodes with imm arguments in `bytecode` library

### DIFF
--- a/test/utils/bytecode.hpp
+++ b/test/utils/bytecode.hpp
@@ -550,16 +550,15 @@ inline std::string decode(bytes_view bytecode)
         {
             s += std::string{" + OP_"} + name;
 
-            if (opcode >= OP_PUSH1 && opcode <= OP_PUSH32)
+            if (size_t imm_size = evmone::instr::traits[opcode].immediate_size; imm_size > 0)
             {
-                const auto push_data_start = it + 1;
-                const auto push_data_size =
-                    std::min(static_cast<std::size_t>(opcode - OP_PUSH1 + 1),
-                        static_cast<std::size_t>(bytecode.end() - push_data_start));
-                if (push_data_size != 0)
+                const auto imm_start = it + 1;
+                imm_size = std::min(imm_size, static_cast<size_t>(bytecode.end() - imm_start));
+
+                if (imm_size != 0)
                 {
-                    s += " + \"" + hex({&*push_data_start, push_data_size}) + '"';
-                    it += push_data_size;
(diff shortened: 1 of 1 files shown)


test/utils/bytecode.hpp
```diff
@@ -550,16 +550,17 @@ inline std::string decode(bytes_view bytecode)
         {
             s += std::string{" + OP_"} + name;
 
-            if (opcode >= OP_PUSH1 && opcode <= OP_PUSH32)
+            if (size_t imm_data_size = evmone::instr::traits[opcode].immediate_size;
```
chfast, line 553:
```suggestion
            if (const size_t imm_data_size = evmone::instr::traits[opcode].immediate_size;
```

test/utils/bytecode.hpp
```diff
@@ -550,16 +550,17 @@ inline std::string decode(bytes_view bytecode)
...
+                imm_data_size > 0)
             {
-                const auto push_data_start = it + 1;
-                const auto push_data_size =
-                    std::min(static_cast<std::size_t>(opcode - OP_PUSH1 + 1),
-                        static_cast<std::size_t>(bytecode.end() - push_data_start));
-                if (push_data_size != 0)
+                const auto imm_data_start = it + 1;
+                imm_data_size = std::min(static_cast<std::size_t>(imm_data_size),
```
chfast, line 557:
```suggestion
                imm_data_size = std::min(imm_data_size,
```

test/utils/bytecode.hpp
```diff
@@ -550,16 +550,17 @@ inline std::string decode(bytes_view bytecode)
...
+                    static_cast<std::size_t>(bytecode.end() - imm_data_start));
```
chfast, line 558:
```suggestion
                    static_cast<size_t>(bytecode.end() - imm_data_start));
```

test/utils/bytecode.hpp
chfast: I also think we could name it just `imm_size` unless it is already commonly named `imm_data` in other places.

chfast (changes requested):
Also add a test for e.g. `RJUMP`.

test/utils/bytecode.hpp
rodiazet: OK for naming. `imm_size` cannot be `const`. It's changed below.

chfast (approved):

PR #784, opened 2024-01-08
state: Move pieces of State implemetation

chfast:
- Group methods revertible by `rollback()` (only `touch()` for now). We later want to hide the journal from users by providing revertible state modification methods.
- Move `State` methods implementations to the `state.cpp` file. This hides the implementation and allows faster builds.
- Move types related to Journal implementation to the `class State` scope.

### DIFF
--- a/test/state/state.hpp
+++ b/test/state/state.hpp
@@ -104,6 +104,20 @@ public:
         return insert(addr, std::move(account));
     }
 
+    [[nodiscard]] auto& get_accounts() noexcept { return m_accounts; }
+
+    [[nodiscard]] const auto& get_accounts() const noexcept { return m_accounts; }
+
+    /// Returns the state journal checkpoint. It can be later used to in rollback()
+    /// to revert changes newer than the checkpoint.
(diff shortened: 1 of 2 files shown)


test/state/state.hpp
```diff
@@ -14,55 +14,55 @@
...
 {
-struct JournalBase
+/// The Ethereum State: the collection of accounts mapped by their addresses.
+///
+/// TODO: This class is copyable for testing. Consider making it non-copyable.
+class State
 {
-    address addr;
-};
+    struct JournalBase
```
rodiazet, line 22: Would be good to have some doc comments for base and deriving structs, but I can live w/o it.

rodiazet:
Can we also add a motivation of this change.

rodiazet (approved):

test/state/state.hpp
chfast: This is going to be updated later.

chfast:
> Can we also add a motivation of this change.

Added:

> We later want to hide the journal from users by providing revertible state modification methods.
> This hides the implementation and allows faster builds.

PR #786, opened 2024-01-09
Implement RETURNDATALOAD (EOF)

### LINKED ISSUE #775: Implement RETURNDATALOAD (as part of EOF)
Originally proposed in a PR to the EIP-7069 (https://github.com/ethereum/EIPs/pull/7820/files). Use [megaspec](https://github.com/ipsilon/eof/blob/main/spec/eof.md) as source of truth

pdobacz:
Closes #775 

Inclusion is still somewhat pending, but this is likely to be included in the EOF instruction set to avoid code size regressions reported by compiler teams. 

For this reason, I'm opening as draft, but this is reviewable. Does including a new opcode require any additional work?

### DIFF
--- a/lib/evmone/advanced_instructions.cpp
+++ b/lib/evmone/advanced_instructions.cpp
@@ -277,6 +277,7 @@ constexpr std::array<instruction_exec_fn, 256> instruction_implementations = [](
     table[OP_DATALOADN] = op_undefined;
     table[OP_DATASIZE] = op_undefined;
     table[OP_DATACOPY] = op_undefined;
+    table[OP_RETURNDATALOAD] = op_undefined;
     table[OP_JUMPF] = op_undefined;
 
     table[OP_DUPN] = op_undefined;
--- a/lib/evmone/baseline_instruction_table.cpp
+++ b/lib/evmone/baseline_instruction_table.cpp
@@ -36,6 +36,7 @@ constexpr auto legacy_cost_tables = []() noexcept {
     tables[EVMC_PRAGUE][OP_DATACOPY] = instr::undefined;
     tables[EVMC_PRAGUE][OP_DUPN] = instr::undefined;
     tables[EVMC_PRAGUE][OP_SWAPN] = instr::undefined;
+    tables[EVMC_PRAGUE][OP_RETURNDATALOAD] = instr::undefined;
     return tables;
 }();
 
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -563,6 +563,28 @@ inline Result extcodecopy(StackTop stack, int64_t gas_left, ExecutionState& stat
     return {EVMC_SUCCESS, gas_left};
 }
 
+inline Result returndataload(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    auto& index = stack.top();
+
+    if (state.return_data.size() < index)
+        return {EVMC_INVALID_MEMORY_ACCESS, gas_left};
+    else
+    {
+        const auto begin = static_cast<size_t>(index);
+        const auto end = begin + 32;
+        if (state.return_data.size() < end)
+            return {EVMC_INVALID_MEMORY_ACCESS, gas_left};
+
+        uint8_t data[32] = {};
+        for (size_t i = 0; i < (end - begin); ++i)
+            data[i] = state.return_data[begin + i];
+
+        index = intx::be::unsafe::load<uint256>(data);
+        return {EVMC_SUCCESS, gas_left};
+    }
+}
+
 inline void returndatasize(StackTop stack, ExecutionState& state) noexcept
 {
     stack.push(state.return_data.size());
--- a/test/unittests/evm_calls_test.cpp
+++ b/test/unittests/evm_calls_test.cpp
@@ -9,6 +9,9 @@
 using namespace evmc::literals;
 using evmone::test::evm;
 
+inline constexpr auto max_uint256 =
+    0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff_bytes32;
+
 TEST_P(evm, delegatecall)
 {
     auto code = bytecode{};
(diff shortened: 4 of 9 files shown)


lib/evmone/advanced_instructions.cpp
```diff
@@ -269,6 +269,7 @@ constexpr std::array<instruction_exec_fn, 256> instruction_implementations = [](
     table[OP_DATALOADN] = op_undefined;
     table[OP_DATASIZE] = op_undefined;
     table[OP_DATACOPY] = op_undefined;
+    table[OP_RETURNDATALOAD] = op_undefined;
```
gumb0, line 272: Also for should be fine to have it in Advanced.

lib/evmone/instructions.hpp
```diff
@@ -563,6 +563,23 @@ inline Result extcodecopy(StackTop stack, int64_t gas_left, ExecutionState& stat
     return {EVMC_SUCCESS, gas_left};
 }
 
+inline Result returndataload(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept
+{
+    auto& index = stack.top();
+
+    const auto begin = static_cast<size_t>(index);
```
gumb0, line 570: Need to check before that `index < retudn_data.size()`, see `dataload` and `calldataload` implementations.

Otherwise this cast just ignores if high bits are set in 256-bit word.

lib/evmone/baseline_instruction_table.cpp
```diff
@@ -34,6 +34,7 @@ constexpr auto legacy_cost_tables = []() noexcept {
     tables[EVMC_PRAGUE][OP_DATALOADN] = instr::undefined;
     tables[EVMC_PRAGUE][OP_DATASIZE] = instr::undefined;
     tables[EVMC_PRAGUE][OP_DATACOPY] = instr::undefined;
+    tables[EVMC_PRAGUE][OP_RETURNDATALOAD] = instr::undefined;
```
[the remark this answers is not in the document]
pdobacz: It could potentially, but we don't have it in a separate EIP, just alongside the revamped calls, which also are batched together with the rest of EOF. In this way RETURNDATALOAD is sort of "attached" to EOF.

Supporting this in legacy would mean we need to do the entire dance with backwards compatibility (a previously undefined instruction), correct? Maybe it is a pragmatic choice to only have it in EOF. I'm not sure what is the usual practice in such cases.

lib/evmone/advanced_instructions.cpp
pdobacz: But only if it is supported in legacy, right? If it isn't then I suppose it would be on the same boat as rest of eof opcodes. Let's handle legacy-or-not first.

lib/evmone/baseline_instruction_table.cpp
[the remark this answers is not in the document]
gumb0: Revamped CALLS are activated in legacy, too. 

The same reasoning applies here: RETURNDATALOAD does not require immediates, and can be useful for legacy, too. And it's even in the same EIP so far.

Usually not much backwards compatibility dancing is happening when introducing new instructions. From EIP:
>  Backwards Compatibility
>
>No existing instructions are changed and so we do not think any backwards compatibility issues can occur.

lib/evmone/instructions.hpp
```diff
@@ -563,6 +563,23 @@ inline Result extcodecopy(StackTop stack, int64_t gas_left, ExecutionState& stat
...
+    const auto end = begin + 32;
+    if (state.return_data.size() < end)
+        return {EVMC_INVALID_MEMORY_ACCESS, gas_left};
+
+    uint8_t data[32] = {};
+    for (size_t i = 0; i < (end - begin); ++i)
+        data[i] = state.return_data[begin + i];
+
+    index = intx::be::load<uint256>(data);
```
gumb0, line 579: more efficient
```suggestion
    index = intx::be::unsafe::load<uint256>(data);
```

test/unittests/evm_calls_test.cpp
```diff
@@ -741,6 +741,63 @@ TEST_P(evm, returndatacopy_outofrange)
...
+
+    execute(124, eof_bytecode(staticcall(0) + returndataload(3) + OP_STOP, 6));
+    EXPECT_EQ(result.status_code, EVMC_INVALID_MEMORY_ACCESS);
+
+    execute(124, eof_bytecode(staticcall(0) + returndataload(1) + OP_STOP, 6));
+    EXPECT_EQ(result.status_code, EVMC_SUCCESS);
+
+    execute(124, eof_bytecode(staticcall(0) + returndataload(2) + OP_STOP, 6));
+    EXPECT_EQ(result.status_code, EVMC_SUCCESS);
+}
```
gumb0, line 799: Some more nice to have tests:
- empty returndata, every offset expected to fail
- 31 byte return data, every offset expected to fail
- 32 byte return data, offsets 1, 31, 32, 0xff...ff expected to fail
- 34 byte return data - already done, can add 0xff...ff
- 64 byte return data, offsets 0, 16, 32 succeed, offstes 33, 0xff...ff expected to fail

lib/evmone/baseline_instruction_table.cpp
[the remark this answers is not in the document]
pdobacz: In the end it was decided to not activate revamped CALLS in legacy, so RETURNDATALOAD will go down the same path.

test/unittests/evm_calls_test.cpp
```diff
@@ -741,6 +741,182 @@ TEST_P(evm, returndatacopy_outofrange)
...
+        execute(124, eof_bytecode(staticcall(0) + returndataload(1) + OP_STOP, 6));
+        EXPECT_EQ(result.status_code, EVMC_INVALID_MEMORY_ACCESS);
+
+        execute(124, eof_bytecode(staticcall(0) + returndataload(31) + OP_STOP, 6));
+        EXPECT_EQ(result.status_code, EVMC_INVALID_MEMORY_ACCESS);
+
+        execute(124, eof_bytecode(staticcall(0) + returndataload(32) + OP_STOP, 6));
+        EXPECT_EQ(result.status_code, EVMC_INVALID_MEMORY_ACCESS);
+
+        const auto max_uint256 =
```
gumb0, line 822: this should be `constexpr`, and better declare it once in the test

test/unittests/evm_calls_test.cpp
```diff
@@ -741,6 +741,182 @@ TEST_P(evm, returndatacopy_outofrange)
...
+    if (is_advanced())
+        return;
+
+    rev = EVMC_PRAGUE;
+    {
+        const uint8_t call_output[31]{};
+        host.call_result.output_data = std::data(call_output);
+        host.call_result.output_size = std::size(call_output);
+
+        execute(124, eof_bytecode(staticcall(0) + returndataload(0) + OP_STOP, 6));
```
gumb0, line 805: Also I would suggest not to repeat `124` everywhere, just omit it, you tested it already in a separate test.

pdobacz: I took it from an existing `returndatacopy` case. But in the case of `returndataload` I suppose it is more ok to drop and simplify

PR #788, opened 2024-01-10
DUPN and SWAPN in EOF only + EOF stack validation

### LINKED ISSUE #538: Restrict DUPN and SWAPN to EOF only
The `DUPN` and `SWAPN` instructions implemented in #529 are enabled in Cancun in both EOF and legacy code. Limit the availability to EOF only.

pdobacz:
Closes #538 

Please take an extra look if the `invoke` specializations added are ok and have no undesired side effects, esp. the `advanced` counterpart. The compiler made me add the latter, even if it's not used.

### DIFF
--- a/lib/evmone/advanced_analysis.hpp
+++ b/lib/evmone/advanced_analysis.hpp
@@ -149,7 +149,7 @@ struct OpTableEntry
 {
     instruction_exec_fn fn;
     int16_t gas_cost;
-    int8_t stack_req;
+    uint8_t stack_req;
     int8_t stack_change;
 };
 
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -370,7 +370,7 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
 
         const auto opcode = static_cast<Opcode>(code[i]);
 
-        auto stack_height_required = instr::traits[opcode].stack_height_required;
+        int stack_height_required = instr::traits[opcode].stack_height_required;
         auto stack_height_change = instr::traits[opcode].stack_height_change;
 
         auto stack_height = stack_heights[i];
@@ -380,7 +380,7 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         {
             const auto fid = read_uint16_be(&code[i + 1]);
 
-            stack_height_required = static_cast<int8_t>(code_types[fid].inputs);
+            stack_height_required = code_types[fid].inputs;
 
             if (stack_height + code_types[fid].max_stack_height - stack_height_required >
                 STACK_SIZE_LIMIT)
@@ -401,26 +401,29 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
 
             if (code_types[fid].outputs == NON_RETURNING_FUNCITON)
             {
-                stack_height_required = static_cast<int8_t>(code_types[fid].inputs);
+                stack_height_required = code_types[fid].inputs;
             }
             else
             {
                 if (code_types[func_index].outputs < code_types[fid].outputs)
                     return EOFValidationError::jumpf_destination_incompatible_outputs;
 
-                stack_height_required =
-                    static_cast<int8_t>(code_types[func_index].outputs + code_types[fid].inputs -
-                                        code_types[fid].outputs);
+                stack_height_required = code_types[func_index].outputs + code_types[fid].inputs -
+                                        code_types[fid].outputs;
                 if (stack_heights[i] > stack_height_required)
                     return EOFValidationError::stack_higher_than_outputs_required;
             }
         }
         else if (opcode == OP_RETF)
         {
-            stack_height_required = static_cast<int8_t>(code_types[func_index].outputs);
+            stack_height_required = code_types[func_index].outputs;
             if (stack_height > code_types[func_index].outputs)
                 return EOFValidationError::stack_higher_than_outputs_required;
         }
+        else if (opcode == OP_DUPN)
+            stack_height_required = code[i + 1] + 1;
+        else if (opcode == OP_SWAPN)
+            stack_height_required = code[i + 1] + 2;
 
         if (stack_height < stack_height_required)
             return EOFValidationError::stack_underflow;
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -902,38 +902,16 @@ inline void swap(StackTop stack) noexcept
     a[3] = t3;
 }
 
-inline code_iterator dupn(StackTop stack, ExecutionState& state, code_iterator pos) noexcept
+inline code_iterator dupn(StackTop stack, code_iterator pos) noexcept
 {
-    const auto n = pos[1] + 1;
-
-    const auto stack_size = &stack.top() - state.stack_space.bottom();
-
-    if (stack_size < n)
-    {
-        state.status = EVMC_STACK_UNDERFLOW;
-        return nullptr;
-    }
-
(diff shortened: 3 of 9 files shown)


lib/evmone/eof.cpp
```diff
@@ -415,6 +415,16 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
             if (stack_height > code_types[func_index].outputs)
                 return EOFValidationError::stack_higher_than_outputs_required;
         }
+        else if (opcode == OP_DUPN)
+        {
+            const auto n = static_cast<int8_t>(code[i + 1]);
```
pdobacz, line 420: I think this isn't right, we should operate on uint8 here, but that requires changes to the entire function. I'll try things, but let me know if there's an easy way out.

chfast:
@rodiazet please review because some things overlap with #742

lib/evmone/eof.cpp
chfast: Yes, this seems wrong because it will probably overflow for values ≥ 128. Please add tests and fix. I think the fix may be to upgrade `auto stack_height_required` to `int stack_height_required`.

Similar issue may be for `static_cast<int8_t>(code_types[fid].inputs)`.

pdobacz: yea, I did this 2ceece90, does this look good?  I think we had tests for this (in this revision), that is, not tests explicitly testing validation but tests which ran those values >= 128, yet magically they've worked. I'll add more specific tests for this in the stack validation department.

Making a self not to see about `.inputs` part

lib/evmone/advanced_analysis.hpp
```diff
@@ -149,7 +149,7 @@ struct OpTableEntry
 {
     instruction_exec_fn fn;
     int16_t gas_cost;
-    int8_t stack_req;
+    u_int8_t stack_req;
```
chfast, line 152:
```suggestion
    uint8_t stack_req;
```

lib/evmone/eof.cpp
```diff
@@ -364,7 +364,8 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
 
         const auto opcode = static_cast<Opcode>(code[i]);
 
-        auto stack_height_required = instr::traits[opcode].stack_height_required;
+        auto stack_height_required =
```
chfast, line 367:
```suggestion
        int stack_height_required =
```
The arithmetic is done with `int` precision at least. This will avoid a lot of casts.

chfast (changes requested):

lib/evmone/eof.cpp
pdobacz: in 76ad9260 i did it in `uint8_t` and most of the casts are gone. I also think it should be enough for the `.inputs` case, inputs are 1 byte wide, and don't fill that space up as per current spec. Does this work?

chfast:
> The compiler made me add the latter, even if it's not used.

I think the compiler generates these functions anyway, they are just not assigned to the instruction table.

lib/evmone/eof.cpp
```diff
@@ -395,26 +395,35 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
         {
-            stack_height_required = static_cast<int8_t>(code_types[func_index].outputs);
+            stack_height_required = code_types[func_index].outputs;
             if (stack_height > code_types[func_index].outputs)
                 return EOFValidationError::stack_higher_than_outputs_required;
         }
+        else if (opcode == OP_DUPN)
+        {
+            const auto n = static_cast<uint8_t>(code[i + 1]);
+            stack_height_required = n + 1;
```
chfast, line 420: This will overflow to `0` if the imm is `0xff`.

chfast (changes requested):

lib/evmone/eof.cpp
pdobacz: let's discuss in other thread

lib/evmone/eof.cpp
```diff
@@ -395,26 +395,35 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
+        }
+        else if (opcode == OP_SWAPN)
+        {
+            const auto n = static_cast<uint8_t>(code[i + 1]);
+            stack_height_required = n + 2;
```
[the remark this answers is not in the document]
chfast: Ah, yes. This is because `n` is promoted to `int` in `n + 2`. So this is correct just the `static_cast` is unnecessary. How about `stack_height_required = code[i + 1] + 2`?

test/unittests/evm_eip663_dupn_swapn_test.cpp
```diff
@@ -86,23 +80,20 @@ TEST_P(evm, dupn_full_stack)
 
     rev = EVMC_PRAGUE;
     auto full_stack_code = bytecode{};
-    for (uint64_t i = 1023; i >= 1; --i)
+    for (uint64_t i = 1022; i >= 1; --i)
```
chfast, line 83: Why?

lib/evmone/eof.cpp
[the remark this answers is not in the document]
pdobacz: yup, this works, I thought I started with it and it didn't work :shrug:, maybe because of that `int8_t` up there. Now it works. Thanks for bearing with me here.

test/unittests/evm_eip663_dupn_swapn_test.cpp
pdobacz: oh yeah, this one. If we push 1023 times here, then the overall stack height goes up to 1024, which causes `max_stack_height_above_limit`, and we indeed have `t.max_stack_height > MAX_STACK_HEIGHT` and `constexpr auto MAX_STACK_HEIGHT = 0x03FF`, which is up to spec btw (megaspec).

The previous stack overflow check was at 1024.

A bit strange but correct I think? Or is there a problem with the spec?

chfast: Make sense: in EOF the stack limit in a function is 1023.

lib/evmone/instructions.hpp
```diff
@@ -902,35 +902,17 @@ inline void swap(StackTop stack) noexcept
     a[3] = t3;
 }
 
-inline code_iterator dupn(StackTop stack, ExecutionState& state, code_iterator pos) noexcept
+inline code_iterator dupn(StackTop stack, code_iterator pos) noexcept
 {
     const auto n = pos[1] + 1;
```
chfast, line 907: This can be simplified

chfast (changes requested):

chfast (commented):
The "Fix" is a bit misleading. Maybe change the commit title to

> Change the stack_height_required range to accommodate DUPN/SWAPN

chfast (commented):
A test for legacy code where the result is `EVMC_UNDEFINED_INSTRUCTION` is missing.

pdobacz:
> The "Fix" is a bit misleading. Maybe change the commit title to
> 
> > Change the stack_height_required range to accommodate DUPN/SWAPN

I'm planning to squash. The commit was split out for the sake of review back&forth

> A test for legacy code where the result is `EVMC_UNDEFINED_INSTRUCTION` is missing.

:+1:

chfast (approved):

pdobacz:
> @rodiazet please review because some things overlap with #742

@rodiazet @chfast Folks, I took a look at #742 myself and it seems both PRs add an `invoke` overload, a different one. I will assume that the conflict will be easily resolvable and go ahead to merge this.

PR #789, opened 2024-01-10
test: Reposition `validate_deployed_code()`

chfast:
Use `validate_deployed_code()` only to validate a test pre-state. Also check the blockchain tests.


chfast:
Obsolete by #795.

PR #795, opened 2024-01-15
test: Generalize `validate_state()`

chfast:
Generalize `validate_deployed_code()` to `validate_state()`. It validates an Ethereum state (likely a test pre-state). New added check: disallow zero-value storage entries.

### DIFF
--- a/test/statetest/statetest.hpp
+++ b/test/statetest/statetest.hpp
@@ -95,9 +95,11 @@ json::json to_json(const std::unordered_map<address, state::Account>& accounts);
 
 StateTransitionTest load_state_test(std::istream& input);
 
-/// Validates deployed EOF containers before running state test.
-/// Throws exception on any invalid EOF in state.
-void validate_deployed_code(const state::State& state, evmc_revision rev);
+/// Validates an Ethereum state:
+/// - checks that there are no zero-value storage entries,
+/// - checks that there are no invalid EOF codes.
+/// Throws std::invalid_argument exception.
+void validate_state(const state::State& state, evmc_revision rev);
 
 /// Execute the state @p test using the @p vm.
 ///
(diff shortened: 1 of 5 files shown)


test/statetest/statetest.hpp
```diff
@@ -95,9 +95,11 @@ json::json to_json(const std::unordered_map<address, state::Account>& accounts);
 
 StateTransitionTest load_state_test(std::istream& input);
 
-/// Validates deployed EOF containers before running state test.
-/// Throws exception on any invalid EOF in state.
-void validate_deployed_code(const state::State& state, evmc_revision rev);
+/// Validates an Ethereum state:
+/// - checks if there is no zero-value storage entries,
+/// - checks if there is no invalid EOF code.
```
gumb0, line 100:
```suggestion
/// - checks that there are no zero-value storage entries,
/// - checks that there is no invalid EOF code.
```

test/statetest/statetest_loader.cpp
```diff
@@ -271,9 +271,9 @@ state::State from_json<state::State>(const json::json& j)
         {
             for (const auto& [j_key, j_value] : storage_it->items())
             {
-                const auto value = from_json<bytes32>(j_value);
-                acc.storage.insert(
-                    {from_json<bytes32>(j_key), {.current = value, .original = value}});
+                if (const auto value = from_json<bytes32>(j_value); value)
```
gumb0, line 274: could better be for readability
```suggestion
                if (const auto value = from_json<bytes32>(j_value); !is_zero(value))
```

gumb0 (approved):

test/statetest/statetest_loader.cpp
chfast: How about this?

```suggestion
                if (const auto value = from_json<bytes32>(j_value))
```

gumb0: fine for me

Issue #752, opened 2023-12-19
evmone t8n report difficulty 1


winsvega:
http://retesteth.ethdevops.io/temp/evmone/evmone_EIP150_block1_0x9fdd68/

winsvega:
If I don't ask evmone to calculate difficulty and provide it in env i steadz then some random tests are failing

http://retesteth.ethdevops.io/results/log//2023-12-20-1703058480-evmone.txt

PR #803, opened 2024-01-31
test: Adjust difficulty if below minimum `0x20000`

rodiazet:
Applies the minimum difficulty cap also to pre-Byzantium difficulty
calculations.

Fixes https://github.com/ethereum/evmone/issues/752.

### DIFF
--- a/test/state/ethash_difficulty.cpp
+++ b/test/state/ethash_difficulty.cpp
@@ -51,23 +51,10 @@ int64_t calculate_difficulty_pre_byzantium(int64_t parent_difficulty, int64_t pa
     return diff;
 }
 
-}  // namespace
-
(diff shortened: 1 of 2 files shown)


test/state/ethash_difficulty.cpp
```diff
@@ -48,7 +48,8 @@ int64_t calculate_difficulty_pre_byzantium(int64_t parent_difficulty, int64_t pa
     else if (period_count == 2)
         diff += 1;
 
-    return diff;
+    // Adjust if below min difficulty `0x20000`
```
chfast, line 51: Add constant `MIN_DIFFICULTY` and reference to the YP.

chfast (changes requested):
A unit test is needed.

chfast (approved):

PR #805, opened 2024-02-01
test: Option to check state root hash

chfast:
Add option to check the expected post state MPT root hash of the state_transition tests.

### DIFF
--- a/test/unittests/state_transition.cpp
+++ b/test/unittests/state_transition.cpp
@@ -135,6 +135,11 @@ void state_transition::TearDown()
         }
     }
 
+    if (expect.state_hash)
+    {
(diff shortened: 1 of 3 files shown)


test/unittests/state_transition_trace_test.cpp
```diff
@@ -16,6 +16,9 @@ TEST_F(state_transition, trace_example)
     expect.status = EVMC_REVERT;
     expect.post[To].exists = true;
 
+    // Example of checking the expected the post state MPT root hash.
```
gumb0, line 19: I don't know why you put it into this test, it could be just in a separate test I think.

gumb0 (approved):

PR #807, opened 2024-02-02
test: Add some mising State Test export features

chfast:
Add support for exporting state transition tests:
- with legacy transactions (pre EIP-1559)
- with access list
- support EVMC spelling of revisions

### DIFF
--- a/test/unittests/state_transition.cpp
+++ b/test/unittests/state_transition.cpp
@@ -33,7 +33,20 @@ public:
 
 void state_transition::TearDown()
 {
+    // Validation:
+
+    if (rev < EVMC_LONDON)
+    {
+        ASSERT_EQ(tx.type, state::Transaction::Type::legacy);
+    }
+    if (tx.type == state::Transaction::Type::legacy)
+    {
+        ASSERT_EQ(tx.max_gas_price, tx.max_priority_gas_price);
+    }
+
     validate_state(pre, rev);
+
+    // Execution:
 
     auto state = pre;
     const auto trace = !expect.trace.empty();
@@ -155,8 +168,16 @@ void state_transition::export_state_test(const TransactionReceipt& receipt, cons
     jtx["sender"] = hex0x(tx.sender);
     jtx["secretKey"] = hex0x(SenderSecretKey);
     jtx["nonce"] = hex0x(tx.nonce);
-    jtx["maxFeePerGas"] = hex0x(tx.max_gas_price);
-    jtx["maxPriorityFeePerGas"] = hex0x(tx.max_priority_gas_price);
+    if (rev < EVMC_LONDON)
+    {
+        assert(tx.max_gas_price == tx.max_priority_gas_price);
+        jtx["gasPrice"] = hex0x(tx.max_gas_price);
+    }
+    else
+    {
+        jtx["maxFeePerGas"] = hex0x(tx.max_gas_price);
+        jtx["maxPriorityFeePerGas"] = hex0x(tx.max_priority_gas_price);
+    }
 
     jtx["data"][0] = hex0x(tx.data);
     jtx["gasLimit"][0] = hex0x(tx.gas_limit);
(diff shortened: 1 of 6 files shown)


test/unittests/state_transition.hpp
```diff
@@ -79,7 +79,7 @@ class state_transition : public ExportableFixture
     Transaction tx{
         .gas_limit = block.gas_limit,
         .max_gas_price = block.base_fee + 1,
-        .max_priority_gas_price = 1,
+        .max_priority_gas_price = block.base_fee + 1,
```
gumb0, line 82: Why does this change? This transaction field in the spec is called `max_priority_fee_per_gas` and is what is added on top of block base fee.

gumb0: Ah I guess because default is legacy transaction? It's confusing that `Transaction` has this field for legacy.

test/utils/utils.cpp
```diff
@@ -13,9 +13,9 @@ evmc_revision to_rev(std::string_view s)
         return EVMC_FRONTIER;
     if (s == "Homestead")
         return EVMC_HOMESTEAD;
-    if (s == "EIP150")
+    if (s == "Tangerine Whistle" || s == "EIP150")
```
chfast, line 16: This may not be the best idea because e.g. geth doesn't understand "Tangerine Whistle".

chfast: I changed this to export as `EIP150` and `EIP158`... We still accept full revision names.

test/unittests/state_transition.hpp
gumb0: Or another idea is that `Transaction` constructor for legacy should enforce this

chfast: Yes they can, but there is no time to do this change now.

test/unittests/state_transition_tx_test.cpp
```diff
@@ -65,3 +65,31 @@ TEST_F(state_transition, blob_tx_insuficient_funds)
     expect.post.at(Coinbase).exists = false;
     expect.status = EVMC_SUCCESS;
 }
+
+TEST_F(state_transition, empty_coinbase_fee_0_sd)
+{
+    rev = EVMC_SPURIOUS_DRAGON;
+
+    block_reward = 0;
+    block.base_fee = 0;
```
gumb0, line 75: Is this a valid block?

gumb0: I guess for old forks it is.

test/unittests/state_transition.cpp
```diff
@@ -197,6 +197,20 @@ void state_transition::export_state_test(const TransactionReceipt& receipt, cons
...
+    {
+        auto& ja = jtx["accessLists"][0];
+        for (const auto& [addr, storage_keys] : tx.access_list)
+        {
+            json::json je;
+            je["address"] = hex0x(addr);
+            auto& jstorage_keys = je["storageKeys"] = json::json::array();
+            for (const auto& k : storage_keys)
+                jstorage_keys.push_back(hex0x(k));
+            ja.push_back(je);
```
gumb0, line 210: Maybe makes sense to `std::move(je)`

test/unittests/state_transition.hpp
gumb0: But it also can be just
```
if (tx.type == Legacy)
  tx.max_priority_gas_price = tx.max_gas_price;
```

chfast: I improved the asserts in `TearDown()` a bit. 

It looks to me that we should have something like `TestBlock` with every field `std::optional` and set default values only after test definition. Otherwise, the test definitions must fight with defaults.

gumb0 (approved):

test/unittests/state_transition_tx_test.cpp
```diff
@@ -3,42 +3,52 @@
...
 
-    rev = EVMC_BERLIN;
-
     expect.tx_error = INSUFFICIENT_FUNDS;
 }
 
 TEST_F(state_transition, blob_tx_insuficient_funds)
 {
+    rev = EVMC_CANCUN;
+    block.base_fee = 1;
```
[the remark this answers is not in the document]
chfast: It is the minimal value by the spec.

PR #811, opened 2024-02-08
test: Introduce TestState and TestAccount

chfast:
This decouples the `state::State` for transaction execution from the state object for test definitions and asserts.

The `state::State` (also called "intra state") has some constructs and data related to transaction execution only (like "current" and "original" storage values). These should not be exposed to tests and make test definitions more complicated.

This separation should also help with adding new public API for EVM with transaction-level execution granularity.

### DIFF
--- a/test/state/test_state.hpp
+++ b/test/state/test_state.hpp
@@ -1,0 +1,67 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2024 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <evmc/evmc.hpp>
+#include <intx/intx.hpp>
+#include <map>
+
+namespace evmone
+{
+namespace state
+{
+class State;
+}
+
+namespace test
+{
+using evmc::address;
+using evmc::bytes;
+using evmc::bytes32;
+using intx::uint256;
+
+/// Ethereum account representation for tests.
+struct TestAccount
+{
+    uint64_t nonce = 0;
+    uint256 balance;
+    std::map<bytes32, bytes32> storage;
+    bytes code;
+
+    bool operator==(const TestAccount&) const noexcept = default;
+};
+
+/// Ethereum State representation for tests.
+///
+/// This is a simplified variant of state::State:
+/// it hides some details related to transaction execution (e.g. original storage values)
+/// and is also easier to work with in tests.
+class TestState : public std::map<address, TestAccount>
+{
+public:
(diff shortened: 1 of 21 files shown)


test/statetest/statetest.hpp
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "../state/state.hpp"
```
chfast, line 6: Still needed?

test/unittests/state_transition.cpp
```diff
@@ -99,46 +95,64 @@ void state_transition::TearDown()
...
+                ASSERT_NE(sit, acc.storage.end()) << addr << ": missing key " << key;
+                EXPECT_EQ(sit->second, value) << addr << ": key " << key;
             }
-            for (const auto& [key, value] : acc->storage)
+            else
             {
-                // Find unexpected storage keys. This will also report entries with value 0.
-                EXPECT_TRUE(expected_acc.storage.contains(key))
-                    << "unexpected storage key " << key << "=" << value.current << " in " << addr;
+                // FIXME:
```
pdobacz, line 136: is this fixme ok to be left behind? I'm staring at this and can't seem to figure it out, pls check

test/state/test_state.hpp
```diff
@@ -0,0 +1,48 @@
...
+{
+    uint64_t nonce = 0;
+    uint256 balance;
+    std::unordered_map<bytes32, bytes32> storage;
+    bytes code;
+
+    bool operator==(const TestAccount&) const noexcept = default;
+};
+
+class TestState : public std::unordered_map<address, TestAccount>
```
pdobacz, line 35: maybe we could summarize in a single sentence, how is it different from `State`? Just the gist of it. Since both are kept in the `test/` subdir, they are both, in a way, test-states, so it might be worth explaining when to expect which

pdobacz (approved):
2 items to look at, otherwise makes sense to me

test/unittests/state_transition.cpp
chfast: No. I need to figure it out. This code has been changed since the start of this PR. The new types makes the comparison easier (e.g. `TestState{} == TestState{}`) but we also want to have rich failure comments.

test/state/test_state.hpp
chfast: Added comments.

test/unittests/state_transition.cpp
chfast: Fixed. Mostly in #920.

PR #814, opened 2024-02-09
Move stack validation tests to separate file

gumb0:


chfast (approved):
Good to go assuming it only moves code.

Issue #783, opened 2024-01-08
Compute `CREATE` address without RLP library


chfast:
We want to decouple RLP library (mostly used in testing infrastructure and MPT hashing) from the core EVM state modifications. The `CREATE` address derivation is the only place where RLP is used. For this we only need to implement RLP encoding for `uint64`.

See the TODO comment at https://github.com/ethereum/evmone/blob/v0.11.0/test/state/host.cpp#L149.

AbhinavMir:
@chfast Does the draft PR above look right? I looked at how Aleth used to [encode JSON objects](https://github.com/ethereum/aleth/blob/master/rlp/main.cpp#L430) and bounced off from there to create a `rlp_encode_uint64` function.

elijahhampton:
Looks like the PR for this wasn't accepted and has gone stale. I can pick this up @chfast

chfast:
Ok, give it a try. For the clarification: any JSON should not be involved.

chfast:
I made a draft for it: #1055.

PR #815, opened 2024-02-11
Implement standalone RLP encoding for uint64 to decouple from EVM state

AbhinavMir:
potential fix https://github.com/ethereum/evmone/issues/783


chfast (changes requested):
No, this does not look good at all. This code doesn't even build.

The job is to copy `rlp::` utils here and trim it to what is needed for computing the create address. Which is: encode `uint64_t` number, encode the 20-byte address, and finally the RLP list of these two.

PR #816, opened 2024-02-12
Implement EIP-7610 (non-empty storage create collision) and upgrade execution-tests

chfast:
This implements [EIP-7610: Revert creation in case of non-empty storage](https://eips.ethereum.org/EIPS/eip-7610).
When a contract creation collides with an existing account it also reverts in case of the storage not being empty.

The ethereum/tests are upgraded to [v13.3](https://github.com/ethereum/tests/releases/tag/v13.3).

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -88,10 +88,19 @@ bytes_view extcode(bytes_view code) noexcept
 }
 
 /// Check if an existing account is the "create collision"
-/// as defined in the [EIP-684](https://eips.ethereum.org/EIPS/eip-684).
+/// as defined in the [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610).
 [[nodiscard]] bool is_create_collision(const Account& acc) noexcept
 {
-    return acc.nonce != 0 || !acc.code.empty();
+    if (acc.nonce != 0 || !acc.code.empty())
+        return true;
+
+    // acc.storage may have entries from access list, even if account storage is empty.
+    // Check for non-zero current values.
+    if (std::ranges::any_of(
+            acc.storage, [](auto& e) noexcept { return !is_zero(e.second.current); }))
+        return true;
+
+    return false;
 }
 }  // namespace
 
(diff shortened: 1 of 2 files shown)


test/state/host.cpp
```diff
@@ -268,14 +275,6 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
 
     new_acc->just_created = true;
 
-    // Clear the new account storage, but keep the access status (from tx access list).
-    // This is only needed for tests and cannot happen in real networks.
-    for (auto& [k, v] : new_acc->storage) [[unlikely]]
```
chfast, line 273: Because the storage is empty now we don't have to do anything with it. This is really nice because this eliminates the only case for storage traversal in EVM.

test/state/host.cpp
```diff
@@ -88,10 +88,17 @@ bytes_view extcode(bytes_view code) noexcept
...
-/// as defined in the [EIP-684](https://eips.ethereum.org/EIPS/eip-684).
+/// as defined in the [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610).
 [[nodiscard]] bool is_create_collision(const Account& acc) noexcept
 {
-    return acc.nonce != 0 || !acc.code.empty();
+    if (acc.nonce != 0 || !acc.code.empty())
+        return true;
+
+    // The storage may have access list entries so look for non-zero values.
+    if (std::ranges::any_of(acc.storage, [](auto& e) { return !is_zero(e.second.current); }))
```
gumb0, line 98:
```suggestion
    if (std::ranges::any_of(acc.storage, [](auto& e) noexcept { return !is_zero(e.second.current); }))
```

test/state/host.cpp
```diff
@@ -88,10 +88,17 @@ bytes_view extcode(bytes_view code) noexcept
...
 /// Check if an existing account is the "create collision"
-/// as defined in the [EIP-684](https://eips.ethereum.org/EIPS/eip-684).
+/// as defined in the [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610).
 [[nodiscard]] bool is_create_collision(const Account& acc) noexcept
 {
-    return acc.nonce != 0 || !acc.code.empty();
+    if (acc.nonce != 0 || !acc.code.empty())
+        return true;
+
+    // The storage may have access list entries so look for non-zero values.
```
gumb0, line 97: it was not very clear at first, I think you meant
```suggestion
    // acc.storage may have entries from access list, even if account storage is empty. Check for non-zero current values.
```

gumb0 (approved):

test/state/host.cpp
chfast: Not easy check because of the access list... so I'll pass.

PR #821, opened 2024-02-19
test: ExportableFixture for JSON tests exporting

chfast:
Extract utils for JSON tests exporting to abstract `ExportableFixture`.


test/unittests/state_transition.cpp
```diff
@@ -2,19 +2,13 @@
...
-
 #include "state_transition.hpp"
 #include <evmone/eof.hpp>
 #include <test/state/mpt_hash.hpp>
 #include <test/statetest/statetest.hpp>
 #include <filesystem>
 #include <fstream>
-
-namespace fs = std::filesystem;
+#include <regex>
```
gumb0, line 11: not needed I think

test/unittests/exportable_fixture.hpp
```diff
@@ -0,0 +1,18 @@
...
+#pragma once
+
+#include <gtest/gtest.h>
+
+namespace evmone::test
+{
+class ExportableFixture : public testing::Test
+{
+protected:
+    std::unique_ptr<std::ostream> export_out;
```
gumb0, line 13: This doesn't work, it has to be `std::unique_ptr<std::ofstream>` otherwise proper file stream destructor is not called and file is not written to disk.

gumb0 (changes requested):

test/unittests/exportable_fixture.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+namespace
+{
+/// Creates the file path for the exported test based on its name.
+fs::path get_export_test_path(const testing::TestInfo& test_info, std::string_view export_dir)
+{
+    const std::string suite_name{test_info.test_suite_name()};
+
+    const auto stem = fs::path{test_info.file()}.stem().string();
+    const std::regex re{suite_name + "_(.*)_test"};
```
gumb0, line 27: This does not work for `eof_validation` suite in `eof_stack_validation_test.cpp` file. But I guess it makes sense to rename the file.

test/unittests/exportable_fixture.hpp
chfast: What a hell? Am I doing something wrong?

test/unittests/exportable_fixture.cpp
chfast: Yes, let's rename it.

test/unittests/exportable_fixture.hpp
chfast: Still not sure why it doesn't work. But this is fixed differently now because we also want to avoid creating empty files. So the file creation is left up to the concrete test suite.

I also added a cmake test that verifies an example file was created during export.

test/unittests/exportable_fixture.cpp
gumb0: Can you rebase and rename in this PR?

test/integration/statetest/CMakeLists.txt
```diff
@@ -131,6 +131,15 @@ set_tests_properties(
     FIXTURES_SETUP EXPORT_JSON_TESTS
 )
 
+add_test(
+    NAME ${PREFIX}/export_to_json_created_files
+    COMMAND ${CMAKE_COMMAND} -E sha256sum ${EXPORT_DIR}/state_transition/block/known_block_hash.json
```
gumb0, line 136: this is just checking the existence of file? could be somethin like
```suggestion
    COMMAND ${CMAKE_COMMAND} -E test -f ${EXPORT_DIR}/state_transition/block/known_block_hash.json
```
?

gumb0 (approved):

test/unittests/exportable_fixture.hpp
gumb0: > Still not sure why it doesn't work

I'm also not sure anymore, becaue apparently destructor is indeed virtual :man_shrugging:

chfast: I was able to reproduce the same issue.

PR #824, opened 2024-02-26
state: Fix failing code deployment in Frontier

chfast:
In Frontier when CREATE fails to deploy the code because of the deployment cost the creation behaves as it has been successful although the created account has empty code.

Fix the CREATE instruction by returning the address of the created account instead of the null address.

The gas refund is also applied as in the case of successful code deployment. Geth agrees with this but this situation probably never happened on Mainnet.

### DIFF
--- a/test/unittests/state_transition_create_test.cpp
+++ b/test/unittests/state_transition_create_test.cpp
@@ -49,3 +49,109 @@ TEST_F(state_transition, create2_max_nonce)
 
     expect.post[*tx.to].nonce = pre.get(*tx.to).nonce;  // Nonce is unchanged.
 }
+
+TEST_F(state_transition, code_deployment_out_of_gas_tw)
+{
+    rev = EVMC_TANGERINE_WHISTLE;  // 63/64 gas rule enabled
+    block.base_fee = 0;
+    const auto initcode = ret(0, 5000);  // create contract with a lot of zeros, deploy cost 1M
+
+    tx.to = To;
+    tx.gas_limit = 1000000;
+    pre.insert(To, {.code = mstore(0, push(initcode)) +
+                            sstore(0, create().input(32 - initcode.size(), initcode.size()))});
+
+    expect.post[To].storage[0x00_bytes32] = 0x00_bytes32;
+}
+
+TEST_F(state_transition, code_deployment_out_of_gas_f)
+{
+    rev = EVMC_FRONTIER;
+    block.base_fee = 0;
+    const auto initcode = ret(0, 1000);  // create contract with a lot of zeros
+
+    tx.to = To;
+    tx.gas_limit = 100000;
+    pre.insert(To, {.code = mstore(0, push(initcode)) +
+                            sstore(0, create().input(32 - initcode.size(), initcode.size()))});
+
+    const auto created = compute_create_address(To, pre.get(To).nonce);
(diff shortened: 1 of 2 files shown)


test/unittests/state_transition_create_test.cpp
```diff
@@ -49,3 +49,115 @@ TEST_F(state_transition, create2_max_nonce)
 
     expect.post[*tx.to].nonce = pre.get(*tx.to).nonce;  // Nonce is unchanged.
 }
+
+TEST_F(state_transition, code_deployment_out_of_gas_tw)
+{
+    rev = EVMC_TANGERINE_WHISTLE;  // 63/64 gas rule enabled
+    block.base_fee = 0;
+    const auto initcode = ret(0, 5000);  // create contract will a lot of zeros, deploy cost 1M
```
gumb0, line 57:
```suggestion
    const auto initcode = ret(0, 5000);  // create contract with a lot of zeros, deploy cost 1M
```

test/unittests/state_transition_create_test.cpp
```diff
@@ -49,3 +49,115 @@ TEST_F(state_transition, create2_max_nonce)
...
+    tx.to = To;
+    tx.gas_limit = 100000;
+    pre.insert(To, {.code = mstore(0, push(initcode)) +
+                            sstore(0, create().input(32 - initcode.size(), initcode.size()))});
+
+    const auto created = compute_create_address(To, pre.get(To).nonce);
+    expect.post[created].code = bytes{};  // code deployment failure creates empty account
+    expect.post[created].nonce = 0;
+    bytes32 ccc{};
+    std::copy(std::begin(created.bytes), std::end(created.bytes), &ccc.bytes[12]);
```
gumb0, line 82: I made a helper for this in this commit, maybe makes sense to extract https://github.com/ethereum/evmone/pull/553/commits/1681575c9bdbb78d348b66722a63e1bd0a4431ee
(in utils.hpp)

test/unittests/state_transition_create_test.cpp
```diff
@@ -49,3 +49,115 @@ TEST_F(state_transition, create2_max_nonce)
...
+    tx.to = To;
+    tx.gas_limit = 100000;
+    pre.insert(To, {.code = mstore(0, push(initcode)) +
+                            sstore(0, create().input(32 - initcode.size(), initcode.size()))});
+
+    expect.post[To].exists = true;
+    const auto created = compute_create_address(To, pre.get(To).nonce);
+    expect.post[created].code = bytes{};  // code deployment failure creates empty account
+    expect.post[created].nonce = 0;
+    expect.post[created].storage[0x00_bytes32] = 0x01_bytes32;  // storage stays
```
gumb0, line 117: Frontier was crazy time

chfast: And this affects EIP-7610 nowadays :)

https://ethereum-magicians.org/t/eip-7610-revert-creation-in-case-of-non-empty-storage

test/unittests/state_transition_create_test.cpp
chfast: Extracted. If approved, I will commit the helper separately to have the bug-fix PR minimal.

gumb0 (approved):

PR #825, opened 2024-02-27
Disable EOF<->Legacy cross-creation

gumb0:
Pulled out of #553


circle.yml
```diff
@@ -535,7 +535,7 @@ jobs:
             ~/tests/EIPTests/BlockchainTests/
       - download_execution_tests:
           repo: ipsilon/tests
-          rev: eof-relaxed-stack-validation-20240221
+          rev: create3-not-supported
```
gumb0, line 538: I took the branch form https://github.com/ethereum/tests/pull/1337 and rolled back some changes there for supporting truncated data in containers. In this PR truncated data is not supported yet.

test/state/host.cpp
```diff
@@ -271,13 +271,22 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
...
+        if (m_rev >= EVMC_LONDON && m_rev < EVMC_PRAGUE)
+        {
+            // EIP-3541: Reject EF code.
             return evmc::Result{EVMC_CONTRACT_VALIDATION_FAILURE};
+        }
+        else if (m_rev >= EVMC_PRAGUE)
+        {
+            // Only CREATE3/4 is allowed to deploy code starting with EF.
+            // It must be valid EOF, which was validated before execution.
+            if (msg.kind == EVMC_CREATE || msg.kind == EVMC_CREATE2)
```
gumb0, line 285: It will be always true on current master, but anticipates support for CREATE3/4

chfast: Maybe use assert for now.

test/state/host.cpp
```diff
@@ -271,13 +271,22 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
...
-    if (m_rev >= EVMC_PRAGUE && (is_eof_container(initcode) || is_eof_container(code)))
+    if (!code.empty() && code[0] == 0xEF)
     {
-        if (validate_eof(m_rev, code) != EOFValidationError::success)
+        if (m_rev >= EVMC_LONDON && m_rev < EVMC_PRAGUE)
+        {
+            // EIP-3541: Reject EF code.
             return evmc::Result{EVMC_CONTRACT_VALIDATION_FAILURE};
+        }
+        else if (m_rev >= EVMC_PRAGUE)
```
chfast, line 281: If you reorder `m_rev` checks they should simplify.

test/unittests/state_transition_create_test.cpp
```diff
@@ -49,3 +49,74 @@ TEST_F(state_transition, create2_max_nonce)
 
     expect.post[*tx.to].nonce = pre.get(*tx.to).nonce;  // Nonce is unchanged.
 }
+
+TEST_F(state_transition, create_tx_deploying_eof)
+{
+    static constexpr auto create_address = 0x3442a1dec1e72f337007125aa67221498cdd759d_address;
```
chfast, line 111: You can use `compute_create_address()`. But also you can omit `expect.post[create_address].exists = false` because it is the default.

lib/evmone/instructions_calls.cpp
```diff
@@ -182,6 +182,10 @@ Result create_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noex
         // init_code_offset may be garbage if init_code_size == 0.
         msg.input_data = &state.memory[init_code_offset];
         msg.input_size = init_code_size;
+
+        // EOF initcode is not allowed for legacy creation
+        if (is_eof_container({msg.input_data, msg.input_size}))
```
chfast, line 187: I think this unnecessarily modifies current behavior. If the initcode starts with `EF` it fails during execution.

test/state/host.cpp
gumb0: Not sure what you mean. Wihout CREATE3/4 it's a failure in all forks, so we could just drop this commit from this PR and get to pre-EOF logic
```cpp
      if (m_rev >= EVMC_LONDON && !code.empty() && code[0] == 0xEF)  // Reject EF code.
        return evmc::Result{EVMC_CONTRACT_VALIDATION_FAILURE};
```
Then this commit will still be part of #553

test/unittests/state_transition_create_test.cpp
gumb0: I'll leave expectation for explicitness.

lib/evmone/instructions_calls.cpp
gumb0: Do you mean to change the spec? I think motivation for light failure was better UX and being aligned with legacy creation transaction being invalid with EOF.
https://github.com/ipsilon/eof/pull/53
cc @pdobacz

pdobacz: Yup, if we skip this, this is a change to the spec we produced (in both aspects as Andrei noted: CREATE and create tx, as they should remain aligned). I don't have a preference except that we settle the spec soon.

test/state/host.cpp
```diff
@@ -283,13 +283,22 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
...
```
chfast, line 293:
```suggestion
        if (m_rev >= EVMC_PRAGUE)
        else if (m_rev >= EVMC_LONDON)
```

PR #836, opened 2024-03-15
Move EOF creation transition tests to separate file

gumb0:


chfast (commented):
Fine, but I'd group them as `state_transition_eof_create_test.cpp`.

chfast (approved):

Issue #829, opened 2024-03-11
evmone t8n fails on blob tx


winsvega:
```
{
    "data" : "0x00",
    "gasLimit" : "0x3d0900",
    "nonce" : "0x00",
    "to" : "0x095e7baea6a6c7c4c2dfeb977efac326af552d87",
    "value" : "0x0186a0",
    "v" : "0x01",
    "r" : "0x4c6a25969991d996e7f40964ba872e8dc6e8acdc422af055d9f404180dd2746c",
    "s" : "0x13b66a111e61fb3ad575d456a1d976d221436f403f342ae8e78c19b177705607",
    "sender" : "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b",
    "chainId" : "0x01",
    "type" : "0x03",
    "maxFeePerGas" : "0x012a05f200",
    "maxPriorityFeePerGas" : "0x02",
    "accessList" : [
        {
            "address" : "0x095e7baea6a6c7c4c2dfeb977efac326af552d87",
            "storageKeys" : [
                "0x0000000000000000000000000000000000000000000000000000000000000000",
                "0x0000000000000000000000000000000000000000000000000000000000000001"
            ]
        }
    ],
    "maxFeePerBlobGas" : "0x0a",
    "blobVersionedHashes" : [
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0001",
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0002",
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0003",
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0004",
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0005",
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0006",
        "0x01a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065f0007"
    ]
}

```

make the tool crash with the error
```
Tool Error:
bad optional access
```

Expected behavior: 

Reject transaction with the error message as any other invalid transaction. calculating the post without tx, adding rejected transaction information

chfast:
I'm not able to reproduce this bug with the given JSON transaction. This transaction parses fine. The problem is likely in some other input.

winsvega:
[...]

PR #837, opened 2024-03-18
test: Improve error of invalid hash in JSON

chfast:
Throw more meaningful exception in case we cannot parse hex hash from JSON (instead of throwing "bad optional access").

Closes https://github.com/ethereum/evmone/issues/829.

### DIFF
--- a/test/statetest/statetest_loader.cpp
+++ b/test/statetest/statetest_loader.cpp
@@ -77,12 +77,14 @@ address from_json<address>(const json::json& j)
 template <>
 hash256 from_json<hash256>(const json::json& j)
 {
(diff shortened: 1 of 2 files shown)


winsvega:
No it stills crashes with the error message. the expected behaviour is to add this transaction to the list of rejected transactions with this error message. and continue to return outAlloc and out with the list of rejected transactions. 

because if a tool crashes it is like it got broken by the test.

chfast:
The JSON encoding of the transaction is invalid: the hash is expected to be 64-char hex string (not 63-char one).

winsvega:
ah sorry my bad then

PR #842, opened 2024-03-20
RLP-encode invalid transactions

### LINKED ISSUE #841: t8n don't load invalid JSON transaction
`evmone-t8n` complaints about

```
Tool Error:
`to` value for blob transaction must not be null
```

but the test is testing what happens if blob transaction will come as a contract creation. usually it goes as RLP but for evmone it is json.  so I think the t8n still should just add it to the rejection list.

_Originally posted by @winsvega in https://github.com/ethereum/evmone/issues/837#issuecomment-2003745595_

chfast:
In case a transaction has a field with invalid value we still should RLP-encode it (instead of throwing an exception). This is legit usage in testing tools like evmone-t8n.

Also add an integration test for `evmone-t8n` based on real input.

Fixes #841.

### DIFF
--- a/test/integration/t8n/CMakeLists.txt
+++ b/test/integration/t8n/CMakeLists.txt
@@ -27,7 +27,22 @@ add_test(
 set_tests_properties(${PREFIX}/${TEST_CASE} PROPERTIES FIXTURES_REQUIRED ${TEST_CASE})
 
 add_test(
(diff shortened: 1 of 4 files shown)


winsvega:
Yay, thats fix everything!

winsvega (approved):
Yay, thats fix everything!

PR #846, opened 2024-03-25
Add native implementation of RIPEMD160

chfast:
Add C++ implementation of the RIPEMD160 and use it for the precompile `0x03`.


lib/evmone_precompiles/ripemd160.cpp
```diff
@@ -0,0 +1,219 @@
...
+// libc++ before version 18 has incorrect std::rotl signature
+// https://github.com/llvm/llvm-project/commit/45500fa08acdf3849de9de470cdee5f4c8ee2f32
+#pragma clang diagnostic ignored "-Wsign-conversion"
+#endif
+
+namespace evmone::crypto
+{
+namespace
+{
+// TODO(C++23): Use std::byteswap.
```
battlmonstr, line 22: consider this:
https://github.com/erigontech/silkworm/blob/master/silkworm/db/snapshots/seg/compressor/huffman_code.cpp#L176

I've used `__cplusplus` instead of `TODO(C++23)` and it is possible to use intrinsics if you know `sizeof(T)`.

lib/evmone_precompiles/ripemd160.cpp
```diff
@@ -0,0 +1,219 @@
...
+        /* s′( 0..15) = */ 8, 9, 9, 11, 13, 15, 15, 5, 7, 7, 8, 11, 14, 14, 12, 6,  //
+        /* s′(16..31) = */ 9, 13, 15, 7, 12, 8, 9, 11, 7, 7, 12, 7, 6, 15, 13, 11,  //
+        /* s′(32..47) = */ 9, 7, 15, 11, 8, 6, 6, 14, 12, 13, 5, 14, 13, 13, 7, 5,  //
+        /* s′(48..63) = */ 15, 5, 8, 11, 14, 14, 6, 14, 6, 9, 12, 9, 12, 5, 15, 8,  //
+        /* s′(64..79) = */ 8, 5, 12, 9, 12, 5, 14, 6, 8, 13, 6, 5, 15, 13, 11, 11,  //
+    },
+};
+
+/// Converts native representation to/from little-endian.
+inline auto to_le(std::integral auto x) noexcept
```
battlmonstr, line 121: similar methods exist in `intx::le` namespace in case you are able to use intx.hpp here, see: https://github.com/erigontech/silkworm/blob/master/silkworm/core/common/endian.hpp

test/state/precompiles.cpp
```diff
@@ -272,7 +283,7 @@ inline constexpr auto traits = []() noexcept {
 #ifdef EVMONE_PRECOMPILES_SILKPRE
     // tbl[static_cast<size_t>(PrecompileId::ecrecover)].execute = silkpre_ecrecover_execute;
     tbl[static_cast<size_t>(PrecompileId::sha256)].execute = silkpre_sha256_execute;
-    tbl[static_cast<size_t>(PrecompileId::ripemd160)].execute = silkpre_ripemd160_execute;
+    // tbl[static_cast<size_t>(PrecompileId::ripemd160)].execute = silkpre_ripemd160_execute;
```
battlmonstr, line 286: why is this disabled?

battlmonstr (approved):

lib/evmone_precompiles/ripemd160.cpp
chfast: On one hand I plan to improve intx's API for this. On the other hand, I wanted to stick to C++ std only for this file.

test/state/precompiles.cpp
chfast: These override evmone's implementations of precompiles with silkpre. This does not make sense any more for RIPEMD160 by default as evmone just got the native implementation which is faster.

lib/evmone_precompiles/ripemd160.cpp
battlmonstr: I prefer to not keep TODOs in code if possible due to https://cwe.mitre.org/data/definitions/546

If you want to avoid people trying C++23, it is better to disable it in the global makelist, and in the error message explain how to disable it if they want to still use C++23 at their own risk (maybe with an extra flag).

Using `__cplusplus` feels simpler to my taste, because once we update our baseline, which is likely to happen, it will be easier to clean up the code.

test/state/precompiles.cpp
battlmonstr: maybe remove it completely instead of commenting?

chfast: This section gives the visualization what has been replaced so far. The goal is get rid of silkpre dependency.

lib/evmone_precompiles/ripemd160.cpp
chfast: The `TODO(C++23)` comments mean: after C++23 is enabled you can come here and simplify/improve the code. I actually did this for C++20 and this is not much work but can discover still missing features in compilers.

Maybe we can use different tag than TODO, but it won't solve the CWE issue.

PR #856, opened 2024-04-05
Add a test for the Creator Contract

pdobacz:
The Creator Contract is a hot topic rn, so here is a simple test of a manual, bytecode implementation I prepared. The implementation is as simple as possible, yet aligned with the pseudocode one from https://github.com/ipsilon/eof/blob/main/spec/eof.md#appendix-creator-contract

The bytecode implementation could be linked up from the megaspec if necessary, or copy-pasted into EIP-7620 - unless we decide to not have the Creator Contract in the end.


pdobacz:
Closing, as https://github.com/ipsilon/eof/pull/78 just got merged

PR #858, opened 2024-04-10
Export state tests with invalid txs

pdobacz:
Closes 

I'm branching off of eof-create4, as this is where a bunch of exportable invalid tx tests were added.

I also tackled a "// TODO: Compare states carefully, they should be identical." for invalid txs - the same state comparison logic is executed now. This required a tiny amount of shuffling in how the default expectations on accounts like sender and coinbase work, but I think this is acceptable


pdobacz:
Ready to review after rebase

test/unittests/state_transition.cpp
```diff
@@ -61,28 +60,37 @@ void state_transition::TearDown()
...
+        ASSERT_TRUE(holds_alternative<TransactionReceipt>(res))
+            << std::get<std::error_code>(res).message();
+        const auto& receipt = std::get<TransactionReceipt>(res);
 
-    ASSERT_TRUE(holds_alternative<TransactionReceipt>(res))
-        << std::get<std::error_code>(res).message();
-    const auto& receipt = std::get<TransactionReceipt>(res);
+        EXPECT_EQ(receipt.status, expect.status);
+        if (expect.gas_used.has_value())
+        {
```
rodiazet, line 86: We can omit brackets here

test/unittests/state_transition.cpp
```diff
@@ -61,28 +60,37 @@ void state_transition::TearDown()
     const auto res = state::transition(state, block, tx, rev, selected_vm, block.gas_limit,
         state::BlockInfo::MAX_BLOB_GAS_PER_BLOCK);
 
-    if (const auto expected_error = make_error_code(expect.tx_error))
+    const auto expected_error = make_error_code(expect.tx_error);
```
rodiazet, line 63: Why? It's never used outside of this `if` scope.

test/unittests/state_transition.cpp
```diff
@@ -92,12 +100,6 @@ void state_transition::TearDown()
         EXPECT_EQ(trace_capture->get_capture(), expect.trace);
     }
 
-    EXPECT_EQ(receipt.status, expect.status);
-    if (expect.gas_used.has_value())
-    {
-        EXPECT_EQ(receipt.gas_used, *expect.gas_used);
-    }
-
```
rodiazet, line 100: Can you explain why it's removed?

test/unittests/state_transition_tx_test.cpp
```diff
@@ -43,6 +43,7 @@ TEST_F(state_transition, invalid_tx_non_existing_sender)
     pre.get_accounts().erase(Sender);
 
     expect.tx_error = INSUFFICIENT_FUNDS;
+    expect.post.at(Sender).exists = false;
```
rodiazet, line 46: Can we use `[] operator` here also?

test/unittests/state_transition.cpp
pdobacz: some interim version leftovers

test/unittests/state_transition.cpp
pdobacz: not removed, just moved to the `else` clause above

rodiazet (approved):

test/unittests/state_transition.cpp
```diff
@@ -225,7 +227,16 @@ void state_transition::export_state_test(const TransactionReceipt& receipt, cons
     auto& jpost = jt["post"][to_test_fork_name(rev)][0];
     jpost["indexes"] = {{"data", 0}, {"gas", 0}, {"value", 0}};
     jpost["hash"] = hex0x(mpt_hash(post.get_accounts()));
-    jpost["logs"] = hex0x(logs_hash(receipt.logs));
+
+    if (holds_alternative<std::error_code>(res))
+    {
+        jpost["expectException"] = std::get<std::error_code>(res).message();
```
gumb0, line 233: These are not the right error messages that are normal in tests.
Example of valid one: https://github.com/ethereum/tests/blob/1c23e3c27ac53b794de0844d2d5e19cd2495b9d8/GeneralStateTests/stExample/invalidTr.json#L26

Perhaps some conversion needed similar to https://github.com/ethereum/evmone/blob/9b3ee17421d0f9ed2da555adef7ed81464566e7d/test/unittests/eof_validation.cpp#L14

The good place to find test error messages is probably retesteth config https://github.com/ethereum/retesteth/blob/fa5fe40627b0b4914295d5d0fe120dd0b053cd15/retesteth/configs/clientconfigs/evmone.cpp#L231

pdobacz: @gumb0 Thanks for flagging this! So to be sure, what I need to do is:
- PR `retesteth` to add the additional entries
- in this spot, call a new function like `get_tests_error_message(std::error_code)`, instead of just `.message()` which has clauses like:
```
            case INIT_CODE_EMPTY:
                return "TR_InitcodeEmpty";   // <<<<<- one of the new `TR_` entries
```
to translate the code into retesteth-compatible value?

gumb0: yes, sounds good

PR #866, opened 2024-04-16
Disallow code sections unreachable from section 0

hugo-dc:
- Validates the code by using a worklist of accessed code sections
- At the end if any code section was not accessed will return unreachable_code_sections error.
- Fixes tests previously containing unreachable code sections

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -14,9 +14,9 @@
 #include <limits>
 #include <numeric>
 #include <ostream>
+#include <queue>
 #include <span>
 #include <stack>
-#include <unordered_set>
 #include <variant>
 #include <vector>
 
@@ -251,7 +251,7 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
 
 EOFValidationError validate_instructions(evmc_revision rev, const EOF1Header& header,
     std::span<const EOF1Header> subcontainer_headers, size_t code_idx, bytes_view container,
-    std::unordered_set<uint16_t>& accessed_code_sections) noexcept
+    std::queue<uint16_t>& code_sections_worklist) noexcept
 {
     const bytes_view code{header.get_code(container, code_idx)};
     assert(!code.empty());  // guaranteed by EOF headers validation
@@ -284,7 +284,7 @@ EOFValidationError validate_instructions(evmc_revision rev, const EOF1Header& he
             if (header.types[fid].outputs == NON_RETURNING_FUNCTION)
                 return EOFValidationError::callf_to_non_returning_function;
             if (code_idx != fid)
-                accessed_code_sections.insert(fid);
+                code_sections_worklist.push(fid);
             i += 2;
         }
         else if (op == OP_RETF)
@@ -301,7 +301,7 @@ EOFValidationError validate_instructions(evmc_revision rev, const EOF1Header& he
             if (header.types[fid].outputs != NON_RETURNING_FUNCTION)
                 is_returning = true;
             if (code_idx != fid)
-                accessed_code_sections.insert(fid);
+                code_sections_worklist.push(fid);
             i += 2;
         }
         else if (op == OP_DATALOADN)
@@ -611,7 +611,8 @@ std::variant<EOF1Header, EOFValidationError> validate_eof1(  // NOLINT(misc-no-r
     }
     const auto data_offset = static_cast<uint16_t>(offset);
 
-    std::unordered_set<uint16_t> accessed_code_sections = {0};
+    std::vector<bool> visited_code_sections(code_sizes.size());
+    std::queue<uint16_t> code_sections_worklist({0});
     EOF1Header header{container[2], code_sizes, code_offsets, data_size, data_offset,
         container_sizes, container_offsets, types};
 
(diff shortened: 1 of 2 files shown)


lib/evmone/eof.cpp
```diff
@@ -13,6 +13,7 @@
 #include <limits>
 #include <numeric>
 #include <ostream>
+#include <queue>
 #include <span>
 #include <stack>
 #include <unordered_set>
```
rodiazet, line 19: Can be removed now.

lib/evmone/eof.cpp
```diff
@@ -566,14 +578,18 @@ std::variant<EOF1Header, EOFValidationError> validate_eof1(
 
         auto msh_or_error =
             validate_max_stack_height(header.get_code(container, code_idx), code_idx, header.types);
+
```
rodiazet, line 581: Unrelated change.

lib/evmone/eof.cpp
```diff
@@ -566,14 +578,18 @@ std::variant<EOF1Header, EOFValidationError> validate_eof1(
...
         if (const auto* error = std::get_if<EOFValidationError>(&msh_or_error))
             return *error;
         if (std::get<int32_t>(msh_or_error) != header.types[code_idx].max_stack_height)
             return EOFValidationError::invalid_max_stack_height;
     }
 
-    if (accessed_code_sections.size() != header.code_sizes.size())
-        return EOFValidationError::unreachable_code_sections;
+    for (auto&& visited_code_section : visited_code_sections)
```
rodiazet, line 588: Why not just `const auto&`?

gumb0: can be just `const auto`, because it's a simple type, but also could use `std::all_of` or `std::any_of` instead of the loop.

lib/evmone/eof.cpp
```diff
@@ -551,13 +551,25 @@ std::variant<EOF1Header, EOFValidationError> validate_eof1(
         offset += code_size;
     }
 
-    std::unordered_set<uint16_t> accessed_code_sections = {0};
+    std::vector<bool> visited_code_sections(code_sizes.size());
+    std::queue<uint16_t> code_sections_worklist;
+
+    code_sections_worklist.push(0);
```
gumb0, line 557:
```suggestion
    std::queue<uint16_t> code_sections_worklist{0};
```

test/unittests/eof_validation_test.cpp
```diff
@@ -629,8 +629,8 @@ TEST_F(eof_validation, max_stack_height)
                       0x400 * OP_POP + OP_STOP + OP_RETF,
         EOFValidationError::max_stack_height_above_limit);
 
-    add_test_case(bytecode{"EF0001 010008 02000200010C01 040000 00 00800000 000003FF"} + OP_STOP +
-                      0x400 * bytecode{1} + 0x400 * OP_POP + OP_RETF,
+    add_test_case(bytecode{"EF0001 010008 02000200040C01 040000 00 00800000 000003FF"} + OP_CALLF +
```
gumb0, line 632: can use `callf(1)`

test/unittests/eof_validation_test.cpp
```diff
@@ -909,5 +911,19 @@ TEST_F(eof_validation, unreachable_code_sections)
         // Code Section 254 calls itself instead of code section 255, leaving code section 255
         // unreachable
         add_test_case(code_sections_256_err_254, EOFValidationError::unreachable_code_sections);
+
+        // Code Section 0 calls section 1, which calls itself, leaving section
+        // 2 unreachable
+        add_test_case(eof_bytecode(OP_STOP).code(jumpf(1), 0, 0x80, 0).code(jumpf(2), 0, 0x80, 0),
```
gumb0, line 917: Section 0 doesn't call section 1 as comment above states

lib/evmone/eof.cpp
hugo-dc: I initially tried that but got this error:

```
error: no matching function for call to ‘std::queue<short unsigned int>::queue(<brace-enclose
d initializer list>)’                                                                                                                                                    
  555 |     std::queue<uint16_t> code_sections_worklist{0};
```

lib/evmone/eof.cpp
hugo-dc: Removed

lib/evmone/eof.cpp
hugo-dc: I changed it to avoid the loop, please take a look.

test/unittests/eof_validation_test.cpp
hugo-dc: It is now using `eof_bytecode`

test/unittests/eof_validation_test.cpp
hugo-dc: Fixed

lib/evmone/eof.cpp
```diff
@@ -551,13 +550,24 @@ std::variant<EOF1Header, EOFValidationError> validate_eof1(
...
-    std::unordered_set<uint16_t> accessed_code_sections = {0};
+    std::vector<bool> visited_code_sections(code_sizes.size());
+    std::queue<uint16_t> code_sections_worklist({0});
+
     EOF1Header header{container[2], code_sizes, code_offsets, data_size, types};
 
-    for (size_t code_idx = 0; code_idx < header.code_sizes.size(); ++code_idx)
+    while (!code_sections_worklist.empty())
     {
+        auto code_idx = code_sections_worklist.front();
```
gumb0, line 560:
```suggestion
        const auto code_idx = code_sections_worklist.front();
```

hugo-dc:
Rebased. 

Added a couple of tests for unreachable code sections in subcontainers.

PR #871, opened 2024-04-19
EOF subcontainer validation without recursion

gumb0:
Main part of the refactoring, that gets rid of recursion, is basically a transformation from:

- validate header
- for all subcontainers
  - validate subcontainer (recursion)
- validate instructions
- validate jumps, stack etc.

to:

- validate header
- queue = { {container, header} }
- while queue not empty
  - cont, header = queue.pop()
  - for all subcontainers
    - validate subcontainer header
    - queue.push( {subcontainer, subcontainer_header} )
  - validate instructions 
  - validate jumps, stack etc.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -623,64 +623,83 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
     return max_stack_height_it->max;
 }
 
-std::variant<EOF1Header, EOFValidationError> validate_eof1(  // NOLINT(misc-no-recursion)
-    evmc_revision rev, bytes_view container) noexcept
+std::variant<EOF1Header, EOFValidationError> validate_eof1(
+    evmc_revision rev, bytes_view main_container) noexcept
 {
-    const auto error_or_header = validate_header(rev, container);
+    const auto error_or_header = validate_header(rev, main_container);
     if (const auto* error = std::get_if<EOFValidationError>(&error_or_header))
         return *error;
-    const auto& header = std::get<EOF1Header>(error_or_header);
+    const auto& main_container_header = std::get<EOF1Header>(error_or_header);
 
-    std::vector<uint16_t> subcontainer_data_offsets;
-    std::vector<uint16_t> subcontainer_data_sizes;
-    for (size_t subcont_idx = 0; subcont_idx < header.container_sizes.size(); ++subcont_idx)
+    // Queue of (container, header) pairs left to process
+    std::queue<std::pair<bytes_view, EOF1Header>> container_queue;
+    container_queue.emplace(main_container, main_container_header);
+
+    while (!container_queue.empty())
     {
-        const bytes_view subcontainer{header.get_container(container, subcont_idx)};
+        // Validate subcontainer headers and get their data sections
+        const auto& [container, header] = container_queue.front();
+        std::vector<uint16_t> subcontainer_data_offsets;
+        std::vector<uint16_t> subcontainer_data_sizes;
+        for (size_t subcont_idx = 0; subcont_idx < header.container_sizes.size(); ++subcont_idx)
+        {
+            const bytes_view subcontainer{header.get_container(container, subcont_idx)};
 
-        auto error_subcont_or_header = validate_eof1(rev, subcontainer);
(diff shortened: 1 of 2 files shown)


pdobacz (approved):

lib/evmone/eof.cpp
```diff
@@ -623,53 +624,70 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
     return max_stack_height_it->max;
 }
 
-std::variant<EOF1Header, EOFValidationError> validate_eof1(  // NOLINT(misc-no-recursion)
-    evmc_revision rev, bytes_view container) noexcept
+std::variant<EOF1Header, EOFValidationError> validate_eof1(
+    evmc_revision rev, bytes_view cont) noexcept
```
rodiazet: If we change `container` to `cont` I would make it in the whole function. So also `subcontainer` to `subcont`

gumb0:
Changed the test to check theoretical limit (container section reaching 64K) instead of EIP-3860 limit.

Number of nesting levels are ~2000 for EIP-3860 an ~2600 for 64K.

lib/evmone/eof.cpp
gumb0: It was to disambiguate from `container` inside the loop, but a bad name, will change to `main_container`.

gumb0: It was to disambiguate from `container` inside the loop, but a bad name, will change to `main_container`.

lib/evmone/eof.cpp
```diff
@@ -249,6 +249,47 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
...
+    const auto& container_sizes = section_headers[CONTAINER_SECTION];
+    std::vector<uint16_t> container_offsets;
+    for (const auto container_size : container_sizes)
+    {
+        container_offsets.emplace_back(static_cast<uint16_t>(offset));
+        offset += container_size;
+    }
+    const auto data_offset = static_cast<uint16_t>(offset);
+
+    return EOF1Header{container[2], code_sizes, code_offsets, data_size, data_offset,
```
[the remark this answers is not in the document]
gumb0: Good idea, changed

lib/evmone/eof.cpp
```diff
@@ -249,8 +250,69 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
...
+{
+    if (!is_eof_container(container))
+        return EOFValidationError::invalid_prefix;
+
+    const auto version = get_eof_version(container);
+    if (version != 1)
+        return EOFValidationError::eof_version_unknown;
+
+    if (rev < EVMC_PRAGUE)
+        return EOFValidationError::eof_version_unknown;
```
rodiazet, line 264: Can we cover this also with a test?

rodiazet (approved):
Add unit test to cover mentioned fragment.

lib/evmone/eof.cpp
gumb0: We can, added one now.

test/unittests/eof_validation_test.cpp
```diff
@@ -1166,3 +1172,12 @@ TEST_F(eof_validation, EOF1_unreferenced_subcontainer_valid)
     const auto embedded = eof_bytecode(bytecode{OP_INVALID});
     add_test_case(eof_bytecode(OP_STOP).container(embedded), EOFValidationError::success);
 }
+
+TEST_F(eof_validation, max_nested_containers)
+{
+    bytecode code = eof_bytecode(OP_INVALID);
+    while (code.size() <= std::numeric_limits<uint16_t>::max())
```
hugo-dc, line 1179: It still can contain more nested containers, only limited by the maximum size of the main container.

For example: `while (code.size() <= std::numeric_limits<uint16_t>::max() + 9)` can still pass this test.

gumb0: Not sure what you mean, with `+9` it's the same nesting level (2621) and container size (65545)

Subcontainer size have to be `<= std::numeric_limits<uint16_t>::max()` because it's size is encoded as 2 bytes in parent header.

PR #876, opened 2024-04-29
EOF validation of subcontainer kinds

gumb0:
Implementing https://github.com/ipsilon/eof/pull/86

Depends on #888 

TODO:
- [x] "outside-in" refactoring - done, can be squashed with `Check that container is initcontainer in creation transaction/TXCREATE`
- [ ] Remove `ContainerKind::runtime_initcode` mode - after outside-in refactoring this requires* #916 so will be done there.
*if subcontainer is unreferenced, you cannot call its validation requiring neither runtime nor initcode mode.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -244,6 +244,17 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
     return types;
 }
 
+enum class ContainerKind : uint8_t
+{
+    /// Container that uses RETURNCONTRACT. Can be used by EOFCREATE/TXCREATE/Creation transaction.
+    initcode,
+    /// Container that uses STOP/RETURN. Can be returned by RETURNCONTRACT.
+    runtime,
+    /// Container that uses only REVERT/INVALID or does not terminate execution.
+    /// Can be used in any context.
+    initcode_runtime,
+};
+
 /// Result of validating instructions in a code section.
 struct InstructionValidationResult
 {
@@ -255,7 +266,8 @@ struct InstructionValidationResult
 };
 
 std::variant<InstructionValidationResult, EOFValidationError> validate_instructions(
-    evmc_revision rev, const EOF1Header& header, size_t code_idx, bytes_view container) noexcept
+    evmc_revision rev, const EOF1Header& header, ContainerKind kind, size_t code_idx,
+    bytes_view container) noexcept
 {
     const bytes_view code{header.get_code(container, code_idx)};
     assert(!code.empty());  // guaranteed by EOF headers validation
@@ -323,8 +335,19 @@ std::variant<InstructionValidationResult, EOFValidationError> validate_instructi
             if (container_idx >= header.container_sizes.size())
                 return EOFValidationError::invalid_container_section_index;
 
-            subcontainer_references.emplace_back(container_idx, Opcode{op});
+            if (op == OP_RETURNCONTRACT)
+            {
+                if (kind == ContainerKind::runtime || kind == ContainerKind::initcode_runtime)
+                    return EOFValidationError::incompatible_container_kind;
+            }
+
+            subcontainer_references.push_back({container_idx, Opcode{op}});
             ++i;
+        }
+        else if (op == OP_RETURN || op == OP_STOP)
+        {
+            if (kind == ContainerKind::initcode || kind == ContainerKind::initcode_runtime)
+                return EOFValidationError::incompatible_container_kind;
         }
         else
             i += instr::traits[op].immediate_size;
@@ -577,15 +600,16 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
     struct ContainerValidation
     {
         bytes_view bytes;
-        bool referenced_by_eofcreate = false;
+        ContainerKind kind;
     };
     // Queue of containers left to process
     std::queue<ContainerValidation> container_queue;
-    container_queue.push({main_container, false});
+
+    container_queue.push({main_container, ContainerKind::initcode_runtime});
 
     while (!container_queue.empty())
     {
-        const auto& [container, referenced_by_eofcreate] = container_queue.front();
+        const auto& [container, container_kind] = container_queue.front();
 
         // Validate header
         auto error_or_header = validate_header(rev, container);
@@ -603,6 +627,7 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
 
         const auto subcontainer_count = header.container_sizes.size();
         std::vector<bool> subcontainer_referenced_by_eofcreate(subcontainer_count, false);
+        std::vector<bool> subcontainer_referenced_by_returncontract(subcontainer_count, false);
 
         while (!code_sections_queue.empty())
         {
@@ -616,7 +641,7 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
 
             // Validate instructions
             const auto instr_validation_result_or_error =
-                validate_instructions(rev, header, code_idx, container);
+                validate_instructions(rev, header, container_kind, code_idx, container);
             if (const auto* error =
                     std::get_if<EOFValidationError>(&instr_validation_result_or_error))
                 return *error;
@@ -629,6 +654,10 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
             {
                 if (opcode == OP_EOFCREATE)
                     subcontainer_referenced_by_eofcreate[index] = true;
+                else if (opcode == OP_RETURNCONTRACT)
+                    subcontainer_referenced_by_returncontract[index] = true;
+                else
+                    intx::unreachable();
             }
 
             // TODO(C++23): can use push_range()
@@ -657,7 +686,8 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
         {
             if (main_container == container)
                 return EOFValidationError::toplevel_container_truncated;
-            if (referenced_by_eofcreate)
+            if (container_kind == ContainerKind::initcode ||
+                container_kind == ContainerKind::initcode_runtime)
                 return EOFValidationError::eofcreate_with_truncated_container;
         }
 
@@ -666,7 +696,18 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
         {
             const bytes_view subcontainer{header.get_container(container, subcont_idx)};
 
-            container_queue.push({subcontainer, subcontainer_referenced_by_eofcreate[subcont_idx]});
+            const bool eofcreate = subcontainer_referenced_by_eofcreate[subcont_idx];
+            const bool returncontract = subcontainer_referenced_by_returncontract[subcont_idx];
+
+            // TODO Validate whether subcontainer was referenced by any instruction
+
+            auto subcontainer_kind = ContainerKind::initcode_runtime;
+            if (!eofcreate)
+                subcontainer_kind = ContainerKind::runtime;
+            else if (!returncontract)
+                subcontainer_kind = ContainerKind::initcode;
+
+            container_queue.push({subcontainer, subcontainer_kind});
         }
 
         container_queue.pop();
@@ -935,6 +976,10 @@ std::string_view get_error_message(EOFValidationError err) noexcept
         return "eofcreate_with_truncated_container";
     case EOFValidationError::toplevel_container_truncated:
         return "toplevel_container_truncated";
+    case EOFValidationError::ambiguous_container_kind:
+        return "ambiguous_container_kind";
+    case EOFValidationError::incompatible_container_kind:
+        return "incompatible_container_kind";
     case EOFValidationError::impossible:
         return "impossible";
     }
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -244,17 +244,6 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
     return types;
 }
 
-enum class ContainerKind : uint8_t
-{
-    /// Container that uses RETURNCONTRACT. Can be used by EOFCREATE/TXCREATE/Creation transaction.
-    initcode,
-    /// Container that uses STOP/RETURN. Can be returned by RETURNCONTRACT.
-    runtime,
-    /// Container that uses only REVERT/INVALID or does not terminate execution.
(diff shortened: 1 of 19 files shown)


lib/evmone/eof.hpp
```diff
@@ -136,16 +137,25 @@ enum class EOFValidationError
...
+    initcode_runtime_container,
+};
+
 /// Determines the EOF version of the container by inspecting container's EOF prefix.
 /// If the prefix is missing or invalid, 0 is returned meaning legacy code.
 [[nodiscard]] uint8_t get_eof_version(bytes_view container) noexcept;
 
 /// Validates whether given container is a valid EOF according to the rules of given revision.
-[[nodiscard]] EVMC_EXPORT EOFValidationError validate_eof(
+[[nodiscard]] EVMC_EXPORT std::variant<EOFValidationError, ContainerKind> validate_eof(
```
gumb0, line 158: Not sure it's the best idea how to return ContainerKind from validation, and it leads to a big diff in tests etc.

gumb0: Another idea is to not touch this and add a variant like `validate_eof_and _get_container_kind` only for the cases where it's needed.

gumb0: Or optional pointer paramter to output `container_kind`, but it feels too C-like.

gumb0: I opted to adding a special function (with an ugly name), because it would be needed only in very special context of TXCREATE and creation transactions.

lib/evmone/eof.cpp
```diff
@@ -718,18 +749,35 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
...
+        {
+            if (referenced_by_returncontract)
+                return EOFValidationError::incompatible_container_kind;
+        }
+        else if (container_kind == ContainerKind::runtime_container)
+        {
+            if (referenced_by_eofcreate)
+                return EOFValidationError::incompatible_container_kind;
+        }
+
```
gumb0, line 764: Validation whether subcontainer is referenced by anything can be added here, or below in "enqueue subcontainers" loop, better in a separate PR.

test/unittests/eof_validation_test.cpp
```diff
@@ -1211,3 +1211,198 @@ TEST_F(eof_validation, max_nested_containers)
 
     add_test_case(code, EOFValidationError::success);
 }
+
+// Summary of validity of combinations of referencing instructions with instructions inside
+// referenced containers.
+// Rows are instructions referencing subcontainers or rules for top-level container.
+// Columns are instructions inside referenced subcontainer or none if it ends with REVERT/INVALID.
+//
+//    |                | STOP   | RETURN | RETURNCONTRACT | none |
```
pdobacz, line 1220: I presume you're omitting `TXCREATE` because it is getting removed? Otherwise would be same as `EOFCREATE`, correct?

Also, to anticipate the creation txs, I'm wondering whether it has same row as `EOFCREATE` or all terminating opcodes are allowed there? :thinking: A creation tx which deploys a bunch of contracts using `EOFCREATE`s, but doesn't `RETURNCONTRACT` itself seems valid. Or just runs some logic and doesn't even create any contract.

Actually... hooold on, what I'm saying is not according to spec of EIP-7698, but maybe sensible anyway? Especially, if we flip the requirement that creation tx top level is initcode-mode, and have it in runtime-mode, **requiring** an `EOFCREATE` inside to actually create something?

I need to look at my creation tx PR to ponder whether it's cleaner or not...

test/unittests/eof_validation_test.cpp
```diff
@@ -1211,3 +1211,198 @@ TEST_F(eof_validation, max_nested_containers)
 
     add_test_case(code, EOFValidationError::success);
 }
+
+// Summary of validity of combinations of referencing instructions with instructions inside
+// referenced containers.
+// Rows are instructions referencing subcontainers or rules for top-level container.
+// Columns are instructions inside referenced subcontainer or none if it ends with REVERT/INVALID.
```
pdobacz, line 1218: regardless of the topic below, I would be more strict on the terminology here: "top-level" or "ends" aren't well defined. An alternative is to remove "top-level" altogether and split the row into `account.code` and `InitcodeTransaction` or sth.
```suggestion
// Rows denote the way subcontainers containing the validated code are referenced by parent containers. The extra row for the top-level container means either one set for an account or found in the `initcodes` field of an `InitcodeTransaction`.
// Columns are terminating instructions used inside the validated code of such a container or none if code only terminates with REVERT/INVALID.
```

test/unittests/eof_validation_test.cpp
```diff
@@ -1211,3 +1211,198 @@ TEST_F(eof_validation, max_nested_containers)
...
+
+    add_test_case(container, EOFValidationError::success);
+}
+
+TEST_F(eof_validation, toplevel_container_returncontract)
+{
+    const auto code = returncontract(0, 0, 0);
+    const auto container = eof_bytecode(code, 2).container(eof_bytecode(OP_INVALID));
+
+    add_test_case(container, EOFValidationError::success);
```
pdobacz, line 1248: Shouldn't this case fail validation?

test/unittests/eof_validation_test.cpp
```diff
@@ -1211,3 +1211,198 @@ TEST_F(eof_validation, max_nested_containers)
...
+
+TEST_F(eof_validation, eofcreate_stop_and_returncontract)
+{
+    const auto runtime_container = eof_bytecode(OP_INVALID);
+    const auto init_code = rjumpi(1, 0) + OP_STOP + returncontract(0, 0, 0);
+    const auto init_container = eof_bytecode(init_code, 2).container(runtime_container);
+    const auto factory_code = eofcreate() + OP_STOP;
+    const bytecode factory_container = eof_bytecode(factory_code, 4).container(init_container);
+
+    add_test_case(factory_container, EOFValidationError::ambiguous_container_kind);
```
pdobacz, line 1396: I know this is WIP but to make sure: we need also cases for ambiguous container kind referenced by returncontract as well as top-level?

test/unittests/eof_validation_test.cpp
```diff
@@ -1211,3 +1211,198 @@ TEST_F(eof_validation, max_nested_containers)
...
+TEST_F(eof_validation, eofcreate_return_and_returncontract)
+{
+    const auto runtime_container = eof_bytecode(OP_INVALID);
+    const auto init_code = rjumpi(5, 0) + ret(0, 0) + returncontract(0, 0, 0);
+    const auto init_container = eof_bytecode(init_code, 2).container(runtime_container);
+    const auto factory_code = eofcreate() + OP_STOP;
+    const bytecode factory_container = eof_bytecode(factory_code, 4).container(init_container);
+
+    add_test_case(factory_container, EOFValidationError::ambiguous_container_kind);
+}
```
pdobacz, line 1408: I would also add a test where the offending terminating opcode is found in a non-first code section.

test/unittests/state_transition_eof_create_test.cpp
```diff
@@ -1162,7 +1062,7 @@ TEST_F(state_transition, txcreate_initcontainer_return)
     tx.to = To;
     pre.insert(*tx.to, {.nonce = 1, .code = factory_container});
 
-    expect.post[*tx.to].nonce = pre.get(*tx.to).nonce + 1;
+    expect.post[*tx.to].nonce = pre.get(*tx.to).nonce;
     expect.post[*tx.to].storage[0x00_bytes32] = 0x00_bytes32;
```
pdobacz, line 1066: this expectation is void. If we want to keep it, we should set the storage on tx.to to non-zero value.

Same comment next test

pdobacz (commented):
The refactored validation pipeline reads very well! The tests are a challenge indeed, but I guess we are more thorough now, than we were with the runtime tests.

test/unittests/eof_validation_test.cpp
gumb0: This looks to be not in the current spec https://github.com/ipsilon/eof/pull/86/files and probably it should be.

On the other hand there's no notion of "top-level" container, it's kinda something we have only in the tests and it's not clearly defined. Like, creation transaction is top-level container, but deployed contract is also top-level container...

gumb0: My hunch is that validation of top-level container, i.e. "general definition of validation" should allow either RETURNCONTRACT or STOP/RETURN. For creation transaction it's an extra rule that its toplevel container is initcontainer. For deployed containers (ones referenced by RETURNCONTRACT), it's a rule that they are runtime containers.

It's already specified like this I believe.

gumb0: New idea is to be explicit in validation tests whether we validate runtime container or initcontainer (extra flag for each case), then there will be no confusing "top-level container" tests.

gumb0: This may require another refactoring and I think we can get rid of `validate_eof_get_container_kind()` (we'll move toplevel container kind checks into `validate_eof()` and creation tx / eofcreate / txcreate will request to validate it as initcontainer.

pdobacz: Makes sense :+1:

test/unittests/eof_validation_test.cpp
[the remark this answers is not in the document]
gumb0: > Note that there also is one more option 😭 : there can be just an infinite loop with RJUMP at the end.

yeah that's why I called it "none". But I'll think more how to improve this, having a column per group seems to be not a bad idea to me.

lib/evmone/eof.cpp
gumb0: added

test/unittests/state_transition_eof_create_test.cpp
gumb0: It's redundant, it's just for documentation to make explicit what's the outcome.

test/unittests/eof_validation_test.cpp
gumb0: I rephrased and restructured the table a bit.

test/unittests/eof_validation_test.cpp
gumb0: Refactoring done, this case now succeeds in initcode mode.

test/unittests/eof_validation_test.cpp
gumb0: This check is done inside subcontainer validation, so doesn't really depend on what this container is referenced by.

test/unittests/eof_validation_test.cpp
gumb0: I also had in mind to add some tests with invalid combinations on a deeper nesting level.

test/eoftest/eoftest_runner.cpp
```diff
@@ -22,23 +22,39 @@ struct EOFValidationTest
...
     std::unordered_map<std::string, Case> cases;
 };
 
+ContainerKind container_kind_from_string(std::string_view s)
+{
+    if (s == "runtime")
+        return ContainerKind::runtime;
+    else if (s == "initcode")
+        return ContainerKind::initcode;
+    else if (s == "initcode_runtime")
```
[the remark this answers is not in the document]
gumb0: It is proposed to remove this from the spec and #916 has a commit that implements the removal.

gumb0:
> * Alternatively we remove the failing cases from `ethereum/tests` and port them to EEST.

I plan to do this.

test/eoftest/eoftest_runner.cpp
```diff
@@ -22,23 +22,39 @@ struct EOFValidationTest
...
+}
+
 void from_json(const json::json& j, EOFValidationTest::Case& o)
 {
     const auto op_code = evmc::from_hex(j.at("code").get<std::string>());
     if (!op_code)
         throw std::invalid_argument{"code is invalid hex string"};
     o.code = *op_code;
 
+    if (const auto it_kind = j.find("kind"); it_kind != j.end())
```
chfast, line 55: Is this standardized?

test/unittests/state_transition_eof_create_test.cpp
chfast: It is not exactly void, the check is only done if a value is assigned to "expect". There are `std::optional`s.

test/unittests/state_transition_eof_create_test.cpp
```diff
@@ -2089,7 +1989,7 @@ TEST_F(state_transition, txcreate_failure_after_txcreate_success)
         sstore(0, txcreate().initcode(keccak256(init_container)).salt(Salt)) +
         sstore(1, txcreate().initcode(keccak256(init_container)).salt(Salt)) +  // address collision
         sstore(2, returndatasize()) + sstore(3, 1) + OP_STOP;
-    const auto factory_container = eof_bytecode(factory_code, 5).container(init_container);
```
chfast, line 2092: What is this change about?

lib/evmone/eof.cpp
```diff
@@ -629,6 +644,10 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
             {
                 if (opcode == OP_EOFCREATE)
                     subcontainer_referenced_by_eofcreate[index] = true;
+                else if (opcode == OP_RETURNCONTRACT)
+                    subcontainer_referenced_by_returncontract[index] = true;
+                else
+                    assert(false);
```
chfast, line 650: use `intx::unreachable()`.

test/eoftest/eoftest_runner.cpp
gumb0: No, this is not supported anywhere, but I plan to try to add this to EEST.

test/unittests/state_transition_eof_create_test.cpp
gumb0: It was a bug in original test, so it was failing for the wrong reason I guess.

test/unittests/state_transition_eof_create_test.cpp
gumb0: So as I understand it is good to have it here.

lib/evmone/eof.cpp
gumb0: changed.

test/unittests/state_transition_eof_create_test.cpp
chfast: Yes.

chfast (approved):

pdobacz (approved):

PR #878, opened 2024-04-30
Allow EOF creation transactions

pdobacz:
First approach to allowing EOF creation transactions, thereby making the creator contract not necessary to deploy first EOF.

Follows the design from https://github.com/ipsilon/eof/pull/78 / https://github.com/ethereum/EIPs/pull/8498

However the validation problem has been resolved differently - it seemed easier to remove the validation rule to never have stray bytes in a container from the main `validation_eof` and to put it in the spots where it applies (TXCREATE initcontainer and subcontainers). This way this rule doesn't apply to creation transaction containers.

The validation state_tests are broken and should be reworked similar to the validation tests we have here, in light of the above change to validation.

The commit history reflects some interim steps, will be squashed before merging.


test/state/host.cpp
```diff
@@ -228,9 +228,40 @@ std::optional<evmc_message> Host::prepare_message(evmc_message msg)
             else
             {
                 assert(msg.kind == EVMC_EOFCREATE);
-                const bytes_view initcontainer{msg.code, msg.code_size};
-                msg.recipient =
-                    compute_eofcreate_address(msg.sender, msg.create2_salt, initcontainer);
+
+                // Indicator of an EOF creation tx - EVMC_EOFCREATE at depth 0.
```
pdobacz, line 232: I'm not a big fan of this. Alternative is to have a next msg.kind like EVMC_EOFCREATIONTX. For starters decided to leverage depth. WDYT?

gumb0: I think it's fine, there're a couple of places above in this same functions with this check, meaning the same.

gumb0:
So this kinda changes the definition of validation (at least for top-level container), which I'm not very sure is acceptable.

My other idea was something like:
- move stray data bytes check from `validate_section_headers()` to a level above to `validate_header()`
- make `validation_section_headers()` public
- call `validation_section_headers()` in `Host::prepare_message()` and sum up all section sizes to find the full container size and split to container/calldata
- validate container normally

pdobacz:
>     * validate container normally

Hm, you mean we redo all the previous steps (from `validation_section_headers`) in this last step?

gumb0:
> > * validate container normally
> 
> Hm, you mean we redo all the previous steps (from `validation_section_headers`) in this last step?

Yes, I would say it's fine. Or make another variant of `validate_eof()` which accepts already parsed section headers.

pdobacz:
The other approach is in 903dae2b. I like it, but it broadens eof.hpp significantly, not sure if this is a problem. I'll give it some more thought later. Or we can just proceed and plan a rewrite of these APIs later?

test/state/host.cpp
```diff
@@ -228,9 +228,45 @@ std::optional<evmc_message> Host::prepare_message(evmc_message msg)
...
+                if (msg.depth == 0)
+                {
+                    // Assert this is a legacy EOF creation tx.
+                    assert(is_eof_container(input));
+
+                    const auto section_headers_or_error = validate_section_headers(input);
+                    if (holds_alternative<EOFValidationError>(section_headers_or_error))
+                        return {};  // Light early exception.
+
+                    const auto header = read_valid_eof1_header(input);
```
gumb0, line 243: it feels hacky to me, we don't read to re-parse the header at this point, we already have all section sizes returned from `validate_section_headers()` and can sum them up to find the split.

test/state/host.cpp
```diff
@@ -228,9 +228,45 @@ std::optional<evmc_message> Host::prepare_message(evmc_message msg)
...
+                    // Follows from the header.can_init condition above.
+                    assert(container_size <= msg.input_size);
+
+                    msg.code = msg.input_data;
+                    msg.code_size = container_size;
+                    msg.input_data = msg.input_data + container_size;
+                    msg.input_size = msg.input_size - container_size;
+                    msg.recipient = compute_create_address(msg.sender, sender_nonce);
+
+                    if (validate_eof(m_rev, {msg.code, msg.code_size}) !=
```
gumb0, line 259: Validation should be done before calculating create address.

lib/evmone/eof.hpp
```diff
@@ -5,17 +5,29 @@
...
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
 #include <string>
+#include <variant>
 #include <vector>
 
 namespace evmone
 {
+constexpr uint8_t MAGIC[] = {0xef, 0x00};
```
gumb0, line 18: I didn't realize it will drag along all the constants into the header... What if instead of making `validation_section_headers()` public as is, we make a public wrapper of it that returns only full container size that we need? maybe like `std::variant<size_t, EOFValidationError> validate_eof_section_headers(bytes_view)`

gumb0: Also `MAGIC` doesn't have to be in the header in any case, does it?

test/state/host.cpp
pdobacz: I wanted to avoid the summing up, which we already handle in two places. BTW, did you see the comment https://github.com/ethereum/evmone/pull/878#issuecomment-2098375588, would this work in your opinion?

In general I would also prefer to take a safe and easy path now and possibly make a more thorough redesign of these validation routines, once we finalize the spec of the validations? There's still a lot of moving parts.

lib/evmone/eof.hpp
pdobacz: re `MAGIC` - yup, I grabbed too much. Re the fist comment, see my other comment and the idea to use `validate_header` instead.

pdobacz:
Okay, another version of. I figured that a new public wrapper of `validate_section_headers` would still not fully cut it, because we like to use the `header.can_init` in the creation transaction pipeline. But there was yet another possibility - to use `validate_header`, just moving the stray bytes check one level up (to validate eof1). It seems like a valid move, because the stray bytes check doesn't logically belong with validating the header... I suppose. This allows us to use that as the header parser in `host.cpp`. WDYT @gumb0 ?

This version needs to rely on #891 so that needs to merge first and then I force push the cherry-picked fix away.

test/unittests/state_transition_eof_create_test.cpp
```diff
@@ -925,6 +914,327 @@ TEST_F(state_transition, eofcreate_call_created_contract)
...
+}
+
+TEST_F(state_transition, creation_tx_auxdata_in_calldata)
+{
+    rev = EVMC_PRAGUE;
+    const auto deploy_data = "abcdef"_hex;
+    const auto aux_data = "aabbccddeeff"_hex;
+    const auto deploy_data_size = static_cast<uint16_t>(deploy_data.size());
+    const auto deploy_container =
+        eof_bytecode(bytecode(OP_INVALID)).data(deploy_data, deploy_data_size);
```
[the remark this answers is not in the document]
pdobacz: yes, but the `aux_data` name is confusing, I only now realized this, as I started figuring out. Should be `input` or something. This is the creation-tx-appended data, so not part of the data section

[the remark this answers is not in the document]
pdobacz: ah, correction - it is the data which becomes aux_data in RETURNCONTRACT, as such, it could have been included in the size. I think I'll add a new test to cover this as well. The difference is in one the aux data is static, in the other it is dynamic.

gumb0:
Looks good now, but there's weird clang-tidy warning now, not sure why.

test/state/host.cpp
```diff
@@ -256,7 +256,7 @@ std::optional<evmc_message> Host::prepare_message(evmc_message msg)
                     if (holds_alternative<EOFValidationError>(header_or_error))
                         return {};  // Light early exception.
 
-                    const auto& header = std::get<EOF1Header>(header_or_error);
+                    const auto header = std::get<EOF1Header>(header_or_error);
```
gumb0, line 259: So `std::get` is actually throwing, I guess that's why the warning. 

1. The reference shouldn't affect this I think, so would be good to report to clang-tidy if it's possible to reproduce on some minimal example
2. To work around we can use `get_if`
```cpp
                   const auto* header = std::get_ig<EOF1Header>(&header_or_error);
                    if (!header)
                        return {};  // Light early exception.
```
3. I think it makes sense to add `noexcept` to `prepare_message()`

pdobacz: But in `validate_eof1` a similar pattern is OK, which is what throws me off completely. Thanks for the suggestions, I'll poke around this

gumb0 (approved):
Looks good, some squashing before merge would be nice

PR #888, opened 2024-05-10
Refactor subcontainer validation

gumb0:
Refactoring pulled out of #876


lib/evmone/eof.cpp
```diff
@@ -309,17 +310,26 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
...
-    std::span<const uint16_t> subcontainer_data_offsets,
-    std::span<const uint16_t> subcontainer_data_sizes, size_t code_idx, bytes_view container,
-    std::queue<uint16_t>& code_sections_worklist) noexcept
+/// Result of validating instructions in a code section.
+struct InstructionValidationResult
+{
+    /// Pairs of (container_index, opcode) of all opcodes referencing subcontainers in this section.
+    std::vector<std::pair<uint8_t, Opcode>> subcontainer_references;
+    /// Set of accessed code section indices.
+    std::unordered_set<uint16_t> accessed_code_sections;
```
gumb0, line 319: this can be a vector, if we're cautios of sets

gumb0:
I think we could optimize out "list of references to subcontainers" if we rely more on non-const reference arguments to instruction validation, but I think this version looks cleaner, I would start with it at least.

lib/evmone/eof.cpp
```diff
@@ -633,42 +639,33 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
 {
-    const auto error_or_header = validate_header(rev, main_container);
-    if (const auto* error = std::get_if<EOFValidationError>(&error_or_header))
-        return *error;
-    const auto& main_container_header = std::get<EOF1Header>(error_or_header);
-
-    // Queue of (container, header) pairs left to process
-    std::queue<std::pair<bytes_view, EOF1Header>> container_queue;
-    container_queue.emplace(main_container, main_container_header);
+    struct Container
```
pdobacz, line 642: maybe a more specific name here, like "ValidatedContainer" or "ContainerValidationInfo"?

pdobacz (approved):
Looks great. I was wondering whether it impacts https://github.com/ethereum/evmone/pull/878 but it seems it doesn't. And you're correct it supersedes https://github.com/ethereum/evmone/pull/887

lib/evmone/eof.cpp
```diff
@@ -694,8 +708,16 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
                     std::get_if<EOFValidationError>(&instr_validation_result_or_error))
                 return *error;
 
-            const auto& [accessed_code_sections] =
+            const auto& [subcontainer_references, accessed_code_sections] =
                 std::get<InstructionValidationResult>(instr_validation_result_or_error);
+
+            // Mark what instructions referenced which subcontainers.
+            for (const auto& [index, opcode] : subcontainer_references)
```
chfast, line 690: Can be `std::ranges::any_of()`?

lib/evmone/eof.cpp
```diff
@@ -639,24 +639,25 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
...
     struct ContainerValidation
     {
         bytes_view bytes;
-        EOF1Header header;
         bool referenced_by_eofcreate = false;
     };
     // Queue of containers left to process
     std::queue<ContainerValidation> container_queue;
-    container_queue.emplace(main_container, main_container_header, false);
+    container_queue.push({main_container, false});
```
chfast, line 649: Why `emplace` → `push` is needed?

lib/evmone/eof.cpp
```diff
@@ -727,11 +715,24 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
...
+                return *error_subcont;
+
+            auto& subcont_header = std::get<EOF1Header>(error_subcont_or_header);
 
-        // enqueue subcontainers
-        for (auto& subcontainer : subcontainers)
-            container_queue.emplace(std::move(subcontainer));
+            container_queue.emplace(subcontainer, std::move(subcont_header),
+                subcontainer_referenced_by_eofcreate[subcont_idx],
+                subcontainer_referenced_by_returncontract[subcont_idx]);
```
chfast: This fields doesn't exist in `ContainerValidation`.

chfast (approved):

PR #889, opened 2024-05-13
Move TXCREATE to Osaka

gumb0:


chfast:
I meged the EVMC upgrade only in #897.

gumb0:
This is currently blocked on updating ethereum/tests

PR #890, opened 2024-05-14
Generate Copier files

hugo-dc:
Generates ethereum tests "Copier" files by executing `evmone-unittests` when an output path is defined in the `EVMONE_EXPORT_TESTS` env variable.

## Related PRs

- https://github.com/ethereum/tests/pull/1379
- https://github.com/ethereum/retesteth/pull/227

## Generated Copier Tests

This generates 2 directories containing the copier files:

- `eof_validation`
- `state_transition`

*Optionally, these directories can be symlinks to the corresponding directories in the ethereum tests:*

- `eof_validation -> <ethereum_tests_path>/src/EOFTestsFiller/efValidation/`
    -  `eof_validation/stack -> ../efStack`
- `state_transition -> <ethereum_tests_path>/src/EIPTestsFiller/StateTests/stEOF/`

This is the list of generated state tests generated (See [ethereum tests PR](https://github.com/ethereum/tests/pull/1379):

eof_validation:

- efStack (src Copier files already existed, Copier files were filled)
- efValidation

state_transition:

- ~~block~~
    - Skipped
        - A flag was added in order to skip generating tests not accepted by ethereum state because of using `withdrawals`, `BLOCKHASH`, ommer blocks, blob base fee.
- call
- create
- eip663
- eof_calls
- eof_create
    - OSAKA tests were skipped (currently not accepted by retesteth)
- extcode
- selfdestruct
- touch
- trace
- transient_storage

## Other changes

- Changed integration tests, previously it was looking for generated tests for the `block` state transition tests, but they are currently skipped, instead it is now checking for the `call` state tests.
- Allow hashes and bytes to be without leading zeros, which is valid for retesteth.
- Generate empty objects in the output json, when there are no objects in the tests, i.e.: `pre: {}`, `storage: {}`.
- Change `senderSecretKey`, retesteth does not accept secret keys to have leading zeroes.
    - This change required changing the expected created address/expected final hash depending on creator address.

### DIFF
--- a/test/unittests/state_transition_block_test.cpp
+++ b/test/unittests/state_transition_block_test.cpp
@@ -13,7 +13,6 @@ TEST_F(state_transition, block_apply_withdrawal)
     static constexpr auto withdrawal_address = 0x8888_address;
 
     block.withdrawals = {{0, 0, withdrawal_address, 3}};
-    block.timestamp = 1;
     tx.to = To;
     expect.post[withdrawal_address].balance = intx::uint256{3} * 1'000'000'000;
 }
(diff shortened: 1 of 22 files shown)


test/unittests/state_transition_block_test.cpp
```diff
@@ -13,6 +13,7 @@ TEST_F(state_transition, block_apply_withdrawal)
     static constexpr auto withdrawal_address = 0x8888_address;
 
     block.withdrawals = {{0, 0, withdrawal_address, 3}};
+    block.timestamp = 1;
```
gumb0, line 16: I think you can put this as a default to `state_transition.block` definition in state_transition.cpp

PR #891, opened 2024-05-15
Assert against overflow in header.data_offset

pdobacz:
While working on #878 I noticed the offset calculation in validate_header can overflow the uint16, if the _declared_ sizes of container sections are large. Currently there is no upper limit on container size in validation (there's only the indirect limit imposed by max_initcode_size), so there can be a large container, which has a subcontainer of size 0xffff, and will have it's `data_offset` not fit the uint16.

We have 2 tests which happened to run into this, but the problem was masked (the data_offset was just really tiny, despite the container had a huge subcontainer). I adjust them so (IMO) they still test what they're supposed to test, but not overflow.


lib/evmone/eof.cpp
```diff
@@ -296,6 +296,8 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
         container_offsets.emplace_back(static_cast<uint16_t>(offset));
         offset += container_size;
     }
+    // NOTE: assertion always satisfied only as long as initcode limits apply (48K).
+    assert(offset <= static_cast<size_t>(std::numeric_limits<uint16_t>::max()));
```
chfast, line 300: Do we need the cast to `size_t`? If yes, use `size_t{}`.

chfast (approved):

PR #893, opened 2024-05-20
Fix EOF initcode handling before EOF is enabled

chfast:
The special rule of handling EOF initcode in CREATE/CREATE2 instructions should not be activated before EOF is activated.

### DIFF
--- a/test/unittests/state_transition_eof_create_test.cpp
+++ b/test/unittests/state_transition_eof_create_test.cpp
@@ -31,7 +31,7 @@ TEST_F(state_transition, create_with_eof_initcode)
     tx.gas_limit = block.gas_limit;
     pre.get(tx.sender).balance = tx.gas_limit * tx.max_gas_price + tx.value + 1;
 
-    const bytecode init_container = eof_bytecode(ret(0, 1));
+    const bytecode init_container = eof_bytecode(OP_INVALID);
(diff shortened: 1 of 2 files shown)


test/unittests/state_transition_eof_create_test.cpp
```diff
@@ -69,6 +91,29 @@ TEST_F(state_transition, create2_with_eof_initcode)
...
 }
 
+TEST_F(state_transition, create2_with_eof_initcode_cancun)
+{
+    rev = EVMC_CANCUN;
+    block.gas_limit = 10'000'000;
+    tx.gas_limit = block.gas_limit;
+    pre.get(tx.sender).balance = tx.gas_limit * tx.max_gas_price + tx.value + 1;
+
+    const bytecode init_container = eof_bytecode(ret(0, 1));
```
pdobacz, line 101: we usually use `eof_bytecode(OP_INVALID)` as the smallest valid eof container. Also, `eof_bytecode(ret(0, 1))` fails validation due to wrong max stack height. However irrelevant to the logic being tested, I think it's safer to use a valid EOF code here.

Ah, I see the same is used above. I think this should be fixed there as well when we're at it.

pdobacz (approved):

holiman (approved):
LGTM

Issue #403, opened 2022-01-05
Implement stack space as uninitialized memory


chfast:
Currently the EVM stack space is provided by `uint256[256]`. This is wasteful because the whole space must be zero-initialized.
Moreover, we are reusing stack space for multiple executions therefore the stack space is dirty in some cases.

This can be implemented with `std::aligned_storage`. The [reference](https://en.cppreference.com/w/cpp/types/aligned_storage) has an example how to implement "static vector".

priteshvermaa:
Hi @chfast,  I would like to contribute on to this issue.

chfast:
Go ahead, @priteshvermaa.

chfast:
The `std::aligned_storage` has been deprecated.

axic:
@chfast so close this?

chfast:
I plan to change how stack space is allocated, so this is related and I can close it then.

PR #907, opened 2024-05-28
StackSpace: use std::aligned_alloc()

chfast:
Use `std::aligned_alloc()` (and `_aligned_malloc()` on MSVC) to allocate the memory for EVM stack.

This removes unnecessary initialization of the memory to zero (closes https://github.com/ethereum/evmone/issues/403).

The memory alignment is set to 32 bytes (as previously) but allows transitioning to "over-aligned" allocation for additional optimization
(see https://github.com/ethereum/evmone/issues/781).

### DIFF
--- a/lib/evmone/execution_state.hpp
+++ b/lib/evmone/execution_state.hpp
@@ -5,6 +5,7 @@
 
 #include <evmc/evmc.hpp>
 #include <intx/intx.hpp>
+#include <memory>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


lib/evmone/execution_state.hpp
```diff
@@ -27,20 +28,47 @@ using intx::uint256;
...
+            _aligned_free(p);
+#else
+            std::free(p);
+#endif
+        }
+    };
+
+    /// The storage allocated for maximum possible number of items.
+    /// Items are aligned to 256 bits for better packing in cache lines.
+    std::unique_ptr<uint256, Deleter> m_stack_space{allocate()};
```
battlmonstr, line 60: I'd put the `allocate()` call in a constructor.

lib/evmone/execution_state.hpp
```diff
@@ -27,20 +28,47 @@ using intx::uint256;
...
 public:
     /// The maximum number of EVM stack items.
     static constexpr auto limit = 1024;
 
+
     /// Returns the pointer to the "bottom", i.e. below the stack space.
     [[nodiscard, clang::no_sanitize("bounds")]] uint256* bottom() noexcept
     {
-        return m_stack_space - 1;
+        return m_stack_space.get() - 1;
```
battlmonstr, line 70: Why does it need to go outside the memory bounds?

battlmonstr (approved):

lib/evmone/execution_state.hpp
chfast: Added constructor.

battlmonstr: :+1:

lib/evmone/execution_state.hpp
```diff
@@ -27,20 +28,48 @@ using intx::uint256;
...
+
+    /// The storage allocated for maximum possible number of items.
+    /// Items are aligned to 256 bits for better packing in cache lines.
+    std::unique_ptr<uint256, Deleter> m_stack_space;
+
 public:
     /// The maximum number of EVM stack items.
     static constexpr auto limit = 1024;
 
+    StackSpace() noexcept : m_stack_space{allocate()} {}
```
gumb0, line 66: I wonder if 
```cpp
std::unique_ptr<uint256[]>(new (std::align_val_t(sizeof(uint256)) uint256[limit]);
``` 
achieves the same.

chfast: It looks like, however this specifies the array element alignment, so I wonder what happens if you specify 64 for uint256. I plan to align the allocation to `2048*32`.

PR #915, opened 2024-06-04
Make EXT*CALL ASE-ready

pdobacz:
Delivers https://github.com/ipsilon/eof/pull/107 / https://github.com/ethereum/EIPs/pull/8535

### DIFF
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -156,7 +156,7 @@ Result extcall_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noe
 {
     static_assert(Op == OP_EXTCALL || Op == OP_EXTDELEGATECALL || Op == OP_EXTSTATICCALL);
 
-    const auto dst = intx::be::trunc<evmc::address>(stack.pop());
+    const auto dst_u256 = stack.pop();
     const auto input_offset_u256 = stack.pop();
     const auto input_size_u256 = stack.pop();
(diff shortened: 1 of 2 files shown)


lib/evmone/instructions_calls.cpp
```diff
@@ -165,6 +165,12 @@ Result extcall_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noe
     stack.push(EXTCALL_ABORT);  // Assume (hard) failure.
     state.return_data.clear();
 
+    // Address space expansion ready check.
+    if (dst_u256 > (uint256{1} << 160) - 1)
+        return {EVMC_OUT_OF_GAS, gas_left};
```
pdobacz, line 170: hm, on 2nd thought it seems I should add a new `EVMC_` return code for this. Pls confirm if so @chfast

lib/evmone/instructions_calls.cpp
```diff
@@ -165,6 +165,12 @@ Result extcall_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noe
     stack.push(EXTCALL_ABORT);  // Assume (hard) failure.
     state.return_data.clear();
 
+    // Address space expansion ready check.
+    if (dst_u256 > (uint256{1} << 160) - 1)
```
chfast, line 169: Let's make it a named constant, e.g. `static constexpr auto ADDRESS_MAX = ...`;

lib/evmone/instructions_calls.cpp
chfast: Not strictly needed, but I think the `EVMC_ARGUMENT_OUT_OF_RANGE` is good fit for now.

chfast (approved):

PR #921, opened 2024-06-05
Disallow truncated data in toplevel containers

gumb0:
This change only affects what is allowed in tests. Truncated data sections are not allowed at top level now, to check truncated data rules, tests should put it into subcontainers.


chfast (approved):

gumb0:
This needs updating tests in ethereum/tests

cc @hugo-dc

PR #924, opened 2024-06-08
precompiles: Add SHA256 implementation

chfast:
This adds the SHA256 implementation from Silkworm and removes the precompile stub.


axic (approved):

chfast:
I tested this additionally on AMD Zen2 CPU with `sha_ni` extension.

rodiazet (approved):

PR #930, opened 2024-06-17
EOF: limit validated container size to MAX_INITCODE_SIZE

pdobacz:
https://github.com/ipsilon/eof/pull/125 / https://github.com/ethereum/EIPs/pull/8670

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -745,6 +745,12 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
     if (rev < EVMC_PRAGUE)
         return EOFValidationError::eof_version_unknown;
 
+    // `offset` variable handled below is known to not be greater than the container size, as
+    // checked in `validate_section_headers`. Combined with the requirement for the container
+    // size to not exceed MAX_INITCODE_SIZE (checked before `validate-header` is called),
+    // this allows us to cast `offset` to narrower integers.
+    assert(container.size() <= MAX_INITCODE_SIZE);
+
     const auto section_headers_or_error = validate_section_headers(container);
     if (const auto* error = std::get_if<EOFValidationError>(&section_headers_or_error))
         return *error;
@@ -765,11 +771,6 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
     const auto type_section_size = section_headers[TYPE_SECTION][0];
     auto offset = header_size + type_section_size;
 
-    // `offset` is being checked to not overflow the container size. NOTE: this is done
-    // implicitly in `validate_section_headers` above. Combined with the
-    // requirement for the container size to not exceed MAX_INITCODE_SIZE, this allows
-    // us to cast to a narrower integer.
-
     for (const auto code_size : code_sizes)
     {
         assert(offset <= std::numeric_limits<uint16_t>::max());
@@ -784,7 +785,6 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
         container_offsets.emplace_back(static_cast<uint16_t>(offset));
         offset += container_size;
     }
-
     const auto data_offset = static_cast<uint16_t>(offset);
 
     return EOF1Header{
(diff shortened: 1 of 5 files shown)


lib/evmone/eof.cpp
```diff
@@ -579,6 +582,10 @@ EOFValidationError validate_eof1(evmc_revision rev, bytes_view main_container) n
         bytes_view bytes;
         bool referenced_by_eofcreate = false;
     };
+
+    if (main_container.size() > MAX_INITCODE_SIZE)
```
pdobacz, line 586: An interesting case come up, concerning this piece: Should this check be done here (top-level `validate_eof`) or in `validate_header`.

An expectation would be to have the logic of `validate_header` protected against oversized containers. It has been since creation txs moved to the public API of eof.hpp, so, in principle, it should also contain the check as one of the first.

OTOH, this would be wasteful, as we'd be repeating the check for all subcontainers, which are always smaller than top-level. Also `validate_header` bare is only ever called in `host.cpp` during the process of validating a creation tx data (to be precise - to discover the split between container and calldata) - but there it is protected by the transaction-level MAX_INITCODE_SIZE check.

Let me know if you have any thoughts on this

pdobacz:
Converted to draft b/c testing something on the CI front, needs to remove the TEMP commits before merging!

lib/evmone/eof.cpp
```diff
@@ -771,8 +784,7 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
         container_offsets.emplace_back(static_cast<uint16_t>(offset));
         offset += container_size;
     }
-    // NOTE: assertion always satisfied only as long as initcode limits apply (48K).
-    assert(offset <= std::numeric_limits<uint16_t>::max());
```
chfast, line 775: I think we should keep the assert.

lib/evmone/eof.cpp
chfast: Does it mean the `validate_header` doesn't work correctly on its own?

pdobacz: > Does it mean the `validate_header` doesn't work correctly on its own?

Only guaranteed for containers which have its size under the limit. In other words - checking size is outside of `validate_header`'s responsibility.

chfast: If this is now the `validate_header()` precondition then add an assert in the `validate_header()`.

pdobacz:
The 2 failed EEST cases are already failing for the base 1.0.4 release + `evmone` master, and are related to the RETURNCONTRACT/EOFCREATE rules introduced elsewhere. No objections to merge as is and fix forward @chfast (I'm already working on this)?

lib/evmone/eof.cpp
gumb0: This looks still removed?

gumb0: I mean I would still keep it even if it's checked in `validate_section_headers()`

pdobacz: ah it moved up a bit and changed form, but the outcome is still the same (I think, please double check if in doubt) - the cast is safe.

gumb0: `assert(container.size() <= MAX_INITCODE_SIZE);` checks that container size is checked outside `validate_header()` and `validate_section_headers()` check that declared containers sizes sum doesn't exceed container size, so the cast is safe, but I would still keep the assertion here (to prevent future errors in case `validate_section_headers()` is changed).

Also currently it's incosistent, you removed this assertion, but kept another one at line 776.

pdobacz: > Also currently it's incosistent, you removed this assertion, but kept another one at line 776.

ouch, right, completely missed that.

OK on keeping the assertion, but I guess just one assertion at the end (the removed one) will be enough?

pdobacz: like 537d394a for example

gumb0: I would keep the one in `code_size` loop, too. If something goes wrong, it's easier to debug when it's detected early.

pdobacz: okay, then let's keep all, one for each cast to uint16 (one was missing originally)

gumb0 (approved):

PR #931, opened 2024-06-18
ci: Run ethereum/tests without optimizations (-Og)

chfast:


pdobacz (approved):
as discussued - let's have this now and possibly optimize later

PR #940, opened 2024-06-28
Extract `MAX_CODE_SIZE` constant

chfast:
Extract the `MAX_CODE_SIZE` and `MAX_INITCODE_SIZE` constants into common header.


chfast:
Some EIPs calls these "parameters", any opinions?

pdobacz (approved):
`constants` is ok, could also be `eip_params/parameters`. Just `parameters` is a bit vague.

chfast:
> `constants` is ok, could also be `eip_params/parameters`. Just `parameters` is a bit vague.

Ok, so let's keep the `constants` for now as I'm not sure this will be for EIP only.

PR #941, opened 2024-07-01
baseline: Improve code analysis API

chfast:
Passing the `evmc_revision` to the code analysis is too fine-grained for this public API. On Ethereum Mainnet EOF can be always enabled, but the options is left for other chain configurations.

### DIFF
--- a/lib/evmone/baseline.hpp
+++ b/lib/evmone/baseline.hpp
@@ -55,6 +55,9 @@ public:
     /// Reference to the EOF header.
     [[nodiscard]] const EOF1Header& eof_header() const noexcept { return m_eof_header; }
 
+    /// Reference to the EOF data section. May be empty.
+    [[nodiscard]] bytes_view eof_data() const noexcept { return m_eof_header.get_data(m_raw_code); }
+
     /// Check if given position is valid jump destination. Use only for legacy code.
     [[nodiscard]] bool check_jumpdest(uint64_t position) const noexcept
     {
--- a/lib/evmone/baseline_execution.cpp
+++ b/lib/evmone/baseline_execution.cpp
@@ -349,8 +349,7 @@ evmc_result execute(evmc_vm* c_vm, const evmc_host_interface* host, evmc_host_co
     }
 
     const auto code_analysis = analyze(container, eof_enabled);
-    const auto data = code_analysis.eof_header().get_data(container);
-    auto state = std::make_unique<ExecutionState>(*msg, rev, *host, ctx, container, data);
+    auto state = std::make_unique<ExecutionState>(*msg, rev, *host, ctx, container);
     return execute(*vm, msg->gas, *state, code_analysis);
 }
 }  // namespace evmone::baseline
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -985,38 +985,40 @@ inline Result mcopy(StackTop stack, int64_t gas_left, ExecutionState& state) noe
 
 inline void dataload(StackTop stack, ExecutionState& state) noexcept
 {
+    const auto data = state.analysis.baseline->eof_data();
     auto& index = stack.top();
 
-    if (state.data.size() < index)
+    if (data.size() < index)
         index = 0;
     else
     {
         const auto begin = static_cast<size_t>(index);
-        const auto end = std::min(begin + 32, state.data.size());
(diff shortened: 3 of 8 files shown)


lib/evmone/baseline.hpp
```diff
@@ -43,8 +43,14 @@ class CodeAnalysis
     {}
 };
 
-/// Analyze the code to build the bitmap of valid JUMPDEST locations.
-EVMC_EXPORT CodeAnalysis analyze(evmc_revision rev, bytes_view code);
+/// Analyze the EVM code in preparation for execution.
+///
+/// For legacy code this builds the map of valid JUMPDESTs.
+/// If EOF is enabled it recognized the EOF code by the code prefix.
```
gumb0, line 49:
```suggestion
/// If EOF is enabled, it recognizes the EOF code by the code prefix.
```

gumb0 (approved):

chfast:
Converted to draft, because I have better idea...

lib/evmone/baseline.hpp
```diff
@@ -22,29 +22,51 @@ class CodeAnalysis
...
+        return m_jumpdest_map[position];
+    }
 };
 
-/// Analyze the code to build the bitmap of valid JUMPDEST locations.
-EVMC_EXPORT CodeAnalysis analyze(evmc_revision rev, bytes_view code);
+/// Analyze the EVM code in preparation for execution.
+///
+/// For legacy code this builds the map of valid JUMPDESTs.
+/// If EOF is enabled, it recognized the EOF code by the code prefix.
```
pdobacz, line 65:
```suggestion
/// If EOF is enabled, it recognizes the EOF code by the code prefix.
```

lib/evmone/instructions.hpp
```diff
@@ -741,14 +741,14 @@ Result sstore(StackTop stack, int64_t gas_left, ExecutionState& state) noexcept;
 /// Internal jump implementation for JUMP/JUMPI instructions.
 inline code_iterator jump_impl(ExecutionState& state, const uint256& dst) noexcept
 {
-    const auto& jumpdest_map = state.analysis.baseline->jumpdest_map;
-    if (dst >= jumpdest_map.size() || !jumpdest_map[static_cast<size_t>(dst)])
+    if (dst[3] != 0 || dst[2] != 0 || dst[1] != 0 ||
```
pdobacz, line 744: Please add a comment on what it does with these !=0s

pdobacz (approved):
LGTM, only minor nits

lib/evmone/instructions.hpp
chfast: good now?

chfast:
I changed the name for the "full unmodified code" to `raw_code`. Any better ideas?

- `raw_code`
- `account_code`
- `original_code`
- `full_code`
- `container_code`

pdobacz:
> I changed the name for the "full unmodified code" to `raw_code`. Any better ideas?
> 
>     * `raw_code`
> 
>     * `account_code`
> 
>     * `original_code`
> 
>     * `full_code`
> 
>     * `container_code`

`code_bytes` or `code_binary`? All the `_code` versions imply it's code, while it is a something containing code. But `raw_code` was ok when I was reading through, just a moment of hesitation and the presence of `executable_code` nearby. Maybe also just `container` or `container_bytes/binary`

lib/evmone/baseline_execution.cpp
```diff
@@ -335,19 +335,21 @@ evmc_result execute(evmc_vm* c_vm, const evmc_host_interface* host, evmc_host_co
 {
     auto vm = static_cast<VM*>(c_vm);
     const bytes_view container{code, code_size};
+    const auto eof_enabled = rev >= EVMC_PRAGUE;
```
gumb0, line 338: Can you use `REV_EOF1` here?

test/unittests/baseline_analysis_test.cpp
chfast, line 1: I need one test for legacy.

gumb0 (approved):

lib/evmone/baseline_execution.cpp
chfast: Added as `instr::REV_EOF1` but revision handling can be improved further. I wonder if we can make the revision opaque for EVMC.

PR #942, opened 2024-07-01
ci: Upgrade ethereum/tests to v14.0

chfast:


circle.yml
```diff
@@ -437,13 +437,14 @@ jobs:
     steps:
       - build
       - download_execution_tests:
-          rev: v13.3
+          rev: v14.0
```
winsvega, line 440: Legacy true?

winsvega (approved):
Yes from now on we are migrating to pyspecs release process. Meaning new tests will go to .py

But all existing tests are not yet managed by .py so it's in legacy.

The develop branch of the test repo will still cover all tests on cancun. 
Slowly removing those ones that converted

circle.yml
chfast: `true` by default.

chfast (commented):
I re-added Cancun/BlockchainTests/ValidBlock in hope of restoring coverage to what it was.

chfast:
> I re-added Cancun/BlockchainTests/ValidBlock in hope of restoring coverage to what it was.

Coverage restored, so I guess this is good to go.

PR #943, opened 2024-07-03
eoftest: Add support for "containerKind"

chfast:
Add support for `"containerKind"` in the EOF validation JSON tests.
Specified in https://github.com/ethereum/execution-spec-tests/pull/651.


chfast:
We may want to add some unit tests for this format.

pdobacz (approved):
LGTM, I cherry picked this to combine with bumping EEST release to 1.0.5 to see if everything works.

re unit test - since we have none for eoftest at the time, let's maybe punt this to an issue?

chfast:
I simplified it to handle only what we need.

pdobacz (approved):
Also good like this :+1:

PR #944, opened 2024-07-03
 ci: EOF execution-spec-tests 1.0.5 

pdobacz:
~~Cherrypicked containerKind support from #943~~

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -401,17 +401,13 @@ jobs:
     steps:
       - build
       - download_execution_spec_tests:
-          repo: ipsilon/execution-spec-tests
-          release: eip7692@v1.0.4+ipsilon_bf37249b
+          release: eip7692@v1.0.5
           fixtures_suffix: eip7692
       - run:
           name: "EOF pre-release execution spec tests (state_tests)"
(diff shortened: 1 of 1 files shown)


test/eoftest/eoftest_runner.cpp
```diff
@@ -41,8 +41,20 @@ void from_json(const json::json& j, EOFValidationTest::Case& o)
...
     o.code = *op_code;
 
-    if (const auto it_initcode = j.find("initcode"); it_initcode != j.end())
+    if (const auto it_kind = j.find("containerKind"); it_kind != j.end())
+    {
+        const auto v = it_kind->get<std::string>();
+        if (v == "INITCODE")
+            o.kind = ContainerKind::initcode;
+        else if (v == "RUNTIME")
+            o.kind = ContainerKind::runtime;
```
pdobacz, line 50: RUNTIME is default and never explicitly listed it seems. Up to a unittest to cover this, I'd add a separate backburner issue for this and the unknown case too.

chfast (approved):
Looks great, just update the title.

PR #951, opened 2024-07-11
precompiles: Improve output buffer handling

chfast:
Replace the fixed-size on-stack temporary buffer for precompiles with a heap-allocated buffer big enough to handle properly any precompile invocation.

This actually keeps the number of allocations the same. Previously the contents of the on-stack buffer were copied to heap by the Result constructor. Now we are creating the heap buffer in the first place and pass the ownership of it to the Result.

This fixes out-of-bounds memory accesses often being found by fuzzers.

### DIFF
--- a/test/state/precompiles.cpp
+++ b/test/state/precompiles.cpp
@@ -364,19 +364,14 @@ evmc::Result call_precompile(evmc_revision rev, const evmc_message& msg) noexcep
     if (gas_left < 0)
         return evmc::Result{EVMC_OUT_OF_GAS};
 
-    // Buffer for the precompile's output.
-    // Big enough to handle all "expmod" tests, but in case does not match the size requirement
-    // from analysis, the result will likely be incorrect.
-    // TODO: Replace with std::pmr::monotonic_buffer_resource?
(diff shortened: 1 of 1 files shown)


test/state/precompiles.cpp
```diff
@@ -364,19 +364,16 @@ evmc::Result call_precompile(evmc_revision rev, const evmc_message& msg) noexcep
...
         return evmc::Result{EVMC_OUT_OF_GAS};
 
-    // Buffer for the precompile's output.
-    // Big enough to handle all "expmod" tests, but in case does not match the size requirement
-    // from analysis, the result will likely be incorrect.
-    // TODO: Replace with std::pmr::monotonic_buffer_resource?
-    uint8_t output_buf[4096];
-    assert(std::size(output_buf) >= max_output_size);
-
+    const auto output_buf = static_cast<uint8_t*>(std::malloc(max_output_size));
```
gumb0, line 367: Wouldn't `new uint8_t[max_output_size]` be better, at least in that it doesn't require a cast?

gumb0 (approved):

PR #957, opened 2024-07-23
Fix parsing EOF version in header

chfast:
Fixes the bug introduced in https://github.com/ethereum/evmone/pull/947 where reading the version byte is missing the bounds check.

This bug is hard to detect when `std::string` is used as the test container. Found by a fuzzer.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -892,7 +892,7 @@ bool append_data_section(bytes& container, bytes_view aux_data)
 
 uint8_t get_eof_version(bytes_view container) noexcept
 {
-    return is_eof_container(container) ? container[2] : 0;
+    return (is_eof_container(container) && container.size() >= 3) ? container[2] : 0;
 }
 
 EOFValidationError validate_eof(
--- a/test/unittests/eof_validation.cpp
+++ b/test/unittests/eof_validation.cpp
@@ -110,9 +110,17 @@ void eof_validation::TearDown()
     for (size_t i = 0; i < test_cases.size(); ++i)
     {
         const auto& test_case = test_cases[i];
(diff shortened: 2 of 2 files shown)


gumb0:
Can we have a test for this?

chfast:
> Can we have a test for this?

In theory we have: https://github.com/ethereum/evmone/blob/master/test/unittests/eof_validation_test.cpp#L38

pdobacz:
I'd love to know more what's going on here (or have a test or some other guarantee this never materializes). How did the fuzzer manage to run into this?

chfast:
> I'd love to know more what's going on here (or have a test or some other guarantee this never materializes). How did the fuzzer manage to run into this?

I think I know how to fix the tests so I will submit it separately. When you use vector or string as the container the out-of-bounds read is not always detected because of the container's additional capacity: this memory is actually allocated and it is merely the out-of-bounds by the vector/string logic. In the string case, there is always the terminating null byte allocated in the container. Normally, address sanitizer misses this unless the container has special annotations for address sanitizer. Availability of these annotations depends on the stdlib vendor and version. We may not have it in CI. See https://github.com/google/sanitizers/wiki/AddressSanitizerContainerOverflow.

pdobacz (approved):
Thanks for the explanation, this is pure magic... Though how did the fuzzer run into this?

chfast:
> Though how did the fuzzer run into this?

It allocates the input buffer of exact size.

chfast:
Do not merge yet.

chfast:
This is the older PR where we use libc++ with address sanitizer. There you can see some container-overflows reported in the `evmone-statetest` tool but not in the EOF unit tests...

chfast:
This helps: https://github.com/ethereum/evmone/pull/958.

PR #958, opened 2024-07-24
test: Improve out-of-buffer detection in EOF tests

chfast:
In EOF validation tests, move the container to new heap-allocated buffer of exact size to easily find out-of-buffer reads with address sanitizer.

### DIFF
--- a/test/unittests/eof_validation.cpp
+++ b/test/unittests/eof_validation.cpp
@@ -110,9 +110,17 @@ void eof_validation::TearDown()
     for (size_t i = 0; i < test_cases.size(); ++i)
     {
         const auto& test_case = test_cases[i];
-        EXPECT_EQ(evmone::validate_eof(rev, test_case.kind, test_case.container), test_case.error)
+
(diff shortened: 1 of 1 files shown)


pdobacz (commented):
:+1:

test/unittests/eof_validation.cpp
```diff
@@ -110,9 +110,17 @@ void eof_validation::TearDown()
     for (size_t i = 0; i < test_cases.size(); ++i)
     {
         const auto& test_case = test_cases[i];
-        EXPECT_EQ(evmone::validate_eof(rev, test_case.kind, test_case.container), test_case.error)
+
+        // Move the container to new heap-allocated buffer of exact size
```
gumb0, line 114: Wouldn't `test_case.container.shrink_to_fit()` achieve the same?

chfast: Not for `std::string` where there is SSO and additional NULL terminator.

PR #961, opened 2024-07-25
WIP EIP-7702 implementation

gumb0:


gumb0:
GitGuardian is warning is clearly a false positive, but also it will go away after devnet3 with this change https://github.com/ethereum/EIPs/pull/8845

lib/evmone/instructions_calls.cpp
```diff
@@ -79,9 +85,13 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
 
     evmc_message msg{.kind = to_call_kind(Op)};
     msg.flags = (Op == OP_STATICCALL) ? uint32_t{EVMC_STATIC} : state.msg->flags;
+    if (dst != code_addr)
+        msg.flags |= EVMC_DELEGATED;
+    else
+        msg.flags &= ~std::underlying_type_t<evmc_flags>{EVMC_DELEGATED};
```
gumb0, line 91: This hack point to that maybe `msg.flags` is not the best way to signal delegation (because it is not propagated unlike static mode).

Alternative could be a separate delegation address field in `evmc_message`, or another flag field.

test/state/state.cpp
```diff
@@ -12,10 +12,15 @@
 #include <evmone/execution_state.hpp>
 #include <algorithm>
 
+using namespace intx;
+
 namespace evmone::state
 {
 namespace
 {
+constexpr auto SECP256K1N = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141_u256;
```
gumb0, line 21: "GitGuardian Security Checks" flag this as a leaked secred, which is obviosly false positive, but also will go away with the next spec change https://github.com/ethereum/eips/pull/8845

lib/evmone/instructions.hpp
```diff
@@ -515,6 +565,7 @@ inline void blobbasefee(StackTop stack, ExecutionState& state) noexcept
     stack.push(intx::be::load<uint256>(state.get_tx_context().blob_base_fee));
 }
 
+// NOLINTNEXTLINE(bugprone-exception-escape)
```
gumb0, line 568: `std::get()` raises this, but it's safe because it follows `std::get_if()` with the opposite option. This is weird, because we have the same pattern in other places (`eof.cpp`) and it's not flagged there.

test/state/state.cpp
```diff
@@ -46,13 +51,16 @@ int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noex
 {
     static constexpr auto call_tx_cost = 21000;
     static constexpr auto create_tx_cost = 53000;
+    static constexpr auto per_empty_account_cost = 25000;
```
pdobacz, line 54: I see this name is copied from EIP-7702, but can we give it some context here?

test/state/state.cpp
```diff
@@ -405,6 +435,54 @@ std::variant<TransactionReceipt, std::error_code> transition(State& state, const
...
+            // Skip if authority has non-delegated code.
+            if (!authority_ptr->code.empty() && !is_code_delegated(authority_ptr->code))
+                continue;
+
+            // Skip if authorization nonce is incorrect.
+            if (auth.nonce != authority_ptr->nonce)
+                continue;
+
+            // Refund if authority account creation is not needed.
+            static constexpr int64_t EXISTING_AUTHORITY_REFUND = 25000 - 2500;
```
pdobacz, line 464: Use constants here? we have `per_empty_account_cost`, need to add the other one.

pdobacz (approved):
LGTM, only minor stuff

test/state/state.cpp
gumb0: Made it two global constants

test/statetest/statetest_loader.cpp
```diff
@@ -341,6 +382,12 @@ static void from_json_tx_common(const json::json& j, state::Transaction& o)
         for (const auto& hash : *it)
             o.blob_hashes.push_back(from_json<bytes32>(hash));
     }
+    else if (const auto au_it = j.find("authorizationList"); au_it != j.end())
+    {
+        o.authorization_list = from_json<state::AuthorizationList>(*au_it);
+        if (o.type <= state::Transaction::Type::eip1559)
```
chfast, line 388: Why only update the transaction type conditionally?

Also I noticed we don't read the transaction type directly from JSON but EEST may produce these.

gumb0: I'm not sure why, I think, @pdobacz, you wrote this, do you remember anything?

gumb0: Removed the condition when rebasing.

pdobacz: I don't remember sorry, and I cannot see the full diff here, I don't know the context (what followed after the `if`) to aid my memory. If I could see my original commit maybe, but it seems I don't have that either.

chfast:
Remember to drop "WARNING: test case with set_code transaction (type 4) skipped".

test/statetest/statetest_loader.cpp
```diff
@@ -128,14 +128,13 @@ state::AuthorizationList from_json<state::AuthorizationList>(const json::json& j
...
-        authorization.chain_id = from_json<intx::uint256>(a.at("chainId"));
+        authorization.chain_id = from_json<uint256>(a.at("chainId"));
         authorization.addr = from_json<address>(a.at("address"));
         authorization.nonce = from_json<uint64_t>(a.at("nonce"));
-        if (a.contains("signer"))
-            authorization.signer = from_json<address>(a.at("signer"));
-        authorization.r = from_json<intx::uint256>(a.at("r"));
-        authorization.s = from_json<intx::uint256>(a.at("s"));
-        authorization.v = from_json<intx::uint256>(a.at("v"));
+        authorization.signer = load_if_exists<address>(a, "signer");
```
gumb0, line 134: I think it sets it to 0 address, when not present in the test, but those are tests with invalid signature and it should stay `std::nullopt`

chfast: Yes, correct. I noticed this by state tests. Fixed now.

test/unittests/state_transition_eip7702_test.cpp
```diff
@@ -0,0 +1,305 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
```
chfast, line 2: Is this so old?

lib/evmone/delegation.hpp
```diff
@@ -23,20 +23,23 @@ inline constexpr bool is_code_delegated(bytes_view code) noexcept
 }
 
 /// Get EIP-7702 delegate address from the code of addr, if it is delegated.
+/// TODO: This should not be inline function.
```
gumb0, line 26: @chfast should I move it to delegation.cpp?

test/unittests/state_transition_eip7702_test.cpp
gumb0: I don't think it can be, will change to 2024

lib/evmone/delegation.hpp
chfast: I think yes. You may also swap parameters, I think `host`, `addr` would be more natural. But up to you.

test/state/state.cpp
```diff
@@ -460,6 +480,85 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+
+            // Skip if authorization nonce is incorrect.
+            if (auth.nonce != 0)
+                continue;
+        }
+
+        // Refund if authority account creation is not needed,
+        // i.e. if it had non-empty balance, nonce, code or storage.
+        if (authority_ptr != nullptr &&
+            (!authority_ptr->is_empty() || authority_ptr->has_initial_storage))
```
chfast, line 530: Inspecting storage is quite problematic here. There should be a better way to answer the question of the account is in the trie.

chfast (commented):
There are some artificial enviroments where this can crash:
- having invalid delegation designator in state,
- having a precompile with code.

Do you think we should add sanity checks for these when loading tests?

test/state/state.cpp
```diff
@@ -460,6 +480,86 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
     assert(sender_acc.nonce < Account::NonceMax);  // Required for valid tx.
     ++sender_acc.nonce;                            // Bump sender nonce.
 
+    auto delegation_refund = 0;
+
+    for (const auto& auth : tx.authorization_list)
```
chfast, line 485: Authority list processing should be a separate function.

chfast (changes requested):

test/state/state.cpp
```diff
@@ -460,6 +480,88 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+    {
+        // 1. Verify the chain id is either 0 or the chain’s current ID.
+        // FIXME: Transaction chain id is not the chain id. Can it be 0?
```
chfast, line 488: Please check if tx.chain_id may be 0.

test/state/state.cpp
```diff
@@ -460,6 +480,88 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+        // 3. Verify if the signer has been successfully recovered from the signature.
+        //    authority = ecrecover(...)
+        if (!auth.signer.has_value())
+            continue;
+
+        // s value must be less than or equal to secp256k1n/2, as specified in EIP-2.
+        if (auth.s > SECP256K1N_OVER_2)
+            continue;
+
+        // FIXME: In previous impl: If authority exists, but is empty, it will not be warm.
```
chfast, line 505: Add a test case.

test/state/state.cpp
```diff
@@ -460,6 +480,88 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+
+        // 7. Add PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST gas to the global refund counter
+        // if authority exists in the trie.
+        // Fully empty accounts (no storage) may not exist in the trie at the beginning of the
+        // transaction so we are looking for new accounts created by processing this authorisation
+        // list that remain empty because of a failed authorisation. At this point we are going to
+        // bump its nonce, and it will remain in the trie. Therefore, no refund should be applied.
+        if (!(authority.erase_if_empty && authority.is_empty()))
+        {
+            // FIXME: Maybe it's better to count number of refunds and compute the total amount
```
chfast, line 533: Consider counting the number of refunds.

test/state/state.cpp
```diff
@@ -460,6 +480,88 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+            authority.code_hash = Account::EMPTY_CODE_HASH;
+        }
+        else
+        {
+            bytes new_code(bytes(DELEGATION_MAGIC) + bytes(auth.addr));
+            if (authority.code != new_code)
+                authority.code_changed = true;
+            authority.code = std::move(new_code);
+
+            // TODO The hash of delegated accounts is not used anywhere,
```
chfast, line 554: This is probably not true any more: now `EXTCODEHASH` returns it.

test/state/state.cpp
chfast: There is a kinda-invalid test case for it: the one with "empty but with storage" account. In this case, in previous implementation this account will not be marked as warm. The test don't exploit the warmness aspect so this is undetected.

lib/evmone/instructions_calls.cpp
chfast: Conclusions are good, but I think this is fine for now.

test/state/state.cpp
chfast: I checked this and `tx.chain_id` potentially can be 0 but it has no special meaning. So if we get a tx with chain_id 0 the authorization will be valid iff the authorisation chain_id is 0 so this is kinda fine.

On the other hand, a tx with chain_id 0 should not reach this point because by EIP-155 the configured chain ID must not be 0. We don't really have a check for this because it looks like state tests don't explore this area.

test/state/state.cpp
```diff
@@ -460,6 +480,89 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+        // In case authority does not exist in the trie, verify that nonce is equal to 0.
+        if (auth.nonce != authority.nonce)
+            continue;
+
+        // 7. Add PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST gas to the global refund counter
+        // if authority exists in the trie.
+        // We are interested in _empty_ accounts (EIP-161) because _empty_ implies they don't exist
+        // in the state (EIP-7523). Successful authorisation validation makes an account non-empty.
+        // We apply the refund only if the account has been non-empty before.
+        // TODO: The additional condition (erase_if_empty) is to handle existent-but-empty which
```
gumb0, line 527: `authority` here doesn't necessary represent what is commited to the state, but intermediate in-memory account object, existing during transaction processing. It can indeed be empty, not only in testing environments (this is complaint about comment formulation)

`erase_if_empty` here seems to be a hacky way to represent the flag "account was created in this transaction". Instead you should have a separate flag that is set when authority is not found in the state the first time it is mentioned.

test/state/state.cpp
```diff
@@ -480,7 +480,7 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
     assert(sender_acc.nonce < Account::NonceMax);  // Required for valid tx.
     ++sender_acc.nonce;                            // Bump sender nonce.
 
-    auto delegation_refund = 0;
+    int64_t delegation_refund = 0;
```
gumb0, line 483: Why is it `int64_t`?

test/state/state.cpp
```diff
@@ -460,6 +480,89 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
...
+        //    authority = ecrecover(...)
+        if (!auth.signer.has_value())
+            continue;
+
+        // s value must be less than or equal to secp256k1n/2, as specified in EIP-2.
+        if (auth.s > SECP256K1N_OVER_2)
+            continue;
+
+        // Get or create the authority account.
+        // It is still empty at this point until nonce bump followed by successful authorization.
```
gumb0, line 506:
```suggestion
        // It is still empty at this point until nonce bump following successful authorization.
```

test/state/state.cpp
chfast: The `erase_if_empty` means "empty account created in this transaction". The name is probably bad because it says what to do with it rather than what it is, but besides the name I think it suits this case pretty well.

gumb0:
> There are some artificial enviroments where this can crash:
> 
> * having invalid delegation designator in state,
> * having a precompile with code.
> 
> Do you think we should add sanity checks for these when loading tests?

I think it makes sense to add in a separate PR

test/state/state.cpp
```diff
@@ -90,6 +90,96 @@ TransactionCost compute_tx_intrinsic_cost(evmc_revision rev, const Transaction&
...
+        // Successful authorisation validation makes an account non-empty.
+        // We apply the refund only if the account has existed before.
+        // We detect "exists in the trie" by inspecting _empty_ property (EIP-161) because _empty_
+        // implies an account doesn't exist in the state (EIP-7523).
+        //
+        // For completeness, we also consider _empty_ accounts in the state as existent
+        // and apply refund for those. The erase_if_empty indicates the account has been created
+        // as empty while processing this authorization list.
+        // TODO: Such accounts are against EIP-7523 and may happen only in testing environments.
+        const auto empty_in_state = authority.is_empty() && !authority.erase_if_empty;
```
gumb0, line 144: Maybe `in_state_but_empty` ?

gumb0: I'd also mention that it means that storage is non-empty

chfast: > I'd also mention that it means that storage is non-empty

This is not accurate. There can be even fully empty account.

chfast (approved):

PR #968, opened 2024-08-08
test: Convert example tests to TestState API

chfast:


chfast:
If you are working with state tests, there is new `TestState` API you should be using. The API is actually `std::map`.

pdobacz (approved):

PR #969, opened 2024-08-08
test: Add tests for create collision with storage

chfast:
Add tests for EIP-7610 behavior: create address colliding with an account having non-empty storage should fail.

### DIFF
--- a/test/unittests/state_transition_create_test.cpp
+++ b/test/unittests/state_transition_create_test.cpp
@@ -176,6 +176,16 @@ TEST_F(state_transition, create_tx_collision)
     expect.post[CREATED].nonce = 2;
 }
 
+TEST_F(state_transition, create_tx_collision_storage)
+{
+    // Test create address collision with an account only having storage (EIP-7610).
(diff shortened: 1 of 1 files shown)


pdobacz (approved):
LGTM. 2 points:
- don't we prefer to rely on EEST for that?
- if not, then for completeness we should have a test for eof creation tx too?

chfast:
> LGTM. 2 points:
> 
>     * don't we prefer to rely on EEST for that?
> 
>     * if not, then for completeness we should have a test for eof creation tx too?

ethereum/tests have this covered, I just need a local test because I plan to rework the feature. The EOF context: this is one of the 100s EVM features we may want to test from EOF.

chfast:
> LGTM. 2 points:
> 
>     * don't we prefer to rely on EEST for that?
> 
>     * if not, then for completeness we should have a test for eof creation tx too?

ethereum/tests have this covered, I just need a local test because I plan to rework the feature. The EOF context: this is one of the 100s EVM features we may want to test from EOF.

PR #970, opened 2024-08-08
ci: Update ethereum/tests EOF tests

chfast:
Switch to the updated `eof-toplevel` branch of ipsilon/tests.
After rebasing the number of changes is smaller than it was.
The ethereum/tests PR: https://github.com/ethereum/tests/pull/1381.

Also run the EOF tests from upstream ethereum/tests v14.0
with longer exclusion list.

PR #971, opened 2024-08-08
ci: Include ethereum/tests@v14.0 in EOF testing

pdobacz:
The motivation behind having this side-by-side the tests from ipsilon fork is to tick-off `ethereum/tests` at some publicly recognized version (will be removed after we catch `ethereum/tests` up, i.e. remove failing ones from there and migrate to EEST)


chfast:
I added the same to https://github.com/ethereum/evmone/pull/970.

PR #953, opened 2024-07-22
EIP-2935: Serve historical block hashes from state

chfast:
Implement [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) "Serve historical block hashes from state" required for the [Prague](https://eips.ethereum.org/EIPS/eip-7600) revision.
This mostly adds new system contract.

PR #975, opened 2024-08-19
blockchaintest: Optional partial state hash check

chfast:
This is blockchain tests execution optimization: for listed test names
only check state root hash of first 5 blocks (to detect early problems)
and last 5 blocks (to do the final check of the chain of blocks).

The current implementation of the MPT hash of the state builds
the trie from scratch (no updates to the trie of the previous block).
For the tests will a long chain of blocks the performance degrades
significantly with 99% time spent in the keccak hash function.

This improves testing of EIP-2935
implemented in https://github.com/ethereum/evmone/pull/953.

### DIFF
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -116,6 +116,22 @@ std::string print_state(const TestState& s)
 
     return out.str();
 }
+
+/// How many first/last state hashes to check in the series of block for the "partial" check.
+constexpr size_t PARTIAL_STATE_HASH_CHECK_MARGIN = 5;
+
+/// List of test names for which only perform the partial state root hash checks.
+/// For tests on the list, only check state hashes for first M (early problems)
+/// and last M (final check of the chain) blocks,
+/// where M is defined by PARTIAL_STATE_HASH_CHECK_MARGIN.
+/// Otherwise, for very long blockchain test we spend a lot of time in MPT hashing
+/// because the implementation of it is very simplistic: the trie is not updated
+/// along the state changes but a new trie is build from scratch for every block.
+constexpr std::array PARTIAL_STATE_HASH_CHECK_TESTS{
+    "a test name",
+};
+
+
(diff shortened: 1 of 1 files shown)


pdobacz (commented):
Is this temporary, until the MPT implementation is optimized, or forever? It sounds a tiny bit scary

chfast:
Temporary but for long time :) Updating MTP seems quite a work and we only need this for testing.

Because the state in the final block is evolution of all previous states I think this is relatively fine. The only situation it can miss is when in a middle block state is incorrect but it corrects itself before the final block.

The other option is to list the test names for which we want to do the light check.

For the EEST stable tests the number of blocks is not bigger than 30.

pdobacz:
> Temporary but for long time :) 

:)
 
> Because the state in the final block is evolution of all previous states I think this is relatively fine. The only situation it can miss is when in a middle block state is incorrect but it corrects itself before the final block.

We had a very similar case of a self-cancelling test (IIRC EXCHANGE), hence my worries

> The other option is to list the test names for which we want to do the light check.
> 
> For the EEST stable tests the number of blocks is not bigger than 30.

Or, can we assume a threshold (like 30), above which we automatically switch to the light check?

We could also check "every 7th" or sth like that, in addition to first 5 + last 5. This still keeps us under the 30 blocks mark, WDYT?

gumb0:
> 
> We had a very similar case of a self-cancelling test (IIRC EXCHANGE), hence my worries
> 

Given that system calls now modify state every block (beacon chain roots since Cancun and historical block hashes since  Prague), it seems highly unlikely that state changes will cancel each other over several blocks?

chfast:
> I think at least since Prague, as block hashes are saved into state, the wrong state changes cannot cancel out easily.

Hm... now I noticed this is weird: we need to compute the block hash which includes the state root hash for EIP-2935, so how all this works as we just disabled this computation...

pdobacz (approved):

kmk142789 (approved):

PR #976, opened 2024-08-19
Refactor `system_call()`

chfast:
Refactor and abstract the `system_call()` to make adding new system contracts easier.

### DIFF
--- a/test/state/system_contracts.cpp
+++ b/test/state/system_contracts.cpp
@@ -1,0 +1,45 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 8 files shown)


battlmonstr (approved):

pdobacz (approved):
thx for the comment, I think it's useful, I only scanned 4788 and missed the empty account part

test/state/system_contracts.cpp
```diff
@@ -0,0 +1,74 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2023 The evmone Authors.
```
gumb0, line 2:
```suggestion
// Copyright 2024 The evmone Authors.
```

chfast: 2023 is the year then this code was introduced.

PR #978, opened 2024-08-21
eof: fix multi-container-section case

pdobacz:
Credit to @chfast to find the case via fuzzing, as well as suggest this version of the fix (alternative version in comment, not sure which is cleaner anymore)

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -72,6 +72,7 @@ EOFValidationError get_section_missing_error(uint8_t section_id) noexcept
         return EOFValidationError::type_section_missing;
     case CODE_SECTION:
         return EOFValidationError::code_section_missing;
+    case CONTAINER_SECTION:
     case DATA_SECTION:
         return EOFValidationError::data_section_missing;
     default:
@@ -106,7 +107,7 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_section_headers(byt
             // If DATA_SECTION is expected, CONTAINER_SECTION is also allowed, because
             // container section is optional.
             if (section_id != expected_section_id &&
-                (expected_section_id != DATA_SECTION || section_id != CONTAINER_SECTION))
+                (expected_section_id != CONTAINER_SECTION || section_id != DATA_SECTION))
                 return get_section_missing_error(expected_section_id);
 
             switch (section_id)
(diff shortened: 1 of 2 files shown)


lib/evmone/eof.cpp
```diff
@@ -106,7 +107,7 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_section_headers(byt
             // If DATA_SECTION is expected, CONTAINER_SECTION is also allowed, because
             // container section is optional.
             if (section_id != expected_section_id &&
-                (expected_section_id != DATA_SECTION || section_id != CONTAINER_SECTION))
+                (expected_section_id != CONTAINER_SECTION || section_id != DATA_SECTION))
```
pdobacz, line 110: alternative is to change the condition to

```
            if (section_id != expected_section_id &&
                !(expected_section_id == DATA_SECTION && section_id == CONTAINER_SECTION &&
                    section_headers[CONTAINER_SECTION].size() == 0))
```

and leave the rest as is

chfast:
Let me try one more thing here.

chfast:
I pushed an alternative fix where we check `id != expected` twice but the helper `get_section_missing_error()` remains sane.

lib/evmone/eof.cpp
```diff
@@ -104,10 +103,11 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_section_headers(byt
         {
             section_id = *it++;
 
-            // If DATA_SECTION is expected, CONTAINER_SECTION is also allowed, because
-            // container section is optional.
-            if (section_id != expected_section_id &&
-                (expected_section_id != CONTAINER_SECTION || section_id != DATA_SECTION))
+            // Skip optional sections.
+            if (section_id != expected_section_id && expected_section_id == CONTAINER_SECTION)
```
pdobacz, line 107: matter of taste, but this reads weirder to me, weirder than the `get_section_missing_error` modification, and weirder than the "2nd" alternative which checks `CONTAINER_SECTION.size()` (which also keeps `get_section_missing_error` intact). The weirdness comes from the fact that the `expected_section_id` isn't a check-modify, but a check-maybe modify-check-modify sequence on each iteration.

Maybe we need a 3rd opinion to pick the least obtrusive fix, @gumb0 ?

gumb0: I think definitely this one here is more understandable.

gumb0 (approved):

chfast:
This issue was found indirectly by fuzzing: I was debugging the fuzzer mutation and noticed this weird EOF header. This indicates this validation issue may not be exploitable. Yet, it is a serious bug.

PR #982, opened 2024-09-02
precompiles: Implement EIP-2537's `bls12_g1add`

rodiazet:
Implementation of the `bls12_g1add` precompile: E1 affine points' addition from BLS12-381 curve according to the EIP-2537 spec https://eips.ethereum.org/EIPS/eip-2537#abi-for-g1-addition.

Depends on https://github.com/ethereum/evmone/pull/984

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -371,7 +371,9 @@ jobs:
           # Tests for in-development EVM revision currently passing.
           working_directory: ~/spec-tests/fixtures/state_tests
           command: >
-            ~/build/bin/evmone-statetest prague/eip2537_bls_12_381_precompiles/bls12_precompiles_before_fork
+            ~/build/bin/evmone-statetest 
+            prague/eip2537_bls_12_381_precompiles/bls12_precompiles_before_fork
+            prague/eip2537_bls_12_381_precompiles/bls12_g1add
       - collect_coverage_gcc
       - upload_coverage:
           flags: execution_spec_tests
--- a/lib/evmone_precompiles/bls.cpp
+++ b/lib/evmone_precompiles/bls.cpp
@@ -1,0 +1,71 @@
+#include "bls.hpp"
+#include <blst.h>
+#include <optional>
+
+namespace evmone::crypto::bls
+{
+namespace
+{
+/// Offset of the beginning of field element. First 16 bytes must be zero according to spec
+/// https://eips.ethereum.org/EIPS/eip-2537#field-elements-encoding
+constexpr auto FP_BYTES_OFFSET = 64 - 48;
+
+/// Validates p1 affine point. Checks that point coordinates are from the BLS12-381 field and
+/// that the point is on curve. https://eips.ethereum.org/EIPS/eip-2537#abi-for-g1-addition
+[[nodiscard]] std::optional<blst_p1_affine> validate_point(
+    const uint8_t _x[64], const uint8_t _y[64]) noexcept
+{
+    constexpr auto is_field_element = [](const uint8_t _p[64]) {
+        return intx::be::unsafe::load<intx::uint512>(_p) < BLS_FIELD_MODULUS;
+    };
+
+    if (!is_field_element(_x))
+        return std::nullopt;
+    if (!is_field_element(_y))
+        return std::nullopt;
+
+    blst_fp x;
+    blst_fp y;
+    blst_fp_from_bendian(&x, &_x[FP_BYTES_OFFSET]);
+    blst_fp_from_bendian(&y, &_y[FP_BYTES_OFFSET]);
+
+    const blst_p1_affine p0_affine{x, y};
+    if (!blst_p1_affine_on_curve(&p0_affine))
+        return std::nullopt;
+
+    return p0_affine;
+}
+
+/// Stores p1 point in two 64-bytes arrays with big endian encoding zero padded.
+void store_result(uint8_t _rx[64], uint8_t _ry[64], const blst_p1* out) noexcept
+{
+    blst_p1_affine result;
+    blst_p1_to_affine(&result, out);
+
+    std::memset(_rx, 0, FP_BYTES_OFFSET);
+    blst_bendian_from_fp(&_rx[FP_BYTES_OFFSET], &result.x);
+    std::memset(_ry, 0, FP_BYTES_OFFSET);
+    blst_bendian_from_fp(&_ry[FP_BYTES_OFFSET], &result.y);
+}
+}  // namespace
+
+[[nodiscard]] bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64],
+    const uint8_t _y0[64], const uint8_t _x1[64], const uint8_t _y1[64]) noexcept
+{
+    const auto p0_affine = validate_point(_x0, _y0);
+    const auto p1_affine = validate_point(_x1, _y1);
+
+    if (!p0_affine.has_value() || !p1_affine.has_value())
+        return false;
+
+    blst_p1 p0;
+    blst_p1_from_affine(&p0, &*p0_affine);
+
+    blst_p1 out;
+    blst_p1_add_or_double_affine(&out, &p0, &*p1_affine);
+
+    store_result(_rx, _ry, &out);
+
+    return true;
+}
+}  // namespace evmone::crypto::bls
--- a/lib/evmone_precompiles/bls.hpp
+++ b/lib/evmone_precompiles/bls.hpp
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <intx/intx.hpp>
+
+using namespace intx::literals;
+
+namespace evmone::crypto::bls
+{
+/// The BLS12-381 field prime number.
+inline constexpr auto BLS_FIELD_MODULUS =
+    0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab_u384;
+
+/// Addition in BLS12-381 curve group.
+///
+/// Computes P ⊕ Q for two points in affine coordinates on the BLS12-381 curve,
+[[nodiscard]] bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64],
+    const uint8_t _y0[64], const uint8_t _x1[64], const uint8_t _y1[64]) noexcept;
+
+}  // namespace evmone::crypto::bls
--- a/test/state/precompiles.cpp
+++ b/test/state/precompiles.cpp
@@ -6,6 +6,7 @@
 #include "precompiles_internal.hpp"
 #include "precompiles_stubs.hpp"
 #include <evmone_precompiles/blake2b.hpp>
+#include <evmone_precompiles/bls.hpp>
 #include <evmone_precompiles/bn254.hpp>
 #include <evmone_precompiles/ripemd160.hpp>
 #include <evmone_precompiles/secp256k1.hpp>
@@ -155,8 +156,8 @@ PrecompileAnalysis point_evaluation_analyze(bytes_view, evmc_revision) noexcept
 
 PrecompileAnalysis bls12_g1add_analyze(bytes_view, evmc_revision) noexcept
 {
-    // TODO: Implement
-    return {GasCostMax, 0};
+    static constexpr auto BLS12_G1ADD_PRECOMPILE_GAS = 500;
+    return {BLS12_G1ADD_PRECOMPILE_GAS, 128};
 }
 
 PrecompileAnalysis bls12_g1mul_analyze(bytes_view, evmc_revision) noexcept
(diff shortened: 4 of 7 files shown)


lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,42 @@
...
+namespace evmone::crypto::bls
+{
+bool g1_add(uint8_t* _rx, uint8_t* _ry, const uint8_t _x0[48], const uint8_t _y0[48],
+    const uint8_t _x1[48], const uint8_t _y1[48])
+{
+    blst::blst_fp x0;
+    blst::blst_fp y0;
+    blst::blst_fp x1;
+    blst::blst_fp y1;
+    blst::blst_fp_from_bendian(&x0, _x0);
```
jwasinger, line 13: You'll want to check that these are less than the `fp` modulus.

test/state/precompiles.cpp
```diff
@@ -292,6 +301,40 @@ ExecutionResult blake2bf_execute(const uint8_t* input, [[maybe_unused]] size_t i
     return {EVMC_SUCCESS, sizeof(h)};
 }
 
+ExecutionResult bls12_g1add_execute(
+    const uint8_t* input, size_t input_size, uint8_t* output, size_t output_size) noexcept
+{
+    static_assert(std::endian::native == std::endian::little,
```
chfast, line 307: Why?

test/state/precompiles.cpp
```diff
@@ -292,6 +301,40 @@ ExecutionResult blake2bf_execute(const uint8_t* input, [[maybe_unused]] size_t i
...
+    if (!verify_element(input))
+        return {EVMC_PRECOMPILE_FAILURE, 0};
+    if (!verify_element(&input[64]))
+        return {EVMC_PRECOMPILE_FAILURE, 0};
+    if (!verify_element(&input[128]))
+        return {EVMC_PRECOMPILE_FAILURE, 0};
+    if (!verify_element(&input[192]))
+        return {EVMC_PRECOMPILE_FAILURE, 0};
+
+    if (!crypto::bls::g1_add(&output[16], &output[16 + 64], &input[16], &input[16 + 64],
```
chfast, line 331: I think I would just send 64 byte field element down the `crypto` function and validate there as `uint512`. Or even a point as 128 bytes... Otherwise we have validation split into two parts.

chfast (changes requested):

lib/evmone_precompiles/bls.cpp
rodiazet: It was checked but on a different level. Now this check is moved to `g1_add` body

test/state/precompiles.cpp
rodiazet: Implemented in https://github.com/ethereum/evmone/pull/984

test/state/precompiles.hpp
```diff
@@ -22,11 +22,13 @@ enum class PrecompileId : uint8_t
     ecpairing = 0x08,
     blake2bf = 0x09,
     point_evaluation = 0x0a,
+    bls12_g1add = 0x0b,
```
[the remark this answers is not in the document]
rodiazet: Implemented in https://github.com/ethereum/evmone/pull/984

test/state/precompiles.cpp
```diff
@@ -346,9 +349,19 @@ ExecutionResult blake2bf_execute(const uint8_t* input, [[maybe_unused]] size_t i
...
 
-ExecutionResult bls12_g1add_execute(const uint8_t*, size_t, uint8_t*, size_t) noexcept
+ExecutionResult bls12_g1add_execute(
+    const uint8_t* input, size_t input_size, uint8_t* output, size_t output_size) noexcept
 {
-    return {EVMC_PRECOMPILE_FAILURE, 0};
+    if (input_size != 256)
+        return {EVMC_PRECOMPILE_FAILURE, 0};
+
+    if (output_size != 128)
```
chfast, line 358: This is guaranteed by the "analysis" so change to `assert`.

circle.yml
```diff
@@ -371,7 +371,7 @@ jobs:
           # Tests for in-development EVM revision currently passing.
           working_directory: ~/spec-tests/fixtures/state_tests
           command: >
-            ~/build/bin/evmone-statetest prague/eip2537_bls_12_381_precompiles/bls12_precompiles_before_fork
+            ~/build/bin/evmone-statetest prague/eip2537_bls_12_381_precompiles/bls12_precompiles_before_fork prague/eip2537_bls_12_381_precompiles/bls12_g1add
```
chfast, line 374: put the paths on separate lines

lib/evmone_precompiles/bls.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <intx/intx.hpp>
+
+// TODO(intx): Add ""_u384.
+inline constexpr auto operator""_u384(const char* s)
```
chfast, line 6: Requires https://github.com/ethereum/evmone/pull/985.

lib/evmone_precompiles/bls.hpp
```diff
@@ -0,0 +1,19 @@
...
+{
+    return intx::from_string<intx::uint384>(s);
+}
+
+namespace evmone::crypto::bls
+{
+inline constexpr auto FieldPrime =
```
chfast, line 13: Because 4844 uses name `BLS_MODULUS` let's call this one `BLS_FIELD_MODULUS`.

lib/evmone_precompiles/bls.hpp
```diff
@@ -0,0 +1,19 @@
...
+    0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab_u384;
+
+bool g1_add(uint8_t* _rx, uint8_t* _ry, const uint8_t _x0[64], const uint8_t _y0[64],
```
chfast, line 16:
```suggestion
bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64], const uint8_t _y0[64],
```

test/state/precompiles.cpp
```diff
@@ -6,15 +6,18 @@
 #include "precompiles_internal.hpp"
 #include "precompiles_stubs.hpp"
 #include <evmone_precompiles/blake2b.hpp>
+#include <evmone_precompiles/bls.hpp>
 #include <evmone_precompiles/bn254.hpp>
 #include <evmone_precompiles/ripemd160.hpp>
 #include <evmone_precompiles/secp256k1.hpp>
 #include <evmone_precompiles/sha256.hpp>
 #include <intx/intx.hpp>
+#include <algorithm>
```
chfast, line 15: Seems not needed.

test/state/precompiles.cpp
```diff
@@ -6,15 +6,18 @@
...
 #include <array>
 #include <bit>
 #include <cassert>
 #include <limits>
+#include <span>
```
chfast, line 20: Seems not needed.

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,57 @@
+#include "bls.hpp"
+#include <blst.hpp>
+
+namespace evmone::crypto::bls
+{
+bool g1_add(uint8_t* _rx, uint8_t* _ry, const uint8_t _x0[64], const uint8_t _y0[64],
+    const uint8_t _x1[64], const uint8_t _y1[64])
+{
+    constexpr auto is_field_element = [](const uint8_t p[64]) {
+        return intx::be::unsafe::load<intx::uint512>(p) < crypto::bls::FieldPrime;
```
chfast, line 10:
```suggestion
        return intx::be::unsafe::load<intx::uint512>(p) < FieldPrime;
```

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (!is_field_element(_x1))
+        return false;
+    if (!is_field_element(_y1))
+        return false;
+
+    blst::blst_fp x0;
+    blst::blst_fp y0;
+    blst::blst_fp x1;
+    blst::blst_fp y1;
+    blst::blst_fp_from_bendian(&x0, &_x0[16]);
```
chfast, line 26: Add a named constant for `16`. E.g. `FP_BYTES_OFFSET = 64-48`.

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,57 @@
...
+    blst::blst_p1_from_affine(&p0, &p0_affine);
+    blst::blst_p1_from_affine(&p1, &p1_affine);
+
+    blst::blst_p1 out;
+    blst::blst_p1_add_or_double(&out, &p0, &p1);
+
+    blst::blst_p1_affine result;
+    blst::blst_p1_to_affine(&result, &out);
+
+    memset(_rx, 0, 16);
```
chfast, line 50:
```suggestion
    std::memset(_rx, 0, 16);
```

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,57 @@
+#include "bls.hpp"
+#include <blst.hpp>
```
chfast, line 2:
```suggestion
#include <blst.h>
```

This also eliminates the need of `blst::`

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (!blst::blst_p1_affine_on_curve(&p1_affine))
+        return false;
+
+    blst::blst_p1 p0;
+    blst::blst_p1 p1;
+    blst::blst_p1_from_affine(&p0, &p0_affine);
+    blst::blst_p1_from_affine(&p1, &p1_affine);
+
+    blst::blst_p1 out;
+    blst::blst_p1_add_or_double(&out, &p0, &p1);
```
chfast, line 45: There is `blst_p1_add_or_double_affine` so you don't have to convert both points to affine.

chfast (changes requested):

lib/evmone_precompiles/bls.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+using namespace intx::literals;
+
+namespace evmone::crypto::bls
+{
+inline constexpr auto BLS_FIELD_MODULUS =
+    0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab_u384;
+
+bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64], const uint8_t _y0[64],
+    const uint8_t _x1[64], const uint8_t _y1[64]);
```
chfast, line 13:
```suggestion
    const uint8_t _x1[64], const uint8_t _y1[64]) noexcept;
```

lib/evmone_precompiles/bls.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+using namespace intx::literals;
+
+namespace evmone::crypto::bls
+{
+/// The bls12-381 field prime number.
+inline constexpr auto BLS_FIELD_MODULUS =
+    0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab_u384;
+
+/// Addition in bls12-381 curve group.
```
chfast, line 13:
```suggestion
/// Addition in BLS12-381 curve group.
```

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,66 @@
...
+    const uint8_t _y0[64], const uint8_t _x1[64], const uint8_t _y1[64]) noexcept
+{
+    const auto p0_affine = validate_point(_x0, _y0);
+    const auto p1_affine = validate_point(_x1, _y1);
+
+    if (!p0_affine.has_value() || !p1_affine.has_value())
+        return false;
+
+    blst_p1 p0;
+    blst_p1_from_affine(&p0, &p0_affine.value());
```
chfast, line 57:
```suggestion
    blst_p1_from_affine(&p0, &*p0_affine);
```
We should use `*optional` instead of `optional.value()` because the later throws an exception and we promised not to throw.

chfast (changes requested):

lib/evmone_precompiles/bls.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+namespace
+{
+constexpr auto FP_BYTES_OFFSET = 64 - 48;
+
+[[nodiscard]] std::optional<blst_p1_affine> validate_point(
+    const uint8_t _x[64], const uint8_t _y[64]) noexcept
+{
+    constexpr auto is_field_element = [](const uint8_t _p[64]) {
+        return intx::be::unsafe::load<intx::uint512>(_p) < crypto::bls::BLS_FIELD_MODULUS;
```
chfast, line 15:
```suggestion
        return intx::be::unsafe::load<intx::uint512>(_p) < BLS_FIELD_MODULUS;
```

chfast (changes requested):
Please add better PR description.

chfast (approved):

PR #989, opened 2024-09-05
eof: Add trait since when an instruction is in EOF

chfast:
Add `std::optional<evmc_revision> eof_since` to the `instr::traits` with the information in which EVM revision an instruction has become valid in EOF.

Using `std::optional` has some inconvenience:
the `std::nullopt` is less-than any concrete value (think "Frontier-1") while for undefined instructions we rather want to assign +infinity.


lib/evmone/instructions_traits.hpp
```diff
@@ -200,6 +200,9 @@ constexpr inline GasCostTable gas_costs = []() noexcept {
 static_assert(gas_costs[EVMC_MAX_REVISION][OP_ADD] > 0, "gas costs missing for a revision");
 
 
+constexpr auto REV_EOF1 = EVMC_PRAGUE;
```
chfast, line 203: Good idea?

lib/evmone/instructions_traits.hpp
```diff
@@ -223,6 +226,8 @@ struct Traits
     /// every EVM revision the value is ::EVMC_FRONTIER. For undefined instructions the value is not
     /// available.
     std::optional<evmc_revision> since;
+
+    std::optional<evmc_revision> eof_since;
```
chfast, line 230: Comment missing.

gumb0:
Another idea is having one `since` revision + 2 flags `legacy` and `eof`, assuming when instruction is activated in both, it's always activated in the same revision.

Argument against can be that it can be seen as some old opcodes are activated before Prague, but also have `eof` flag, so might be interpreted as "activated in eof before Prague", but I think it shouldn't influnce execution, it's equivalent to the table before this change.

chfast:
> Another idea is having one `since` revision + 2 flags `legacy` and `eof`, assuming when instruction is activated in both, it's always activated in the same revision.
> 
> Argument against can be that it can be seen as some old opcodes are activated before Prague, but also have `eof` flag, so might be interpreted as "activated in eof before Prague", but I think it shouldn't influnce execution, it's equivalent to the table before this change.

I considered this but this is less generic, requires one more field, and the main usage (build cost table) would need to use both revision+flag anyway.

lib/evmone/instructions_traits.hpp
chfast: Is the name good? For `TXCREATE` we need something like "EOF v1.1".

gumb0: I think it's okay. Also okay to drop `TXCREATE`

pdobacz (approved):
lgtm, will there also be `eof_until`, in case we ever start removing instructions from subsequent EOF majors?

lib/evmone/instructions_traits.hpp
```diff
@@ -223,6 +226,10 @@ struct Traits
     /// every EVM revision the value is ::EVMC_FRONTIER. For undefined instructions the value is not
     /// available.
     std::optional<evmc_revision> since;
+
+    /// The EVM revision in which the instruction has been valid in EOF. For instructions invalid
```
gumb0, line 230:
```suggestion
    /// The EVM revision in which the instruction has become valid in EOF. For instructions invalid
```

gumb0 (approved):

PR #991, opened 2024-09-09
eof: Delay data reference init until execution

chfast:


pdobacz (commented):
makes sense, but isn't a bit risky? we're propagating an incomplete `ExecutionState` instance, and it seems access to the "unset" `data = {}` will silently return zeros (or garbage), in case any path "forgets" to set or reset it.

Is there a way to invalidate an `ExecutionState`'s unset `data` field somehow?

chfast:
The main purpose of this was to remove the "EOF data" field from the execution state and this was achieved in https://github.com/ethereum/evmone/pull/941/commits/6e2f5d9d5f4ce7a38e71a4ec1bd3f016d0742737.

PR #993, opened 2024-09-09
test: Ensure exported state tests have all fields

pdobacz:
I was wrong when I said that the format inconsistency had a test which we can make pass by amending the exported format - there were no exclusions which my fix would fix. So to have some testing mechanism of correctness, we can maybe add an assertion in `statetest_loader` like this? I don't want to overdo it and this seems safe to do.

I guess we can also leave it without the assertion.

I don't know yet about the other format inconsistencies found, so I might add more commits here

### DIFF
--- a/test/statetest/statetest_export.cpp
+++ b/test/statetest/statetest_export.cpp
@@ -8,7 +8,7 @@ namespace evmone::test
 {
 json::json to_json(const TestState& state)
 {
-    json::json j;
+    json::json j = json::json({});
     for (const auto& [addr, acc] : state)
     {
         auto& j_acc = j[hex0x(addr)];
--- a/test/statetest/statetest_export.cpp
+++ b/test/statetest/statetest_export.cpp
@@ -8,7 +8,7 @@ namespace evmone::test
 {
 json::json to_json(const TestState& state)
 {
-    json::json j = json::json({});
+    json::json j = json::json::object();
     for (const auto& [addr, acc] : state)
     {
         auto& j_acc = j[hex0x(addr)];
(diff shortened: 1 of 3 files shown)


test/statetest/statetest_export.cpp
```diff
@@ -8,7 +8,7 @@ namespace evmone::test
 {
 json::json to_json(const TestState& state)
 {
-    json::json j;
+    json::json j = json::json({});
```
chfast, line 11:
```suggestion
    json::json j = json::json::object();
```

chfast (changes requested):

chfast (approved):

pdobacz:
@chfast Hey I think I figured out that it is ok to add accessLists and currentRandom like that (see the recent commit). Both were reported missing by EELS team, PTAL

test/unittests/state_transition.cpp
```diff
@@ -211,6 +212,9 @@ void state_transition::export_state_test(
     jtx["gasLimit"][0] = hex0x(tx.gas_limit);
     jtx["value"][0] = hex0x(tx.value);
 
+    if (tx.type >= Transaction::Type::access_list)
+        jtx["accessLists"][0] = json::json::array({});
```
chfast, line 216:
```suggestion
        jtx["accessLists"][0] = json::json::array();
```

test/unittests/state_transition.cpp
```diff
@@ -211,6 +212,9 @@ void state_transition::export_state_test(
     jtx["gasLimit"][0] = hex0x(tx.gas_limit);
     jtx["value"][0] = hex0x(tx.value);
 
+    if (tx.type >= Transaction::Type::access_list)
```
chfast, line 215: This seems hacky (but I also remember the JSON representation of access list was a mess). Can you add a comment what it does? Can we actually dump the access list when it is not empty?

chfast (approved):
Looks good. Please update the description because it will become the git commit message.

test/unittests/state_transition.cpp
pdobacz: > Can we actually dump the access list when it is not empty?

not sure what you mean here?

chfast: Oh I see now. The non-empty `"accessLists"` is handled next. Just add a comment: "force access list output even if empty".

PR #994, opened 2024-09-10
precompiles: Implement EIP-2537's `bls12_g1mul`

rodiazet:
Implementation of the `bls12_g1mul` precompile: E1 affine point's multiplication from BLS12-381 curve by a scalar according to the EIP-2537 spec https://eips.ethereum.org/EIPS/eip-2537#abi-for-g1-multiplication.

Depends on https://github.com/ethereum/evmone/pull/982

### DIFF
--- a/lib/evmone_precompiles/bls.cpp
+++ b/lib/evmone_precompiles/bls.cpp
@@ -68,4 +68,28 @@ void store_result(uint8_t _rx[64], uint8_t _ry[64], const blst_p1* out) noexcept
 
     return true;
 }
+
+[[nodiscard]] bool g1_mul(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x[64],
+    const uint8_t _y[64], const uint8_t _c[32]) noexcept
+{
+    blst_scalar scalar;
+    blst_scalar_from_bendian(&scalar, _c);
+
+    const auto p_affine = validate_point(_x, _y);
+    if (!p_affine.has_value())
+        return false;
+
+    blst_p1 p;
+    blst_p1_from_affine(&p, &*p_affine);
+
+    if (!blst_p1_in_g1(&p))
+        return false;
+
+    blst_p1 out;
+    blst_p1_mult(&out, &p, scalar.b, 256);
+
+    store_result(_rx, _ry, &out);
+
+    return true;
+}
 }  // namespace evmone::crypto::bls
--- a/lib/evmone_precompiles/bls.hpp
+++ b/lib/evmone_precompiles/bls.hpp
@@ -16,4 +16,11 @@ inline constexpr auto BLS_FIELD_MODULUS =
 [[nodiscard]] bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64],
     const uint8_t _y0[64], const uint8_t _x1[64], const uint8_t _y1[64]) noexcept;
 
+/// Scalar multiplication in BLS12-381 curve G1 subgroup.
+///
+/// Computes [c]P for a point in affine coordinate on the BLS12-381 curve, performs subgroup check
+/// according to spec https://eips.ethereum.org/EIPS/eip-2537#abi-for-g1-multiplication
+[[nodiscard]] bool g1_mul(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x[64],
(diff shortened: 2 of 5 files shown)


lib/evmone_precompiles/bls.cpp
```diff
@@ -68,4 +73,29 @@ void store_result(uint8_t _rx[64], uint8_t _ry[64], const blst_p1* out) noexcept
 
     return true;
 }
+
+[[nodiscard]] bool g1_mul(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x[64],
+    const uint8_t _y[64], const uint8_t _c[32]) noexcept
+{
+    const auto c = intx::be::unsafe::load<intx::uint256>(_c) % MAIN_SUBGROUP_ORDER;
```
chfast, line 80: Instead of shrinking the scalar here, try to pass all the bits to `blst_p1_mult()`.

chfast (changes requested):

lib/evmone_precompiles/bls.cpp
```diff
@@ -6,6 +6,8 @@ namespace evmone::crypto::bls
 {
 namespace
 {
+using namespace intx::literals;
```
chfast, line 9: Not needed any more.

lib/evmone_precompiles/bls.cpp
```diff
@@ -68,4 +70,28 @@ void store_result(uint8_t _rx[64], uint8_t _ry[64], const blst_p1* out) noexcept
 
     return true;
 }
+
+[[nodiscard]] bool g1_mul(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x[64],
+    const uint8_t _y[64], const uint8_t _c[32]) noexcept
+{
+    blst_scalar scalar;
+    blst_scalar_from_bendian(&scalar, _c);
```
chfast, line 78: I wonder what it actually does. Because the later `blst_p1_mult` just uses the bytes. Are they the same bytes as the input?

lib/evmone_precompiles/bls.hpp
```diff
@@ -16,4 +16,10 @@ inline constexpr auto BLS_FIELD_MODULUS =
 [[nodiscard]] bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64],
     const uint8_t _y0[64], const uint8_t _x1[64], const uint8_t _y1[64]) noexcept;
 
+/// Scalar multiplication in BLS12-381 curve group.
```
chfast, line 19: Mention this is in the G1 subgroup and that the subgroup check is performed.

lib/evmone_precompiles/bls.cpp
```diff
@@ -68,4 +70,28 @@ void store_result(uint8_t _rx[64], uint8_t _ry[64], const blst_p1* out) noexcept
...
+
+    const auto p_affine = validate_point(_x, _y);
+    if (!p_affine.has_value())
+        return false;
+
+    blst_p1 p;
+    blst_p1_from_affine(&p, &*p_affine);
+
+    if (!blst_p1_in_g1(&p))
```
chfast, line 87: The subgroup check is also a multiplication. Do you think can we combine both multiplication for better performance?

rodiazet: I would be surprised if they implement group check just by multiplication by group order. There are planty of algorithms which make it faster. I will check the implementation of `blst_p1_in_g1`

rodiazet: There are two different implementations of this check in `blst`. Non of them allows this kind of optimization.

lib/evmone_precompiles/bls.hpp
```diff
@@ -16,4 +16,10 @@ inline constexpr auto BLS_FIELD_MODULUS =
 [[nodiscard]] bool g1_add(uint8_t _rx[64], uint8_t _ry[64], const uint8_t _x0[64],
     const uint8_t _y0[64], const uint8_t _x1[64], const uint8_t _y1[64]) noexcept;
 
+/// Scalar multiplication in BLS12-381 curve G1 subgroup. Point subgroup check is performed.
```
chfast, line 19: I'd put the "Point subgroup check..." in the later paragraph with details docs.

chfast (approved):

chfast (approved):

PR #1000, opened 2024-09-11
test: Export withdrawals to JSON state tests

chfast:


chfast:
I implemented exporting `"withdrawals"` to JSON to fix round-trip JSON testing, but the test still don't pass.

kmk142789 (approved):

PR #1004, opened 2024-09-12
Fix `ExecutionState::reset()`

chfast:
Reset previously omitted fields related to EOF execution: `deploy_container` and `call_stack`.


chfast:
> maybe rearrange the reset statements to reflect order of declarations (also move `call_stack` up?).

oh, I thought they are in order. I'll check.

chfast:
The reset handles the fields in the order the fields are declared in the class. This is as if it is some kind of constructor.
BTW, I wonder if this is some kind of design pattern.

PR #1007, opened 2024-09-13
test: Add a state test reaching CALL max depth

chfast:


pdobacz:
There's such test in EEST already https://github.com/ethereum/execution-spec-tests/blob/main/tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py#L874 (judging by name only, I didn't read the diff)

chfast:
Yes, I just need one example locally.

It actually crashes because of the stack overflow. This is because we don't run EEST with Clang nor MSVC debug builds.

PR #1008, opened 2024-09-16
Refactor Host::execute_message() to allow CALLs with recipient != code_address

gumb0:
This is needed to prepare for EIP-7702 https://github.com/ethereum/evmone/pull/961

The main goal is to get rid of assertion that requires `recipient == code_address` in case `kind == EVMC_CALL`. (In 7702 world CALL can have `code_address` pointing to a different (delegate) address.)

The idea is to split a variable `dst_acc` that was containing either `recepient` or `code_address` account into two separate variables for each thing.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -376,26 +376,29 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
             m_state.journal_create(msg.recipient, exists);
     }
 
-    assert(msg.kind != EVMC_CALL || evmc::address{msg.recipient} == msg.code_address);
-    auto* const dst_acc =
-        (msg.kind == EVMC_CALL) ? &m_state.touch(msg.recipient) : m_state.find(msg.code_address);
-
-    if (msg.kind == EVMC_CALL && !evmc::is_zero(msg.value))
+    if (msg.kind == EVMC_CALL)
     {
-        // Transfer value: sender → recipient.
-        // The sender's balance is already checked therefore the sender account must exist.
-        const auto value = intx::be::load<intx::uint256>(msg.value);
-        assert(m_state.get(msg.sender).balance >= value);
-        m_state.journal_balance_change(msg.sender, m_state.get(msg.sender).balance);
-        m_state.journal_balance_change(msg.recipient, dst_acc->balance);
-        m_state.get(msg.sender).balance -= value;
-        dst_acc->balance += value;
+        auto& dst_acc = m_state.touch(msg.recipient);
+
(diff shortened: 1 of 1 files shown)


test/state/host.cpp
```diff
@@ -376,26 +376,28 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
...
     {
-        // Transfer value: sender → recipient.
-        // The sender's balance is already checked therefore the sender account must exist.
-        const auto value = intx::be::load<intx::uint256>(msg.value);
-        assert(m_state.get(msg.sender).balance >= value);
-        m_state.journal_balance_change(msg.sender, m_state.get(msg.sender).balance);
-        m_state.journal_balance_change(msg.recipient, dst_acc->balance);
-        m_state.get(msg.sender).balance -= value;
-        dst_acc->balance += value;
+        auto& dst_acc = m_state.touch(msg.recipient);
```
chfast, line 381: If we send value we can actually skip touching (because the destination account will have this value or it has code to zero it later). Can you check this hypothesis (all tests passing) or at least leave a comment?

test/state/host.cpp
```diff
@@ -376,26 +376,28 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
...
+            m_state.get(msg.sender).balance -= value;
+            dst_acc.balance += value;
+        }
     }
 
     if (is_precompile(m_rev, msg.code_address))
         return call_precompile(m_rev, msg);
 
-    const auto code = dst_acc != nullptr ? bytes_view{dst_acc->code} : bytes_view{};
+    const auto* const code_acc = m_state.find(msg.code_address);
```
chfast, line 399: In case `msg.recipient` is `msg.code_address` this is unnecessary second lookup. Please leave a comment.

gumb0: Good point, I can aslo try to optimize it / make conditional.

chfast: As you wish, but the current code looks nice and clean...

pdobacz (approved):

test/state/host.cpp
chfast: Are you planning to handle this?

gumb0: I think I'll leave it, seems to get too messy. Will add a comment.

chfast (approved):

test/state/host.cpp
chfast: Good call. This actually disappears in new State API because there are separate methods to get account and code.

PR #1017, opened 2024-09-18
test: Add 2 tests for SELFDESTRUCT

chfast:


pdobacz (approved):
does it make sense to add to EEST as well (or instead)?

chfast:
> does it make sense to add to EEST as well (or instead)?

This is covered in EEST. I just try to cover holes which allowed bugs in evmone.

PR #1027, opened 2024-09-23
test: Ensure exported tests have accessLists for eip1559 txs

pdobacz:
Opening as draft to discuss if I'm understanding things correctly here.

The problem was that our exported test txs were missing `accessLists` field, even though they contained `eip1559` fields. I think making the default state_transition tx be eip1559, and making it explicit wherever we want legacy instead is the solution, but I'm not sure.

Example exported JSON without `accessLists`:

```
    "transaction": {
      "data": [
        "0x"
      ],
      "gasLimit": [
        "0xf4240"
      ],
      "maxFeePerGas": "0x3e8",
      "maxPriorityFeePerGas": "0x3e8",
      "nonce": "0x1",
      "secretKey": "0x00000000000000000000000000000000000000000000000000000002b1263d2b",
      "sender": "0xe100713fc15400d1e94096a545879e7c6407001e",
      "to": "0x000000000000000000000000000000000000c0de",
      "value": [
        "0x0"
      ]
    }
```

from `fixtures/state_tests/state_transition/eof_examples_callf.json`.

Does this make sense?


chfast:
It depends who is complaining about the missing `accessLists`. If only evmone, then we can just relax the parsing rules. All this is likely because we are forced to guess the transaction type by the fields...

pdobacz:
> It depends who is complaining about the missing `accessLists`. If only evmone, then we can just relax the parsing rules. All this is likely because we are forced to guess the transaction type by the fields...

Nope, it's consumers of the exported tests.

pdobacz:
FWIW I diffed the 0.13.0 tests and ones exported from this PR branch and the only difference is the addition of some empty `accessLists` fields in some JSONs

chfast:
Ok, let's go with this. The only alternative I have is to do the tx type guessing during export...

Just add a comment why this is the default transaction type.

test/unittests/state_transition_tx_test.cpp
```diff
@@ -72,6 +75,7 @@ TEST_F(state_transition, empty_coinbase_fee_0_sd)
     block.base_fee = 0;  // should be 0 before London
     tx.max_gas_price = 0;
     tx.max_priority_gas_price = 0;
+    tx.type = Transaction::Type::legacy;
```
chfast, line 78: Put the `tx.type` as the first tx assignment.

chfast (approved):

PR #1030, opened 2024-09-25
test: Move invalid tx messages to state utils

chfast:
Moves the function `get_invalid_tx_message()` function from `evmone::state` (state.cpp) to `evmone::test` (statetest_export.cpp).


rodiazet:
Can you add reasoning for this change?

rodiazet (approved):

chfast:
I'm cleaning up the `state.hpp`/`state.cpp` because we are going to make the `State` a private component. This change moves this single helper to testing utils because it only is related to JSON tests.

Issue #1003, opened 2024-09-12
EOF: Don't copy type section


chfast:
We unnecessarily copy the contents of the type section to `std::vector<EOFCodeType>`. It should be enough to read a type from the type section bytes on demand. E.g. `EOFCodeType get_type(size_t index)`.

abloom22:
I'd like to work on this issue. Could you please assign it to me?

chfast:
@aryeh-bloom this is assigned to you, are you working on this?

aryeh-bloom:
@chfast yes sorry, I was out of town for most of last week and unable to get anything done at my computer

elijahhampton:
@chfast I created a PR for this issue here: https://github.com/ethereum/evmone/pull/1034

PR #1034, opened 2024-09-27
Optimize EOF by reading types on demand

elijahhampton:
Introduces `EOF1Header::get_type` function to read individual EOF
code types. This is fast procedure because we need to read at most 4
bytes and a know offset. By doing so we avoid allocating separated
vector to types.

Closes https://github.com/ethereum/evmone/issues/1003.

### DIFF
--- a/lib/evmone/eof.cpp
+++ b/lib/evmone/eof.cpp
@@ -205,43 +205,43 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_section_headers(byt
     if (remaining_container_size < section_bodies_without_data)
         return EOFValidationError::invalid_section_bodies_size;
 
-    if (section_headers[TYPE_SECTION][0] != section_headers[CODE_SECTION].size() * 4)
+    if (section_headers[TYPE_SECTION][0] !=
+        section_headers[CODE_SECTION].size() * EOF1Header::TYPE_ENTRY_SIZE)
         return EOFValidationError::invalid_type_section_size;
 
     return section_headers;
 }
 
-std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
-    bytes_view container, size_t header_size, uint16_t type_section_size) noexcept
+EOFValidationError validate_types(
+    bytes_view container, size_t type_section_offset, uint16_t type_section_size) noexcept
 {
     assert(!container.empty());  // guaranteed by EOF headers validation
 
-    std::vector<EOFCodeType> types;
 
     // guaranteed by EOF headers validation
-    assert(header_size + type_section_size < container.size());
+    assert(type_section_offset + type_section_size < container.size());
 
-    for (auto offset = header_size; offset < header_size + type_section_size; offset += 4)
+    const auto num_types = type_section_size / EOF1Header::TYPE_ENTRY_SIZE;
+    for (size_t i = 0; i < num_types; ++i)
     {
-        types.emplace_back(
-            container[offset], container[offset + 1], read_uint16_be(&container[offset + 2]));
-    }
+        const auto offset = type_section_offset + (i * EOF1Header::TYPE_ENTRY_SIZE);
+        const auto inputs = container[offset];
+        const auto outputs = container[offset + 1];
+        const auto max_stack_height = read_uint16_be(&container[offset + 2]);
 
-    // check 1st section is (0, 0x80)
-    if (types[0].inputs != 0 || types[0].outputs != NON_RETURNING_FUNCTION)
-        return EOFValidationError::invalid_first_section_type;
+        // First type should be (0, 0x80)
+        if (i == 0 && (inputs != 0 || outputs != NON_RETURNING_FUNCTION))
+            return EOFValidationError::invalid_first_section_type;
 
-    for (const auto& t : types)
-    {
-        if ((t.outputs > OUTPUTS_INPUTS_NUMBER_LIMIT && t.outputs != NON_RETURNING_FUNCTION) ||
-            t.inputs > OUTPUTS_INPUTS_NUMBER_LIMIT)
+        if ((outputs > OUTPUTS_INPUTS_NUMBER_LIMIT && outputs != NON_RETURNING_FUNCTION) ||
+            inputs > OUTPUTS_INPUTS_NUMBER_LIMIT)
             return EOFValidationError::inputs_outputs_num_above_limit;
 
-        if (t.max_stack_height > MAX_STACK_HEIGHT)
+        if (max_stack_height > MAX_STACK_HEIGHT)
             return EOFValidationError::max_stack_height_above_limit;
     }
 
-    return types;
+    return EOFValidationError::success;
 }
 
 /// Result of validating instructions in a code section.
@@ -286,9 +286,11 @@ std::variant<InstructionValidationResult, EOFValidationError> validate_instructi
         else if (op == OP_CALLF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
-            if (fid >= header.types.size())
+            if (fid >= header.code_sizes.size())
                 return EOFValidationError::invalid_code_section_index;
-            if (header.types[fid].outputs == NON_RETURNING_FUNCTION)
+
+            const auto type = header.get_type(container, fid);
+            if (type.outputs == NON_RETURNING_FUNCTION)
                 return EOFValidationError::callf_to_non_returning_function;
             if (code_idx != fid)
                 accessed_code_sections.insert(fid);
@@ -302,10 +304,12 @@ std::variant<InstructionValidationResult, EOFValidationError> validate_instructi
         else if (op == OP_JUMPF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
-            if (fid >= header.types.size())
+            if (fid >= header.code_sizes.size())
                 return EOFValidationError::invalid_code_section_index;
+
+            const auto type = header.get_type(container, fid);
             // JUMPF into returning function means current function is returning.
-            if (header.types[fid].outputs != NON_RETURNING_FUNCTION)
+            if (type.outputs != NON_RETURNING_FUNCTION)
                 is_returning = true;
             if (code_idx != fid)
                 accessed_code_sections.insert(fid);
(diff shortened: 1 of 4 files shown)


elijahhampton:
This PR addresses: https://github.com/ethereum/evmone/issues/1003

lib/evmone/eof.cpp
```diff
@@ -79,6 +79,12 @@ EOFValidationError get_section_missing_error(uint8_t section_id) noexcept
     }
 }
 
+EOFCodeType get_type(bytes_view container, size_t header_size, size_t index)
+{
+    const size_t offset = header_size + index * 4;
+    return {container[offset], container[offset + 1], read_uint16_be(&container[offset + 2])};
```
chfast, line 85: This is good way of reading the `EOFCodeType`.

lib/evmone/eof.cpp
```diff
@@ -220,14 +226,14 @@ std::variant<std::vector<EOFCodeType>, EOFValidationError> validate_types(
...
 
     // guaranteed by EOF headers validation
     assert(header_size + type_section_size < container.size());
 
-    for (auto offset = header_size; offset < header_size + type_section_size; offset += 4)
+    for (size_t i = 0; i < type_section_size / 4; ++i)
     {
-        types.emplace_back(
-            container[offset], container[offset + 1], read_uint16_be(&container[offset + 2]));
+        types.emplace_back(get_type(container, header_size, i));
```
chfast, line 236: However, this is not what was intended. We want to drop `std::vector<EOFCodeType> types`. And we want to replace all uses of  `header.types[fid]` with `header.get_type(container, index)`.

Hint: you will have to pass the container to the `EOF1Header::get_type()` because the header object don't have the reference to the container. We may change this later, but that's separate topic.

chfast (changes requested):

lib/evmone/eof.cpp
elijahhampton: Originally I had the method in the header class passing in the container, but I re read the issue and tried to mimic the suggested function signature.

I'll make the change

chfast: Sorry for the confusion then :)

elijahhampton: @chfast Fixed it

lib/evmone/eof.hpp
```diff
@@ -15,6 +15,7 @@
 
 namespace evmone
 {
+inline uint16_t read_uint16_be(auto it) noexcept;
```
chfast, line 18: Move both of these helper functions here (`read_int16_be` and `read_uint16_be`).

lib/evmone/eof.hpp
```diff
@@ -61,8 +62,25 @@ struct EOF1Header
     std::vector<uint16_t> container_sizes;
     /// Offset of every container section start;
     std::vector<uint16_t> container_offsets;
+    /// Size of every type section.
+    uint16_t type_section_size = 0;
```
chfast, line 66: Put these field after `version` as the type section is the first one.

lib/evmone/instructions.hpp
```diff
@@ -1101,9 +1101,9 @@ inline code_iterator callf(StackTop stack, ExecutionState& state, code_iterator
     const auto index = read_uint16_be(&pos[1]);
     const auto& header = state.analysis.baseline->eof_header();
     const auto stack_size = &stack.top() - state.stack_space.bottom();
+    const EOFCodeType code_type = header.get_type(state.original_code, index);
```
chfast, line 1104:
```suggestion
    const auto callee_type = header.get_type(state.original_code, index);
```

lib/evmone/eof.cpp
```diff
@@ -254,6 +253,7 @@ struct InstructionValidationResult
     std::unordered_set<uint16_t> accessed_code_sections;
 };
 
+
```
chfast, line 256: Avoid adding new empty lines in unrelated code.

lib/evmone/eof.cpp
```diff
@@ -286,9 +287,11 @@ std::variant<InstructionValidationResult, EOFValidationError> validate_instructi
         else if (op == OP_CALLF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
-            if (fid >= header.types.size())
+            if (fid >= header.type_section_size)
                 return EOFValidationError::invalid_code_section_index;
-            if (header.types[fid].outputs == NON_RETURNING_FUNCTION)
+
+            const EOFCodeType code_type = header.get_type(container, fid);
```
chfast, line 293:
```suggestion
            const auto type = header.get_type(container, fid);
```

lib/evmone/eof.cpp
```diff
@@ -302,10 +305,12 @@ std::variant<InstructionValidationResult, EOFValidationError> validate_instructi
         else if (op == OP_JUMPF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
-            if (fid >= header.types.size())
+            if (fid >= header.type_section_size)
                 return EOFValidationError::invalid_code_section_index;
+
+            const EOFCodeType code_type = header.get_type(container, fid);
```
chfast, line 311:
```suggestion
            const auto type = header.get_type(container, fid);
```

lib/evmone/eof.cpp
```diff
@@ -448,37 +455,41 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         if (opcode == OP_CALLF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
+            if (fid >= header.get_type_count())
+                return EOFValidationError::invalid_code_section_index;
 
-            stack_height_required = code_types[fid].inputs;
+            const EOFCodeType code_type = header.get_type(container, fid);
```
chfast, line 461:
```suggestion
            const auto callee_type = header.get_type(container, fid);
```

lib/evmone/eof.cpp
```diff
@@ -427,8 +433,9 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
 
     assert(!code.empty());
 
+    const EOFCodeType code_type_by_func_index = header.get_type(container, func_index);
```
chfast, line 436:
```suggestion
    const auto type = header.get_type(container, func_index);
```

lib/evmone/eof.cpp
```diff
@@ -448,37 +455,41 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
         if (opcode == OP_CALLF)
         {
             const auto fid = read_uint16_be(&code[i + 1]);
+            if (fid >= header.get_type_count())
```
chfast, line 458: This is a new check. Why is it needed now?

elijahhampton: I removed the addition of those checks. Turns out the check I replaced in validate_instructions was not strict enough. Now it is and all test pass without the additional check.

lib/evmone/eof.hpp
```diff
@@ -43,6 +61,11 @@ struct EOF1Header
     /// The EOF version, 0 means legacy code.
     uint8_t version = 0;
 
+    /// Size of every type section.
+    uint16_t type_section_size = 0;
+    /// Offset of type container section start.
```
chfast, line 66:
```suggestion
    /// Offset of type section start.
```

lib/evmone/eof.hpp
```diff
@@ -43,6 +61,11 @@ struct EOF1Header
...
+    size_t type_section_offset = 0;
```
chfast, line 67: I think the offset field should be before the size field.

lib/evmone/eof.hpp
```diff
@@ -62,7 +85,20 @@ struct EOF1Header
     /// Offset of every container section start;
     std::vector<uint16_t> container_offsets;
 
-    std::vector<EOFCodeType> types;
+    /// A helper to extract reference to a specific type section.
+    [[nodiscard]] EOFCodeType get_type(bytes_view container, size_t type_idx) const noexcept
+    {
+        static constexpr size_t TYPE_ENTRY_SIZE = 4;  // Size of each type entry in bytes
```
chfast, line 91:
```suggestion
        static constexpr size_t TYPE_ENTRY_SIZE = sizeof(EOFCodeType);  // Size of each type entry in bytes
```
Put this constant outside of function and use in other places that now uses `4`. E.g. `type_section_size != code_sizes.size() * 4`.

lib/evmone/eof.cpp
```diff
@@ -582,6 +586,7 @@ std::variant<EOFValidationError, int32_t> validate_max_stack_height(
 
     const auto max_stack_height_it = std::max_element(stack_heights.begin(), stack_heights.end(),
         [](StackHeightRange lhs, StackHeightRange rhs) noexcept { return lhs.max < rhs.max; });
+
```
chfast, line 589: Unrelated empty line.

lib/evmone/eof.cpp
```diff
@@ -759,14 +765,17 @@ std::variant<EOF1Header, EOFValidationError> validate_header(
...
-    if (const auto* error = std::get_if<EOFValidationError>(&types_or_error))
-        return *error;
-    const auto& types = std::get<std::vector<EOFCodeType>>(types_or_error);
+    const auto type_section_offset = header_size;
+    const auto type_section_size = section_headers[TYPE_SECTION].front();
+
+    if (type_section_size != code_sizes.size() * 4)
+        return EOFValidationError::invalid_type_section_size;
+
+    auto validation_error = validate_types(container, type_section_offset, type_section_size);
```
chfast, line 774: This is quite an issue: `validate_type()` should not be used in `validate_header()` and invoke separately after `validate_header()`. This way we can just pass `EOF1Header` to `validate_types()` and use its `.get_type()` method.

But maybe this requires a separate PR. What do you think @gumb0?

lib/evmone/eof.hpp
```diff
@@ -43,6 +61,11 @@ struct EOF1Header
     /// The EOF version, 0 means legacy code.
     uint8_t version = 0;
 
+    /// Size of every type section.
+    uint16_t type_section_size = 0;
```
gumb0, line 65: We don't need to store the size at all, because header validation guarantees that `type_section_size == code_sizes.siz() * 4`

chfast (approved):

PR #1049, opened 2024-10-10
statetest: Drop support for "fake" `BLOCKHASH`

chfast:
The state tests had a procedure of generating non-empty BLOCKHASH outputs based on the block number. This is deprecated feature not used in any existing tests. Remove support for it in evmone.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -478,11 +478,7 @@ bytes32 Host::get_block_hash(int64_t block_number) const noexcept
     if (const auto& it = m_block.known_block_hashes.find(block_number);
         it != m_block.known_block_hashes.end())
         return it->second;
-
-    // Convention for testing: if the block hash in unknown return the predefined "fake" value.
-    // https://github.com/ethereum/go-ethereum/blob/v1.12.2/tests/state_test_util.go#L432
-    const auto s = std::to_string(block_number);
-    return keccak256({reinterpret_cast<const uint8_t*>(s.data()), s.size()});
+    return {};
 }
 
 void Host::emit_log(const address& addr, const uint8_t* data, size_t data_size,
--- a/test/unittests/state_transition_block_test.cpp
+++ b/test/unittests/state_transition_block_test.cpp
@@ -32,17 +32,6 @@ TEST_F(state_transition, known_block_hash)
         0x0000000000000000000000000000000000000000000000000000000000000111_bytes32;
 }
 
-TEST_F(state_transition, known_block_hash_fake)
-{
-    block.number = 2;
-    tx.to = To;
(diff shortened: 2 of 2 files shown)


winsvega:
Perhaps error out of invalid test?

chfast:
> Perhaps error out of invalid test?

How exactly?

test/state/host.cpp
```diff
@@ -478,11 +478,7 @@ bytes32 Host::get_block_hash(int64_t block_number) const noexcept
     if (const auto& it = m_block.known_block_hashes.find(block_number);
         it != m_block.known_block_hashes.end())
         return it->second;
-
-    // Convention for testing: if the block hash in unknown return the predefined "fake" value.
-    // https://github.com/ethereum/go-ethereum/blob/v1.12.2/tests/state_test_util.go#L432
-    const auto s = std::to_string(block_number);
-    return keccak256({reinterpret_cast<const uint8_t*>(s.data()), s.size()});
+    return {};
```
gumb0, line 481: The line looks to be covered, so still used somewhere?

chfast:
We may get a protection against this in EEST: https://github.com/ethereum/execution-spec-tests/issues/899.

But it may also break goevmlab fuzzer (cc @holiman). However, they may use `"blockHashes"` to provide the value for `BLOCKHASH(0)`.

test/state/host.cpp
chfast: The coverage comes from ethereum/tests. I'll need to check but IIRC there were some random tests having `BLOCKHASH`.

chfast: This test has it: https://github.com/ethereum/legacytests/blob/master/Constantinople/GeneralStateTests/stRandom/randomStatetest101.json.

gumb0: I guess value is not used in it then

gumb0 (approved):

holiman:
What do you mean "deprecated feature" - since the opcode `BLOCKHASH` is still not deprecated, what is a node supposed to do on that opcode, when executing a statetest?

chfast:
> What do you mean "deprecated feature" - since the opcode `BLOCKHASH` is still not deprecated, what is a node supposed to do on that opcode, when executing a statetest?

State tests try to avoid it at all cost to the point a state test with `BLOCKHASH` is considered invalid. See https://github.com/ethereum/execution-spec-tests/issues/899.

holiman:
I think that is insufficient. Random tests are highly useful. I don't much care what the vm returns, but having all vms behave identically is important. 

Why do this?

chfast:
> Why do this?

It bothers me this is mixed with the semi-production ready code. I'd prefer returning null hash but I think we can find a place for the old behavior...

chfast:
The #1059 moves this feature "more" into "tests only". This should be fine for now.

PR #1051, opened 2024-10-14
Framework for experimenting with jumpdest analysis

chfast:
Add new evmone::experimental library as a place for experimentation with different jumpdest analysis implementations.


test/experimental/jumpdest_analysis.hpp
```diff
@@ -0,0 +1,26 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2020 The evmone Authors.
```
gumb0, line 2: Also exists from 2020? I have seen only reference implementation before.

gumb0 (approved):

test/experimental/jumpdest_analysis.hpp
gumb0: > Also exists since 2020? I have seen only reference implementation before.

chfast: Yes, this is this old.

PR #1052, opened 2024-10-14
eof: Move type validation outside of header validation

chfast:
This allows separate header and type validation and simplify the implementation of the former.


pdobacz (approved):
LGTM. Such refactor was [considered earlier](https://github.com/ethereum/evmone/pull/878#issuecomment-2098375588). I completely don't recall the reasons why this didn't go through back then. Looks all good now :shrug:

chfast:
I think #1034 was a prerequisite. I fuzzed these two PRs.

PR #1061, opened 2024-10-28
test: Split know block hashes into TestBlockHashes

chfast:
This removes `.known_block_hashes` from `state::BlockInfo`. Block hashes in tests are now provided by `TestBlockHashes`.

The main goal here is to remove test-only fields from `state::BlockInfo`.

### DIFF
--- a/test/state/test_state.hpp
+++ b/test/state/test_state.hpp
@@ -72,23 +72,19 @@ public:
     void apply(const state::StateDiff& diff);
 };
 
-class TestBlockHashes : public state::BlockHashes
+class TestBlockHashes : public state::BlockHashes, public std::unordered_map<int64_t, bytes32>
 {
-    /// The map block_number => block hash of known blocks.
-    const std::unordered_map<int64_t, bytes32>& known_block_hashes_;
-
 public:
-    explicit TestBlockHashes(const std::unordered_map<int64_t, bytes32>& known_block_hashes)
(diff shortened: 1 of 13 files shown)


chfast:
Now I think it may be better to introduce `TestEnv` matching JSON `"env"` that aggregates `state::BlockInfo` and implements `state::BlockHashes`.

test/state/test_state.hpp
```diff
@@ -72,31 +72,34 @@ class TestState : public state::StateView, public std::map<address, TestAccount>
     void apply(const state::StateDiff& diff);
 };
 
-class TestBlockHashes : public state::BlockHashes
+class TestBlockHashes : public state::BlockHashes, public std::unordered_map<int64_t, bytes32>
 {
     /// The map block_number => block hash of known blocks.
-    const std::unordered_map<int64_t, bytes32>& known_block_hashes_;
+    // const std::unordered_map<int64_t, bytes32>& known_block_hashes_;
```
pdobacz, line 78: commented lines here and below should be removed instead?

pdobacz (approved):
LGTM, one potential nitpick. TestEnv also sounds good, not sure if better or not. If making this step is not strictly necessary now, maybe defer the decision?

test/state/test_state.hpp
chfast: thx.

PR #1064, opened 2024-10-29
t8n: Include empty requests[Hash] in JSON output

### LINKED ISSUE #1062: T8N updates for Prague Devnet-4 Changes
This is a tracking issue to update the result output from t8n to include the two new fields that are necessary to fill tests after Prague Devnet-4:
- `requests_hash`: Requests hash after execution of all transactions according to EIP-7685: https://eips.ethereum.org/EIPS/eip-7685#block-header
- `requests`: List of byte arrays (of exact length of three elements as of Prague) containing the `request_data` (without `request_type`) of each request type after execution of all transactions.

pdobacz:
Fixes #1062. 

Two open items:
1/ the t8n.cpp code is quite rudimentary, but I don't want to overdo it, considering it's going to get an overhaul when implementing EIPs 7251/6110/7002. This is just enough to satisfy the t8n format required by EIP-7685
2/ I've spent a lot of time trying to find a CMakeLists setup that works. I have no idea if this is proper, I only superficially caught up with cmake docs. Also not sure if the proposed placement of sha256.xpp is OK.


test/t8n/t8n.cpp
```diff
@@ -241,7 +242,38 @@ int main(int argc, const char* argv[])
         if (rev >= EVMC_PRAGUE)
         {
             // EIP-7685: General purpose execution layer requests
-            j_result["requestsRoot"] = hex0x(state::EMPTY_MPT_HASH);
+            j_result["requests"] = json::json::array();
+            // TODO: actual requests should be used in the following lines, for now all empty.
+            const uint8_t withdrawals[1] = {0x00};
```
pdobacz, line 247: :see_no_evil: This is horrible, but let me know if this is good enough for the time being, or if there are any low-hanging cleanups that can be made here. I don't want to overthink it unnecessarily

pdobacz: actually, having typed this, I got a thought that I could just hard code the empty hash, without all the sha256 work.

lib/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# evmone: Fast Ethereum Virtual Machine implementation
+# Copyright 2024 The evmone Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+add_library(crypto STATIC)
+add_library(evmone::crypto ALIAS crypto)
+target_compile_features(crypto PRIVATE cxx_std_20)
+target_link_libraries(crypto PUBLIC evmc::evmc_cpp)
+target_sources(
+    crypto PRIVATE
```
chfast, line 10: I think creating the `evmone::crypto` library may be needed, but I was rather thinking of renaming `evmoe::precompiles` to `evmone::crypto`. Do we need this now?

chfast: I meant renaming `evmone::precompiles` library `evmone::crypto`. For you just add `evmone::precompiles` as the dependency. It depends on it implicitly already.

test/integration/t8n/CMakeLists.txt
```diff
@@ -57,3 +57,38 @@ set_tests_properties(
     FIXTURES_CLEANUP ${TEST_CASE}
     PASS_REGULAR_EXPRESSION ${EXPECTED_OUT_ALLOC}
 )
+
+set(TEST_CASE osaka_empty_requests)
```
chfast, line 61: I'm fine with it, but requests are specific to Prague

chfast (approved):

test/integration/t8n/CMakeLists.txt
pdobacz: ah, of course needs fixing :facepalm: thx

PR #1065, opened 2024-10-29
t8n: Include empty requests[Hash] in JSON output - hardcoded

### LINKED ISSUE #1062: T8N updates for Prague Devnet-4 Changes

pdobacz:
Hardcoded version of #1064 

Fixes #1062 

If the sha256 refactor from #1064 is not useful, we can just roll with this


pdobacz:
close in favor of the original #1064

PR #1067, opened 2024-11-04
cable: external git info

battlmonstr:
#### Problem:

A git commit sha or a branch name change causes relinking of executable targets with silkworm-buildinfo library even if nothing else has changed.

During development (e.g. local Debug builds) it is more important to have fast incremental builds than having the correct git commit sha and a branch name baked into the binary.

#### Solution:

Add arguments to cable_add_buildinfo_library:
* GIT_DESCRIBE
* GIT_BRANCH
* GIT_ORIGIN_URL

If passed they override corresponding git commands.

In Silkworm we can hardcode them in Debug builds to avoid relinking and save time when doing small changes and git rebase/squash/reword operations.

See related Silkworm PR: 
https://github.com/erigontech/silkworm/pull/2473

#### Test

Silkworm `git commit --amend --date=now && time make` output before the change:

```
...
[11/70] Updating silkworm-buildinfo:
       Project Version:  0.1.0-dev+commit.11e1cbbc.dirty (prerelease)
       System Name:      darwin
       System Processor: arm64
       Compiler ID:      appleclang
       Compiler Version: 15.0.0.15000309
       Build Type:       debug
       Git Info:         //11e1cbbc69056575cae00370af1883e5f6504944 (dirty)
       Git Branch:       pr/test
       Git Origin URL:   git@github.com:erigontech/silkworm.git
       Timestamp:        2024-11-04T10:57:47
[68/69] Linking CXX executable cmd/dev/snapshots
make  104.37s user 9.23s system 509% cpu 22.279 total
```

Silkworm `git commit --amend --date=now && time make` output after the change:

```
...
[1/47] CableBuildInfo: updating gitinfo.txt
make  0.08s user 0.07s system 84% cpu 0.176 total
```

#### Note

I've also tried passing GIT_COMMIT_HASH directly to buildinfo.cmake instead of via gitinfo.cmake, but it didn't help, because gitinfo.cmake is regenerating gitinfo.txt, and gitinfo.txt is set as a dependency to buildinfo.c, so even though buildinfo.c contents was not changed, it was "touched" and silkworm-buildinfo library got rebuilt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ Documentation of all notable changes to the **evmone** project.
 The format is based on [Keep a Changelog],
 and this project adheres to [Semantic Versioning].
 
+## [0.14.0] — 2024-XX-XX
+
+### Added
+
+- cable: external git info
+  [#1067](https://github.com/ethereum/evmone/pull/1067)
+
 ## [0.13.0] — 2024-09-23
 
 This release adds BLS precompiles and a system contract for [Prague]
--- a/cmake/cable/CableBuildInfo.cmake
+++ b/cmake/cable/CableBuildInfo.cmake
@@ -22,8 +22,15 @@ include(GNUInstallDirs)
 set(cable_buildinfo_template_dir ${CMAKE_CURRENT_LIST_DIR}/buildinfo)
 
 function(cable_add_buildinfo_library)
-
-    cmake_parse_arguments("" "" PROJECT_NAME;EXPORT "" ${ARGN})
+    set(
+        ARG_NAMES
+        PROJECT_NAME
+        EXPORT
+        GIT_DESCRIBE
+        GIT_BRANCH
+        GIT_ORIGIN_URL
+    )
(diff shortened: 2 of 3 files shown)


chfast (commented):
I don't really have opinions here. I used `git describe` years ago because it seems ideal for this job: single command gets you all information and is based on the git tags as the versioning system. Anyway, I'm not sure it is worth to maintain the cable as the external dependency, so you may consider copying to your project.

Anyway, I can merge these changes if this helps. Just please describe the changes in the CHANGELOG at the beginning of the file.

chfast:
> @chfast yes, `git describe` does the job. The problem is that with all the given parameters it becomes way too strict for our intents and purposes.
> 
> Who maintains Cable? Is it a 3rd-party dependency? I don't mind making it a direct dependency of silkworm, but prefer doing it as a separate change later, because I have a limited time budget for such work.
> 
> I've updated the CHANGELOG ✅

I do. https://github.com/ethereum/cable

PR #1069, opened 2024-11-05
state: Separate transaction validation from transition

chfast:
This change separates transaction validation in
`state::validate_transaction()` from `state::transition()` which previously was also validating the transaction.

Readability goes up in cost of slightly lower efficiency. The change is also good for API because transaction validation usually happens in different place than execution.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -297,7 +297,7 @@ void State::rollback(size_t checkpoint)
 
 /// Validates transaction and computes its execution gas limit (the amount of gas provided to EVM).
 /// @return  Execution gas limit or transaction validation error.
-std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
+std::variant<int64_t, std::error_code> validate_transaction(const StateView& state_view,
     const BlockInfo& block, const Transaction& tx, evmc_revision rev, int64_t block_gas_left,
     int64_t blob_gas_left) noexcept
 {
@@ -350,6 +350,11 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     if (tx.max_gas_price < block.base_fee)
         return make_error_code(FEE_CAP_LESS_THEN_BLOCKS);
 
+    // We need some information about the sender so lookup the account in the state.
+    // TODO: During transaction execution this account will be also needed, so we may pass it along.
+    const auto sender_acc = state_view.get_account(tx.sender).value_or(
+        StateView::Account{.code_hash = Account::EMPTY_CODE_HASH});
+
     if (sender_acc.code_hash != Account::EMPTY_CODE_HASH)
         return make_error_code(SENDER_NOT_EOA);  // Origin must not be a contract (EIP-3607).
 
@@ -369,7 +374,7 @@ std::variant<int64_t, std::error_code> validate_transaction(const Account& sende
     // Compute and check if sender has enough balance for the theoretical maximum transaction cost.
     // Note this is different from tx_max_cost computed with effective gas price later.
     // The computation cannot overflow if done with 512-bit precision.
-    auto max_total_fee = umul(intx::uint256{tx.gas_limit}, tx.max_gas_price);
+    auto max_total_fee = umul(uint256{tx.gas_limit}, tx.max_gas_price);
     max_total_fee += tx.value;
 
     if (tx.type == Transaction::Type::blob)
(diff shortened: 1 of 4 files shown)


chfast:
> LGTM, can we fix one thing up when we're at it, it caught me off guard a bit when revieweing: `blockchaintest_runner.cpp` seems to be the only one to not qualify `test::transition` with `test::`, so at first I thought `state::transition` was meant there and something was not right.
> 
> Can you add the `test::` there?

Rather not now, because this file is not modified here and there are multiple cases like this. Tests should use `evmone::test` and rarely but explicitly `state::`.

test/state/state.cpp
```diff
@@ -415,48 +420,32 @@ StateDiff finalize(const StateView& state_view, evmc_revision rev, const address
     return state.build_diff(rev);
 }
 
-std::variant<TransactionReceipt, std::error_code> transition(const StateView& state_view,
-    const BlockInfo& block, const BlockHashes& block_hashes, const Transaction& tx,
-    evmc_revision rev, evmc::VM& vm, int64_t block_gas_left, int64_t blob_gas_left)
+TransactionReceipt transition(const StateView& state_view, const BlockInfo& block,
+    const BlockHashes& block_hashes, const Transaction& tx, evmc_revision rev, evmc::VM& vm,
```
rodiazet, line 424: I would change `tx` to `validated_tx`

test/state/state.cpp
```diff
@@ -415,48 +420,32 @@ StateDiff finalize(const StateView& state_view, evmc_revision rev, const address
...
-    const auto execution_gas_limit = get<int64_t>(validation_result);
-
     const auto base_fee = (rev >= EVMC_LONDON) ? block.base_fee : 0;
-    assert(tx.max_gas_price >= base_fee);                   // Checked at the front.
-    assert(tx.max_gas_price >= tx.max_priority_gas_price);  // Checked at the front.
+    assert(tx.max_gas_price >= base_fee);                   // Required for valid tx.
+    assert(tx.max_gas_price >= tx.max_priority_gas_price);  // Required for valid tx.
     const auto priority_gas_price =
         std::min(tx.max_priority_gas_price, tx.max_gas_price - base_fee);
     const auto effective_gas_price = base_fee + priority_gas_price;
```
rodiazet, line 439: I would remove the empty line and and add one after assert.

test/state/state.hpp
```diff
@@ -137,11 +137,17 @@ class State
...
+TransactionReceipt transition(const StateView& state, const BlockInfo& block,
+    const BlockHashes& block_hashes, const Transaction& tx, evmc_revision rev, evmc::VM& vm,
+    int64_t execution_gas_limit);
 
-std::variant<int64_t, std::error_code> validate_transaction(const Account& sender_acc,
-    const BlockInfo& block, const Transaction& tx, evmc_revision rev, int64_t block_gas_left,
-    int64_t blob_gas_left) noexcept;
+/// Validate a transaction.
+///
+/// @return  Computed execution gas limit or validation error.
```
rodiazet, line 149: Remove unneeded space after `@return`

test/state/state.cpp
```diff
@@ -415,48 +420,32 @@ StateDiff finalize(const StateView& state_view, evmc_revision rev, const address
...
+    auto& sender_acc = state.get_or_insert(tx.sender);
+    assert(sender_acc.nonce < Account::NonceMax);  // Required for valid tx.
     ++sender_acc.nonce;                            // Bump sender nonce.
 
-    const auto execution_gas_limit = get<int64_t>(validation_result);
-
     const auto base_fee = (rev >= EVMC_LONDON) ? block.base_fee : 0;
-    assert(tx.max_gas_price >= base_fee);                   // Checked at the front.
-    assert(tx.max_gas_price >= tx.max_priority_gas_price);  // Checked at the front.
+    assert(tx.max_gas_price >= base_fee);                   // Required for valid tx.
```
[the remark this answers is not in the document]
chfast: Asserts protect computations happening in the following lines. E.g. this one guarantees that `tx.max_gas_price - base_fee` don't overflow. I think there is room for improvement here, but not in this PR.

rodiazet (changes requested):
Correct. Some nitpicks reported.

test/state/state.cpp
chfast: That's another ~20 changes. I think not, if anything we can try to encode this with a type.

test/state/state.cpp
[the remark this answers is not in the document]
rodiazet: Sorry I was referring the assert one line lower. This one has all both sides known from the beginning of the function.

test/state/state.cpp
chfast: :point_up:

test/state/state.cpp
rodiazet: OK.

rodiazet (approved):

PR #1077, opened 2024-12-03
EIP-4844 - pass blockchain tests

pdobacz:
```
build/bin/evmone-blockchaintest ~/Downloads/fixtures_test_prague_devnet_5/blockchain_tests/cancun/eip4844_blobs/
...
[==========] 43 tests from 7 test suites ran. (4250 ms total)
[  PASSED  ] 43 tests.
```
:green_apple: 

~~NOTE: I'll squash the commits before merging.~~

~~NOTE2: opening as draft to work through CI issues. Not reviewable yet, but if I'm completely off the scent somewhere, please let know~~

Ready for review

### DIFF
--- a/test/blockchaintest/blockchaintest_loader.cpp
+++ b/test/blockchaintest/blockchaintest_loader.cpp
@@ -18,6 +18,13 @@ T load_if_exists(const json::json& j, std::string_view key)
         return from_json<T>(*it);
     return {};
 }
+template <typename T>
+std::optional<T> load_optional(const json::json& j, std::string_view key)
+{
+    if (const auto it = j.find(key); it != j.end())
+        return std::optional(from_json<T>(*it));
+    return std::nullopt;
+}
 }  // namespace
 
 template <>
@@ -41,7 +48,8 @@ BlockHeader from_json<BlockHeader>(const json::json& j)
         .transactions_root = from_json<hash256>(j.at("transactionsTrie")),
         .withdrawal_root = load_if_exists<hash256>(j, "withdrawalsRoot"),
         .parent_beacon_block_root = load_if_exists<hash256>(j, "parentBeaconBlockRoot"),
-        .excess_blob_gas = load_if_exists<uint64_t>(j, "excessBlobGas"),
+        .excess_blob_gas = load_optional<uint64_t>(j, "excessBlobGas"),
+        .blob_gas_used = load_optional<uint64_t>(j, "blobGasUsed"),
         .requests_hash = load_if_exists<hash256>(j, "requestsHash"),
     };
 }
@@ -62,8 +70,8 @@ static TestBlock load_test_block(const json::json& j, evmc_revision rev)
         tb.block_info.prev_randao = tb.expected_block_header.prev_randao;
         tb.block_info.base_fee = tb.expected_block_header.base_fee_per_gas;
         tb.block_info.parent_beacon_block_root = tb.expected_block_header.parent_beacon_block_root;
-        tb.block_info.blob_base_fee =
-            compute_blob_gas_price(tb.expected_block_header.excess_blob_gas);
+        tb.block_info.excess_blob_gas = tb.expected_block_header.excess_blob_gas;
+        tb.block_info.blob_gas_used = tb.expected_block_header.blob_gas_used;
 
         // Override prev_randao with difficulty pre-Merge
         if (rev < EVMC_PARIS)
--- a/test/blockchaintest/blockchaintest_loader.cpp
+++ b/test/blockchaintest/blockchaintest_loader.cpp
@@ -48,8 +48,8 @@ BlockHeader from_json<BlockHeader>(const json::json& j)
         .transactions_root = from_json<hash256>(j.at("transactionsTrie")),
         .withdrawal_root = load_if_exists<hash256>(j, "withdrawalsRoot"),
         .parent_beacon_block_root = load_if_exists<hash256>(j, "parentBeaconBlockRoot"),
-        .excess_blob_gas = load_optional<uint64_t>(j, "excessBlobGas"),
         .blob_gas_used = load_optional<uint64_t>(j, "blobGasUsed"),
+        .excess_blob_gas = load_optional<uint64_t>(j, "excessBlobGas"),
         .requests_hash = load_if_exists<hash256>(j, "requestsHash"),
     };
 }
@@ -70,8 +70,13 @@ static TestBlock load_test_block(const json::json& j, evmc_revision rev)
         tb.block_info.prev_randao = tb.expected_block_header.prev_randao;
         tb.block_info.base_fee = tb.expected_block_header.base_fee_per_gas;
         tb.block_info.parent_beacon_block_root = tb.expected_block_header.parent_beacon_block_root;
-        tb.block_info.excess_blob_gas = tb.expected_block_header.excess_blob_gas;
         tb.block_info.blob_gas_used = tb.expected_block_header.blob_gas_used;
+        tb.block_info.excess_blob_gas = tb.expected_block_header.excess_blob_gas;
+
+        tb.block_info.blob_base_fee = tb.block_info.excess_blob_gas.has_value() ?
+                                          std::optional(state::compute_blob_gas_price(
+                                              tb.block_info.excess_blob_gas.value())) :
+                                          std::nullopt;
 
         // Override prev_randao with difficulty pre-Merge
         if (rev < EVMC_PARIS)
--- a/test/blockchaintest/blockchaintest_loader.cpp
+++ b/test/blockchaintest/blockchaintest_loader.cpp
@@ -22,7 +22,7 @@ template <typename T>
 std::optional<T> load_optional(const json::json& j, std::string_view key)
 {
     if (const auto it = j.find(key); it != j.end())
-        return std::optional(from_json<T>(*it));
+        return from_json<T>(*it);
     return std::nullopt;
 }
 }  // namespace
@@ -73,10 +73,10 @@ static TestBlock load_test_block(const json::json& j, evmc_revision rev)
         tb.block_info.blob_gas_used = tb.expected_block_header.blob_gas_used;
         tb.block_info.excess_blob_gas = tb.expected_block_header.excess_blob_gas;
 
-        tb.block_info.blob_base_fee = tb.block_info.excess_blob_gas.has_value() ?
-                                          std::optional(state::compute_blob_gas_price(
-                                              tb.block_info.excess_blob_gas.value())) :
-                                          std::nullopt;
+        tb.block_info.blob_base_fee =
+            tb.block_info.excess_blob_gas.has_value() ?
+                std::optional(state::compute_blob_gas_price(*tb.block_info.excess_blob_gas)) :
+                std::nullopt;
 
         // Override prev_randao with difficulty pre-Merge
         if (rev < EVMC_PARIS)
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -8,6 +8,7 @@
 #include "../state/system_contracts.hpp"
 #include "../test/statetest/statetest.hpp"
 #include "blockchaintest.hpp"
+#include <evmone_precompiles/kzg.hpp>
 #include <gtest/gtest.h>
 
 namespace evmone::test
@@ -42,7 +43,7 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
 
     std::vector<state::Log> txs_logs;
     int64_t block_gas_left = block.gas_limit;
-    int64_t blob_gas_left = 0;
+    auto blob_gas_left = static_cast<int64_t>(block.blob_gas_used.value_or(0));
 
     std::vector<RejectedTransaction> rejected_txs;
     std::vector<state::TransactionReceipt> receipts;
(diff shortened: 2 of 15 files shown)


test/state/block.hpp
```diff
@@ -51,11 +51,9 @@ struct BlockInfo
 
     /// The "excess blob gas" parameter from EIP-4844
     /// for computing the blob gas price in the current block.
-    uint64_t excess_blob_gas = 0;
+    std::optional<uint64_t> excess_blob_gas;
 
-    /// The blob gas price parameter from EIP-4844.
-    /// This value is not stored in block headers directly but computed from excess_blob_gas.
-    intx::uint256 blob_base_fee = 0;
+    std::optional<uint64_t> blob_gas_used;
```
chfast, line 56: Add doc comment.
Also place before `excess_blob_gas` to keep the spec order from https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md#header-extension.

test/state/block.hpp
```diff
@@ -64,6 +62,10 @@ struct BlockInfo
 /// Computes the current blob gas price based on the excess blob gas.
 intx::uint256 compute_blob_gas_price(uint64_t excess_blob_gas) noexcept;
 
+/// Computes the current excess blob gas based on the parent.
+uint64_t calc_excess_blob_gas(
+    uint64_t parent_excess_blob_gas, uint64_t parent_blob_gas_used) noexcept;
```
chfast, line 67: Also swap parameters?

test/state/host.cpp
```diff
@@ -493,7 +493,7 @@ evmc_tx_context Host::get_tx_context() const noexcept
         m_block.prev_randao,
         0x01_bytes32,  // Chain ID is expected to be 1.
         uint256be{m_block.base_fee},
-        intx::be::store<uint256be>(m_block.blob_base_fee),
+        intx::be::store<uint256be>(compute_blob_gas_price(m_block.excess_blob_gas.value_or(0))),
```
chfast, line 496: I think we should not compute this each access. Maybe we need two sets of fields in `BlockInfo`: official/header fields and computed values.

If big deal, we can handle this by separate PR.

test/statetest/statetest_loader.cpp
```diff
@@ -300,7 +297,8 @@ TestState from_json<TestState>(const json::json& j)
 /// Load common parts of Transaction or TestMultiTransaction.
 static void from_json_tx_common(const json::json& j, state::Transaction& o)
 {
-    o.sender = from_json<evmc::address>(j.at("sender"));
+    // Invalid block tests have only v, r, s, no sender.
```
chfast, line 300:
```suggestion
    // Sender is not resolved in transactions from invalid blocks.
```

pdobacz: Are you attached to the particular word "resolved". I would rather see "included" or "present"?

test/state/host.cpp
pdobacz: Oh right, get_tx_context() is called within the instruction, yeah, better to fix right away :(. Alternatively this can be kept in Host? WDYT? if we can avoid mixing header/computed values, maybe it's better? or is Host just inappropriate for this?

test/statetest/statetest_loader.cpp
pdobacz: `recovered` is equally (or even more) confusing as resolved. We're looking at test JSON files here, and sometimes `sender` is just not provided. Sentence with "recovered" could wrongly imply that for tests with valid blocks it is recovered during JSON reading or sth.

chfast: "provided" is fine to me.

test/state/host.cpp
chfast: Better to put it back to `BlockInfo` because this is also used in transaction validation and transition. These are also important cases.

pdobacz: argh, I've done it and only after I recalled from previous week, that I have already been there. The problem which arises is that it makes `compute_blob_gas_price` come before block validation, making problematic inputs (excess_blob_gas) come through, which previously wasn't the case. `fake_exponential` seems to go into infinite loop for such inputs.

I'll need to revisit this and look more closely at what exactly is happening

chfast: I imagine we keep the newly added field and also have `blob_base_fee` "after" them computed as some point...

pdobacz: I explored this idea, but I'm struggling to find a good point to compute that. It would sacrifice some `const`s and make a mess, or force us to introduce more complexity to manage that.

I have a counter-proposal. I dug deeper into `fake_exponential`, and it seems to me like not only it infinite loops but also easily overflows over 256 bits. If I cap the result at uint256::max on mul overflow (I can post code in a moment), things work OK. I think the tacit assumption in `fake_exponential` spec is that it should use arbitrary precision, and I peeked at some impls and they actually do that.

So I'd rather fix `fake_exponential` to not loop/overflow, and rely on current block validation, which would reject the block properly. 

I see three options here:
1. switch from uint256 to big int of some sort
2. detect overflow and cap `compute_blob_gas_price` to uint256::max
3. detect overflow and return nullopt from `compute_blob_gas_price`

2 is the easiest (will push in a sec), but has the disadvantage of there being a block with potentially incorrect blob_base_fee field, hanging around until we reject it (on validation of excess/used blob gas fields).

3 seems better but not sure if the opts won't proliferate around the code base excessively. Maybe not

pdobacz:
@chfast I've almost forgotten about this one :sweat_smile: . Rebased and tests seem to pass, ready to review

test/blockchaintest/blockchaintest_loader.cpp
```diff
@@ -18,6 +18,13 @@ T load_if_exists(const json::json& j, std::string_view key)
         return from_json<T>(*it);
     return {};
 }
+template <typename T>
+std::optional<T> load_optional(const json::json& j, std::string_view key)
+{
+    if (const auto it = j.find(key); it != j.end())
+        return std::optional(from_json<T>(*it));
```
chfast, line 25:
```suggestion
        returnfrom_json<T>(*it);
```

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -95,10 +96,33 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
...
 
-bool validate_block(evmc_revision, const TestBlock&, const BlockHeader&) noexcept
+bool validate_block(
+    evmc_revision rev, const TestBlock& test_block, const BlockHeader& parent_header) noexcept
 {
     // NOTE: includes only block validity unrelated to individual txs. See `apply_block`.
 
+    if (rev >= EVMC_CANCUN)
+    {
+        if (!test_block.block_info.excess_blob_gas.has_value() or
```
chfast, line 106:
```suggestion
        if (!test_block.block_info.excess_blob_gas.has_value() ||
```

chfast, line 106: Add comment that they are mandatory.

test/state/block.hpp
```diff
@@ -49,13 +49,15 @@ struct BlockInfo
...
+
     /// The "excess blob gas" parameter from EIP-4844
     /// for computing the blob gas price in the current block.
-    uint64_t excess_blob_gas = 0;
+    std::optional<uint64_t> excess_blob_gas;
 
-    /// The blob gas price parameter from EIP-4844.
-    /// This value is not stored in block headers directly but computed from excess_blob_gas.
-    intx::uint256 blob_base_fee = 0;
+    /// Blob gas price from EIP-4844, computed from excess_blob_gas
```
chfast, line 59:
```suggestion
    /// Blob gas price from EIP-4844, computed from excess_blob_gas.
```

test/state/block.hpp
```diff
@@ -64,6 +66,10 @@ struct BlockInfo
 /// Computes the current blob gas price based on the excess blob gas.
 intx::uint256 compute_blob_gas_price(uint64_t excess_blob_gas) noexcept;
 
+/// Computes the current excess blob gas based on the parent.
```
chfast, line 69: "on the parent"? Do you mean from the information of the parent block?

test/state/state.cpp
```diff
@@ -311,7 +311,7 @@ std::variant<int64_t, std::error_code> validate_transaction(const StateView& sta
         if (tx.blob_hashes.empty())
             return make_error_code(EMPTY_BLOB_HASHES_LIST);
 
-        if (tx.max_blob_gas_price < block.blob_base_fee)
+        if (tx.max_blob_gas_price < block.blob_base_fee.value())
```
chfast, line 314:
```suggestion
        if (tx.max_blob_gas_price < *block.blob_base_fee)
```

test/state/state.cpp
```diff
@@ -444,7 +444,9 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
 
     if (tx.type == Transaction::Type::blob)
     {
-        const auto blob_fee = tx.blob_gas_used() * block.blob_base_fee;
+        // This uint64 * uint256 cannot overflow, because tx.blob_gas_used has limits enforced
+        // before this stage.
+        const auto blob_fee = tx.blob_gas_used() * block.blob_base_fee.value();
```
chfast, line 449:
```suggestion
        const auto blob_fee = tx.blob_gas_used() * *block.blob_base_fee;
```

test/t8n/t8n.cpp
```diff
@@ -246,9 +246,10 @@ int main(int argc, const char* argv[])
         j_result["gasUsed"] = hex0x(cumulative_gas_used);
         if (rev >= EVMC_CANCUN)
         {
-            j_result["blobGasUsed"] =
-                hex0x(state::BlockInfo::MAX_BLOB_GAS_PER_BLOCK - blob_gas_left);
-            j_result["currentExcessBlobGas"] = hex0x(block.excess_blob_gas);
+            j_result["blobGasUsed"] = hex0x(
+                static_cast<int64_t>(state::BlockInfo::MAX_BLOB_GAS_PER_BLOCK) - blob_gas_left);
+            if (block.excess_blob_gas.has_value())
+                j_result["currentExcessBlobGas"] = hex0x(block.excess_blob_gas.value());
```
chfast, line 252:
```suggestion
                j_result["currentExcessBlobGas"] = hex0x(*block.excess_blob_gas);
```

test/state/state.cpp
rodiazet: I would also add assertion checking that `block.block_base_fee.has_value()`

test/state/state.cpp
```diff
@@ -444,7 +444,9 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
 
     if (tx.type == Transaction::Type::blob)
     {
-        const auto blob_fee = tx.blob_gas_used() * block.blob_base_fee;
+        // This uint64 * uint256 cannot overflow, because tx.blob_gas_used has limits enforced
```
rodiazet, line 447: I would add assertion checking this.

test/state/state.cpp
rodiazet: I would add assertion checking that `block.blob_base_fee.ahs_value()`

test/t8n/t8n.cpp
rodiazet: Add consider assertion here too.

rodiazet (commented):
OK in general. Some nitpicks added. Most important ones with not throwing access to optional already mentioned by @chfast

test/state/block.hpp
pdobacz: yeah, mental shortcut

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -95,10 +96,33 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
...
+                parent_header.blob_gas_used.value_or(0), parent_header.excess_blob_gas.value_or(0)))
+            return false;
+
+        // Ensure the total blob gas spent is at most equal to the limit
+        if (*test_block.block_info.blob_gas_used > state::BlockInfo::MAX_BLOB_GAS_PER_BLOCK)
+            return false;
+    }
+    else
+    {
+        if (test_block.block_info.excess_blob_gas.has_value() or
```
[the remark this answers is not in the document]
rodiazet: I mean `or -> ||`

test/t8n/t8n.cpp
pdobacz: it is already in an if checking value

test/blockchaintest/blockchaintest_loader.cpp
```diff
@@ -62,8 +70,13 @@ static TestBlock load_test_block(const json::json& j, evmc_revision rev)
...
         tb.block_info.base_fee = tb.expected_block_header.base_fee_per_gas;
         tb.block_info.parent_beacon_block_root = tb.expected_block_header.parent_beacon_block_root;
-        tb.block_info.blob_base_fee =
-            compute_blob_gas_price(tb.expected_block_header.excess_blob_gas);
+        tb.block_info.blob_gas_used = tb.expected_block_header.blob_gas_used;
+        tb.block_info.excess_blob_gas = tb.expected_block_header.excess_blob_gas;
+
+        tb.block_info.blob_base_fee = tb.block_info.excess_blob_gas.has_value() ?
+                                          std::optional(state::compute_blob_gas_price(
+                                              tb.block_info.excess_blob_gas.value())) :
```
rodiazet, line 78:
```suggestion
                                              *tb.block_info.excess_blob_gas)) :
```

rodiazet (changes requested):
One last nit.

test/blockchaintest/blockchaintest_loader.cpp
pdobacz: thx, I grepped 2 more occurrences

rodiazet (approved):

chfast (approved):

PR #1079, opened 2024-12-04
EIP-6110: Supply validator deposits on chain

gumb0:
https://eips.ethereum.org/EIPS/eip-6110

### DIFF
--- a/test/state/requests.cpp
+++ b/test/state/requests.cpp
@@ -7,6 +7,12 @@
 
 namespace evmone::state
 {
+namespace
+{
+/// The address of the deposit contract.
+constexpr auto DEPOSIT_CONTRACT_ADDRESS = 0x00000000219ab540356cBB839Cbe05303d7705Fa_address;
+}  // namespace
+
 hash256 calculate_requests_hash(std::span<const Requests> block_requests_list)
 {
     bytes requests_hash_list;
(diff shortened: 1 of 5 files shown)


test/state/requests.hpp
```diff
@@ -37,9 +38,15 @@ struct Requests
     Type type = Type::deposit;
     /// Requests data - an opaque byte array, contains zero or more encoded request objects.
     evmc::bytes data;
+
+    explicit Requests(Type _type, evmc::bytes _data = {}) noexcept
```
gumb0, line 42: Needed to make `vector::emplace_back()` work for this struct on some compilers.

chfast: I use other workaround: `emplace_back(Requests{...})`. I wonder which pattern is easier to find later...

test/state/requests.cpp
```diff
@@ -30,4 +36,56 @@ hash256 calculate_requests_hash(std::span<const Requests> block_requests_list)
...
+                //     bytes pubkey,
+                //     bytes withdrawal_credentials,
+                //     bytes amount,
+                //     bytes signature,
+                //     bytes index
+                // );
+                //
+                // In ABI every bytes array data is prepended by a word with its size.
+                // Skip over first 5 words (offsets of the values) and over pubkey size.
+                constexpr auto pubkey_offset = 32 * 5 + 32;
```
chfast, line 61:
```suggestion
                static constexpr auto PUBKEY_OFFSET = 32 * 5 + 32;
```

test/state/requests.hpp
gumb0: Usually me too, but this time I think I got error from clang-tidy "unnecessary temporary object created"

test/state/requests.cpp
chfast: upper case?

chfast (approved):

pdobacz (approved):

PR #1083, opened 2024-12-09
EIP-7685: General purpose execution layer requests

gumb0:
https://eips.ethereum.org/EIPS/eip-7685

- Introduce `Requests` type.
- Calculate hash of the list of requests after transactions execution. 
- Read expected hash from blockchain test files and compare with result of calculation.
- Use requests hash calculation function in t8n.

(EIP-7685 tests are not enabled yet, because they require all of EIP-6110, 7002 and 7251 to be implemented.)


test/state/requests.cpp
```diff
@@ -0,0 +1,33 @@
...
+    bytes requests_hash_list;
+    requests_hash_list.reserve(sizeof(bytes32) * block_requests_list.size());
+
+    for (const auto& requests : block_requests_list)
+    {
+        // TODO recent change in the spec, uncomment for devnet-5
+        //        if (requests.data.empty())
+        //            continue;
+
+        const bytes requests_bytes = requests.type + requests.data;
```
gumb0, line 21: This can be optimized to get rid of copying, if we represent `Requests` as just `struct Requests { bytes data; }` and put type as the first byte of `data` right away when collecting requests.

Let me know if it's worth it. I think not really, as it only affects blockchain test runner performance.

test/state/requests.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <span>
+#include <vector>
+
+namespace evmone::state
+{
+/// `requests` object.
+struct Requests
+{
+    /// Request type.
+    uint8_t type = 0;
```
pdobacz, line 18: is having a default good here?

pdobacz: also, maybe introduce a Request::Type enum like for tx?

pdobacz (commented):
LGTM overall, I left some minor comments. Also, maybe we're ready to undo the workarounds from https://github.com/ethereum/evmone/pull/1064/files ?

test/state/requests.hpp
gumb0: I was meaning to just introduce constants for types in respective later PRs, but enum is a better idea, I'll add it in the next PR.

Having a default in general is good, otherwise it may be left uninitialized.

gumb0:
> Also, maybe we're ready to undo the workarounds from https://github.com/ethereum/evmone/pull/1064/files ?

Ah I forgot about t8n, thanks for the reminder. I'll look into it, it probably will go to a separate PR, too.

test/state/requests.hpp
pdobacz: > Having a default in general is good, otherwise it may be left uninitialized.

ah right, that would not error out :(. OK then

pdobacz:
> Ah I forgot about t8n, thanks for the reminder. I'll look into it, it probably will go to a separate PR, too.

OK, but also maybe this in particular is better off here to make the change more atomic? Also, t8n has an integration test, so your code would also get automatically covered by that for extra assurance.

pdobacz (approved):

test/state/requests.hpp
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include "hash_utils.hpp"
+#include <evmc/evmc.hpp>
+#include <span>
+#include <vector>
+
+namespace evmone::state
+{
+/// `requests` object.
```
chfast, line 14: Link to the spec?

test/state/requests.hpp
```diff
@@ -0,0 +1,29 @@
...
+    uint8_t type = 0;
+    /// Request data - an opaque byte array, contains zero or more encoded request objects.
+    evmc::bytes data;
+};
+
+/// Block requests list - container of `requests` objects, ordered by request type.
+using RequestsList = std::vector<Requests>;
+
+/// Calculate commitment value of block requests list
+hash256 calculate_requests_hash(std::span<const Requests> block_requests_list);
```
chfast, line 27:
```suggestion
hash256 calculate_requests_hash(std::span<const Requests> requests_list);
```

test/state/requests.hpp
```diff
@@ -0,0 +1,29 @@
...
+struct Requests
+{
+    /// Request type.
+    uint8_t type = 0;
+    /// Request data - an opaque byte array, contains zero or more encoded request objects.
+    evmc::bytes data;
+};
+
+/// Block requests list - container of `requests` objects, ordered by request type.
+using RequestsList = std::vector<Requests>;
```
chfast, line 24: I'd skip this alias because we also use `std::span<Requests>`.

test/state/requests.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include "requests.hpp"
+#include <evmone_precompiles/sha256.hpp>
+
+namespace evmone::state
+{
+hash256 calculate_requests_hash(std::span<const Requests> block_requests_list)
+{
+    bytes requests_hash_list;
+    requests_hash_list.reserve(sizeof(bytes32) * block_requests_list.size());
```
chfast, line 13:
```suggestion
    requests_hash_list.reserve(sizeof(hash256) * block_requests_list.size());
```

test/state/requests.cpp
chfast: I think this is worth it when all EIPs are implemented. So let's make an issue to track it.

test/state/requests.hpp
gumb0: added enum here

test/state/requests.hpp
gumb0: Added

test/t8n/t8n.cpp
```diff
@@ -5,6 +5,7 @@
 #include "../state/errors.hpp"
 #include "../state/ethash_difficulty.hpp"
 #include "../state/mpt_hash.hpp"
+#include "../state/requests.hpp"
```
pdobacz, line 8: can drop sha256.hpp :point_down: and (I think so) the changes in  test/t8n/CMakeLists.txt that allowed the include

pdobacz (approved):

chfast (approved):

test/state/requests.cpp
gumb0: > I think this is worth it when all EIPs are implemented. So let's make an issue to track it.

https://github.com/ethereum/evmone/issues/1086

gumb0: > BTW, is there rationale for this double-hashing?

Good question, I couldn't find any in the EIP, other than previously it was a root of MPT and now is hash of flat list of hashes.
Asked on magicians https://ethereum-magicians.org/t/eip-7685-general-purpose-execution-layer-requests/19668/20

PR #1084, opened 2024-12-10
EIP-7002 and EIP-7251: withdrawal and consolidation requests

gumb0:


test/state/system_contracts.hpp
```diff
@@ -18,15 +19,29 @@ constexpr auto BEACON_ROOTS_ADDRESS = 0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02
...
 /// Executes code of pre-defined accounts via pseudo-transaction from the system sender (0xff...fe).
 /// The sender's nonce is not increased.
-[[nodiscard]] StateDiff system_call(const StateView& state_view, const BlockInfo& block,
+[[nodiscard]] StateDiff system_call_block_start(const StateView& state_view, const BlockInfo& block,
     const BlockHashes& block_hashes, evmc_revision rev, evmc::VM& vm);
+
+/// Performs the system call: invokes system contracts at the end of the block.
+///
+/// Executes code of pre-defined accounts via pseudo-transaction from the system sender (0xff...fe).
+/// The sender's nonce is not increased.
```
pdobacz, line 43: add a 1-sentence summary explanation why these are separate. Judging by just the API, they wouldn't need to be separate, because Requests vector could be returned empty, so I think leaving a hint why is worthwhile.

pdobacz: I didn't know why not have one function with just empty requests returned at block start without looking inside the implementation and comparing. Actually, I'm not 100% sure even now :thinking: why?

gumb0: It's different contracts called at the start and end of block...

test/state/system_contracts.cpp
```diff
@@ -72,4 +94,44 @@ StateDiff system_call(const StateView& state_view, const BlockInfo& block,
     // TODO: Should we return empty diff if no system contracts?
     return state.build_diff(rev);
 }
+
+std::pair<StateDiff, std::vector<Requests>> system_call_block_end(const StateView& state_view,
+    const BlockInfo& block, const state::BlockHashes& block_hashes, evmc_revision rev, evmc::VM& vm)
+{
+    State state{state_view};
+    std::vector<Requests> requests;
+    for (const auto& [since, addr, get_input, request_type] : SYSTEM_CONTRACTS_BLOCK_END)
```
gumb0, line 103: Actually maybe it makes senst not to duplicate the loop code, and reuse the same, I'll try it

test/state/system_contracts.hpp
pdobacz: > It's different contracts called at the start and end of block...

oh, lol, I hallucinated that the contracts list is an input to the function, but yea it's hardcoded... anyhow, a rephrasing could help I guess, maybe sth like `invokes _the_ system contracts _configured to run_ at the start of the block`. Same as the refactor you've mentioned elsewhere, let's see...

test/state/system_contracts.cpp
```diff
@@ -51,25 +67,52 @@ StateDiff system_call_block_start(const StateView& state_view, const BlockInfo&
...
+StateDiff system_call_block_start(const StateView& state_view, const BlockInfo& block,
+    const BlockHashes& block_hashes, evmc_revision rev, evmc::VM& vm)
+{
+    // No requests are generated in block start system calls.
+    const auto [state_diff, _] =
+        system_call(SYSTEM_CONTRACTS_BLOCK_START, state_view, block, block_hashes, rev, vm);
+    return state_diff;
+}
+
+std::pair<StateDiff, std::vector<Requests>> system_call_block_end(const StateView& state_view,
```
chfast, line 119: Does the EIP-7002 make changes to the state?

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -77,11 +78,16 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
...
 
     auto requests =
-        (rev >= EVMC_PRAGUE ? std::vector{collect_deposit_requests(receipts),
-                                  state::Requests(state::Requests::Type::withdrawal),
-                                  state::Requests(state::Requests::Type::consolidation)} :
+        (rev >= EVMC_PRAGUE ? std::vector<state::Requests>{collect_deposit_requests(receipts)} :
                               std::vector<state::Requests>{});
 
+    auto system_call_requests = system_call_block_end(state, block, block_hashes, rev, vm);
+    std::move(
```
chfast, line 85: Try `std::ranges::move()`.

test/state/system_contracts.cpp
gumb0: Yes it does, and 7251 too.

For both there is some kind of queue of requests stored in a contract, and system call dequeues items.

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -77,11 +78,16 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
     }
 
     auto requests =
-        (rev >= EVMC_PRAGUE ? std::vector{collect_deposit_requests(receipts),
-                                  state::Requests(state::Requests::Type::withdrawal),
-                                  state::Requests(state::Requests::Type::consolidation)} :
+        (rev >= EVMC_PRAGUE ? std::vector<state::Requests>{collect_deposit_requests(receipts)} :
```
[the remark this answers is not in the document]
gumb0: True, this will be simpler with `if`

test/state/system_contracts.cpp
```diff
@@ -18,10 +18,13 @@ struct SystemContract
     evmc_revision since = EVMC_MAX_REVISION;  ///< EVM revision in which added.
     address addr;                             ///< Address of the system contract.
     GetInputFn* get_input = nullptr;          ///< How to get the input for the system call.
+    /// Type of requests returned by block end system call.
+    /// Ignored for block start system contracts.
+    Requests::Type request_type = Requests::Type::deposit;
```
chfast, line 23: This abstraction is quite weak because we have disjoint set of objects with `get_input` and `requires_type`. Fine for me for now though.

test/state/system_contracts.cpp
```diff
@@ -51,25 +73,52 @@ StateDiff system_call(const StateView& state_view, const BlockInfo& block,
...
             .sender = SYSTEM_ADDRESS,
-            .input_data = input.bytes,
-            .input_size = std::size(input.bytes),
+            .input_data = input.data(),
+            .input_size = input.size(),
         };
 
         const Transaction empty_tx{};
         Host host{rev, vm, state, block, block_hashes, empty_tx};
         [[maybe_unused]] const auto res = vm.execute(host, rev, msg, code.data(), code.size());
```
chfast, line 100:
```suggestion
        const auto res = vm.execute(host, rev, msg, code.data(), code.size());
```

chfast (approved):
Maybe squash commits related to EIP-7251. They are tiny.

test/state/system_contracts.hpp
pdobacz: The changes addressed my concerns, thank you!

pdobacz (approved):

PR #1088, opened 2024-12-17
tests: Enable invalid block blockchain tests

pdobacz:
I took the liberty to coarsely disable 4844 (and 4788) blockchain tests across the board, as they will be re-enabled in #1077, which needs to be rebased on top of these changes (invalid block blockchain tests handling in that branch is currently not complete!).

Before merging I'll squash the 1st and last commits together, they aim at the same thing.

Note, that a next step would be to respect the `expectException` field from the test and only accept specific reasons for block invalidity. For now I only handle valid-invalid.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -361,8 +361,11 @@ jobs:
       - run:
           name: "Execution spec tests (blockchain_tests)"
           working_directory: ~/build
+          # 4844 and 4788 temporarily turned off b/c they rely on blob features.
           command: >
-            bin/evmone-blockchaintest ~/spec-tests/fixtures/blockchain_tests
+            bin/evmone-blockchaintest
+            --gtest_filter='*:-*eip4844*:*eip4788*'
+            ~/spec-tests/fixtures/blockchain_tests
       - download_execution_spec_tests:
           release: pectra-devnet-4@v1.0.1
           fixtures_suffix: pectra-devnet-4
@@ -402,8 +405,10 @@ jobs:
       - run:
           name: "EOF pre-release execution spec tests (blockchain_tests)"
           working_directory: ~/build
+          # 4844 and 4788 temporarily turned off b/c they rely on blob features.
           command: >
             bin/evmone-blockchaintest ~/spec-tests/fixtures/blockchain_tests/osaka
+            --gtest_filter='*:-*stEIP4844*'
       - run:
           name: "EOF pre-release execution spec tests (eof_tests)"
           working_directory: ~/build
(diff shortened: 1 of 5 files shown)


circle.yml
```diff
@@ -362,7 +362,9 @@ jobs:
           name: "Execution spec tests (blockchain_tests)"
           working_directory: ~/build
           command: >
-            bin/evmone-blockchaintest ~/spec-tests/fixtures/blockchain_tests
+            bin/evmone-blockchaintest
+            --gtest_filter='*:-*eip4844*:*eip4788*'
```
chfast, line 366: Add a TODO comment why they are disabled.

test/blockchaintest/blockchaintest_loader.cpp
```diff
@@ -124,7 +115,17 @@ BlockchainTest load_blockchain_test_case(const std::string& name, const json::js
     bt.rev = to_rev_schedule(j.at("network").get<std::string>());
 
     for (const auto& el : j.at("blocks"))
-        bt.test_blocks.emplace_back(load_test_block(el, bt.rev.get_revision(0)));
+    {
+        if (const auto it = el.find("rlp_decoded"); it != el.end())
+        {
+            auto test_block = load_test_block(el.at("rlp_decoded"), bt.rev.get_revision(0));
```
chfast, line 121: Do I get it right that "rlp_decoded" is some special scheme for invalid blocks? Please add short comment.

chfast (approved):

test/blockchaintest/blockchaintest_loader.cpp
pdobacz: yup, turns out it's even documented https://ethereum.github.io/execution-spec-tests/main/consuming_tests/blockchain_test/#invalidfixtureblock

test/blockchaintest/blockchaintest_loader.cpp
```diff
@@ -124,7 +115,19 @@ BlockchainTest load_blockchain_test_case(const std::string& name, const json::js
...
     for (const auto& el : j.at("blocks"))
-        bt.test_blocks.emplace_back(load_test_block(el, bt.rev.get_revision(0)));
+    {
+        // `rlp_decoded` holds the `FixtureBlock` element with the relevant block data for
+        // invalid blocks within a test. It should be a sibling element to `expectException`.
+        if (const auto it = el.find("rlp_decoded"); it != el.end())
+        {
+            auto test_block = load_test_block(el.at("rlp_decoded"), bt.rev.get_revision(0));
+            assert(el.find("expectException") != el.end());
+            test_block.valid = false;
```
gumb0, line 125: maybe beter
```suggestion
            assert(test_block.valid == false);
```

test/blockchaintest/blockchaintest_loader.cpp
pdobacz: > So setting `test.block` to `false` is redundant in `load_test_block()` ?

Hm, good catch. Formally, it seems there can be no `rlp_decoded`. Then code here will read using `load_test_block(el, ...)`, and it will find the `exceptException` field there, and continue. It works correctly but not for the right reasons.

In practical terms now I see this is only the case for `eofwrap` fixtures, and per current release it actually makes tests fail if I omit this line :/.

Let me think a bit how to untie this knot easily. Worst case I have to go first and fix EEST's eofwrap, release there and bump here, and then that line in `load_test_block` should be redundant. We'll also not allow cases where `rlp_decoded` is not there, :crossed_fingers: there won't be any.

test/blockchaintest/blockchaintest_loader.cpp
pdobacz: not the same thing, since `rlp_decoded` won't contain `expectException`

gumb0: I meant that the call to `load_test_block()` on the previous line already always sets `test_block.valid = false` iff `expectException` is present, so these conditions are equivalent and you don't need to do second assignment, just assert that it was already set.

But maybe you plan to change this logic anyway.

pdobacz: > on the previous line already always sets test_block.valid = false iff expectException is present

yes, this is the catch! `expectException` is never present in `rlp_decoded`, its only present in the `blocks` array element of type `InvalidFixtureBlock`. And yes, I think I'll need to change this logic somehow, but the eofwrap stuff is getting in the way :/. 

So to illustrate, valid blocks are like
```
blocks: [{decoded fields like header, txs etc, but never expectException}[
```
invalid ones:
```
blocks: [{rlp:..., expectException: ..., rlp_decoded: {decoded fields like header, txs etc, but never expectException}}]
```

So as stated above, code works correctly, but for the wrong reasons

test/blockchaintest/blockchaintest_loader.cpp
pdobacz: okay, I think I cleaned this up. Code now reflects the structure of a blockchain test better. There is unfortunately a situation when there is no `rlp_decoded`, when the rlp representation of the block is invalid (duh), so we still skip those tests, but just a few (of 4844 tests at least).

The redundant check is removed. I've also submitted https://github.com/ethereum/execution-spec-tests/pull/1028 to EEST to fix eof wrap, until that is merged and I do a release there, CI will fail. Actually, unless I get a very quick feedback there, I'll point our CI to `ipsilon` "private" release to not stall this over the holidays.

pdobacz: ooooh, right, those particular encoding tests are under 4844 so they are excluded, and CI is fine, so I think no obstacles to merge here. It is the followup #1077 which will have to wait.

PR #1090, opened 2024-12-23
LRU cache utility

chfast:
Adds generic single-threaded implementation of
the least recently used (LRU) cache.

This is to be used to cache the code and code analysis.

### DIFF
--- a/lib/evmone/lru_cache.hpp
+++ b/lib/evmone/lru_cache.hpp
@@ -1,0 +1,162 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2024 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <cassert>
+#include <list>
+#include <optional>
+#include <unordered_map>
+
+namespace evmone
+{
+/// Least Recently Used (LRU) cache.
+///
+/// A map of Key to Value with a fixed capacity. When the cache is full, a newly inserted entry
+/// replaces (evicts) the least recently used entry.
+/// All operations have O(1) complexity.
+template <typename Key, typename Value>
+class LRUCache
+{
+    struct LRUEntry
+    {
+        /// Reference to the existing key in the map.
+        ///
+        /// This is needed to get the LRU element in the map when eviction is needed.
+        ///  Pointers to node-based map entries are always valid.
+        /// TODO: Optimal solution would be to use the map iterator. They are also always valid
+        ///   because the map capacity is reserved up front and rehashing never happens.
+        ///   However, the type definition would be recursive: Map(List(Map::iterator)), so we need
+        ///   to use some kind of type erasure. We prototyped such implementation, but decided not
+        ///   to include it in the first version. Similar solution is also described in
+        ///   https://stackoverflow.com/a/54808013/725174.
+        const Key& key;
+
+        /// The cached value.
+        Value value;
+    };
+
+    using LRUList = std::list<LRUEntry>;
+    using LRUIterator = typename LRUList::iterator;
+    using Map = std::unordered_map<Key, LRUIterator>;
+
+    /// The fixed capacity of the cache.
+    const size_t capacity_;
+
+    /// The list to order the cache entries by the usage. The front element is the least recently
+    /// used entry.
+    ///
+    /// In classic implementations the order in the list is reversed (the front element is the most
+    /// recently used entry). We decided to keep the order as is because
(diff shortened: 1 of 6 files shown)


rodiazet (approved):
I sadly can't complain about anything. Let's ask Copilot then . :)

lib/evmone/lru_cache.hpp
```diff
@@ -0,0 +1,157 @@
...
+        ///   https://stackoverflow.com/a/54808013/725174.
+        const Key& key;
+
+        /// The cached value.
+        Value value;
+    };
+
+    using LRUList = std::list<LRUEntry>;
+    using LRUIterator = typename LRUList::iterator;
+    using Map = std::unordered_map<Key, LRUIterator>;
```
gumb0, line 39: Would it be possible/better to store values in the map entries? Then there'd be just one link `LRUEntry => MapNode`.
Now it's double-direction: `LRUEntry::key => MapNode::key` and `MapNode::value -> LRUEntry`

lib/evmone/lru_cache.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    /// The map of Keys to Values via the LRU list indirection.
+    ///
+    /// The Value don't have to be in the LRU list but instead can be placed in the map directly
+    /// next to the LRU iterator. We decided to keep this classic layout because we didn't notice
+    /// any performance difference.
+    Map map_;
+
+    /// Marks an element as the most recently used by moving it to the back of the LRU list.
+    void move_to_back(LRUIterator it) noexcept { lru_list_.splice(lru_list_.end(), lru_list_, it); }
```
gumb0, line 63: Nit: the name kinda leaks internal representation (order of elements in the list). Maybe better something like `touch()`, `use()`, `access()` ?

gumb0: Actually I didn't notice it's a private method, then it's fine

lib/evmone/lru_cache.hpp
```diff
@@ -0,0 +1,157 @@
...
+                // Rollback the eviction by re-inserting the node with original key back.
+                swap(key, node2.key());
+                map_.insert(std::move(node2));
+
+                // Returned iterator points to the element matching the key
+                // which value must be updated.
+                lru_it = it->second;
+            }
+            lru_it->value = std::move(value);  // Replace/update the value.
+            move_to_back(lru_it);
```
gumb0, line 133: Juggling of keys is quite confusing, I'd maybe add more comments like
```suggestion
            auto node = map_.extract(lru_it->key); // node.key() is LRU key
            swap(node.key(), key); // node.key() is new key, key is LRU key
            if (auto [it, inserted, node2] = map_.insert(std::move(node)); !inserted)
            {
                // node2.key() is new key (already in cache)
                // Failed re-insertion means the element with the new key is already in the cache.
                // Rollback the eviction by re-inserting the node with original key back.
                swap(key, node2.key()); // key is new key (already in cache), node2.key() is LRU key
                map_.insert(std::move(node2));

                // Returned iterator points to the element matching the key
                // which value must be updated.
                lru_it = it->second;
            }
            lru_it->value = std::move(value);  // Replace/update the value.
            move_to_back(lru_it);
```

lib/evmone/lru_cache.hpp
chfast: I think I thought about it, e.g. naming it `bump_usage()`. In the end I decided not be because we also use `lru_list_` directly in other places. E.g. we should also replace `ru_list_.emplace` with something like `new_use()`?

The main motivation was to wrap `splice()` which takes me usually some time to figure out.

lib/evmone/lru_cache.hpp
chfast: Yeah, this value update for existing key feature is quite annoying to implement. And in many cases it is never used... I will try to improve the comments.

lib/evmone/lru_cache.hpp
chfast: Yes, this is possible. Then the LRU list has only the pointer/reference/iterator to the map. And the map is Key => (LRUIterator, Value).

I'm kind of on your side here, but because the performance difference was negligible I decided to keep the "classic" layout for now. There is also small nuance: you cannot have `std::list<const Key&>` so you need to wrap the reference or use a pointer.

This is to be revisited in case we are going to use intrusive list and/or map iterators.

chfast (commented):
I probably should mention O(1) get/put complexity.

PR #1092, opened 2024-12-27
Get rid of extra requests data copying when calculating requests hash

### LINKED ISSUE #1086: Optimize requests hash calculation
There's copying of requests data, that could be avoided:

https://github.com/ethereum/evmone/blob/2939b9f66aa118adf1d673820f8433c5f102154d/test/state/requests.cpp#L21

Two possible ways:

1. Represent Requests as just struct `Requests { bytes data; }` and put type as the first byte of data right away when collecting requests.
2.  init-update-final API for sha256.

gumb0:
Change Requests representation to byte array of type + requests data.

Closes #1086

### DIFF
--- a/test/state/requests.hpp
+++ b/test/state/requests.hpp
@@ -34,14 +34,24 @@ struct Requests
         consolidation = 2,
     };
 
-    /// Requests type.
-    Type type = Type::deposit;
-    /// Requests data - an opaque byte array, contains zero or more encoded request objects.
-    evmc::bytes data;
+    /// Raw encoded data of requests object: first byte is type, the rest is request objects.
+    evmc::bytes raw_data;
 
-    explicit Requests(Type _type, evmc::bytes _data = {}) noexcept
-      : type(_type), data(std::move(_data))
-    {}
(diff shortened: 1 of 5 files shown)


test/state/requests.hpp
```diff
@@ -34,14 +34,21 @@ struct Requests
         consolidation = 2,
     };
 
+    /// Raw encoded data of requests object: first byte is type, the rest is request objects.
+    evmc::bytes raw_data;
+
+    explicit Requests(Type _type, const evmc::bytes& data = {})
```
chfast, line 40: Usage of `bytes` type seems inefficient. In the `system_call()` we are forced to create a temporary object of `bytes` just to be quicly replaced with "type + data". It may be better to use `bytes_view` here.

chfast (changes requested):

test/state/requests.hpp
gumb0: true, I forgot that we can get rid of `bytes` on the `system_call()` side

chfast (approved):

PR #1094, opened 2025-01-07
Exclude empty requests data in commitment (EIP-7685)

chfast:
Apply the [EIP-7685 update] included in [pectra-devnet-5].

[EIP-7685 update]: https://github.com/ethereum/EIPs/pull/8989
[pectra-devnet-5]: https://notes.ethereum.org/@ethpandaops/pectra-devnet-5

This requires a fix in tests for calldata repricing logic according to https://github.com/ethereum/EIPs/pull/9227

### DIFF
--- a/test/t8n/t8n.cpp
+++ b/test/t8n/t8n.cpp
@@ -255,8 +255,12 @@ int main(int argc, const char* argv[])
         {
             // EIP-7685: General purpose execution layer requests
             j_result["requests"] = json::json::array();
-            for (size_t i = 0; i < requests.size(); ++i)
-                j_result["requests"][i] = hex0x(requests[i].data());
+            for (const auto& r : requests)
+            {
+                if (!r.data().empty())
+                    // Only report non-empty requests. Include the leading type byte.
+                    j_result["requests"].emplace_back(hex0x(r.raw_data));
+            }
 
             auto requests_hash = calculate_requests_hash(requests);
 
--- a/circle.yml
+++ b/circle.yml
@@ -378,11 +378,13 @@ jobs:
       - run:
           name: "Execution spec tests (develop, blockchain_tests)"
           # Tests for in-development EVM revision currently passing.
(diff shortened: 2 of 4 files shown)


test/t8n/t8n.cpp
```diff
@@ -255,7 +255,11 @@ int main(int argc, const char* argv[])
             // EIP-7685: General purpose execution layer requests
             j_result["requests"] = json::json::array();
             for (size_t i = 0; i < requests.size(); ++i)
-                j_result["requests"][i] = hex0x(requests[i].data());
+            {
+                if (!requests[i].data().empty())
+                    // Skip empty requests, include the type byte
```
chfast, line 260: Where is the type byte added?

pdobacz: `raw_data` includes it (next line). Do we want to concatenate explicitly here?

chfast: Ok, so these are two comments?
1. Skip empty requests
2. Include non-empty requests with the  leading type byte.

pdobacz: yes... maybe this is over-commented, but I just mirrored the other similar `if` in the hashing routine

chfast: Just split the comment: one part next to the `if`, other part before the `emplace_back`.

chfast: Or "Only report non-empty requests. Include the leading type byte."

pdobacz:
Just for reference: This PR is required for https://github.com/ethereum/execution-spec-tests/pull/1028 to be able to `eofwrap` and release EOF EEST tests. It is likely also required to be able to fill Prague EEST tests with `evmone`

pdobacz (approved):
lgtm, but let's have more eyes on this, I authored parts of this

gumb0 (approved):

PR #1095, opened 2025-01-07
Implement EIP-7623 "Increase calldata cost"

chfast:
This implements [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623).
In effect, during transaction validation we compute minimal transaction
gas cost. After transaction execution the amount of gas used is
increased to this value if lower.

~~Requires #1107 #1108~~. Merged

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -39,12 +39,20 @@ int64_t compute_access_list_cost(const AccessList& access_list) noexcept
     return cost;
 }
 
-int64_t compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noexcept
+struct TransactionCost
+{
+    int64_t intrinsic = 0;
+    int64_t min = 0;
+};
+
+/// Compute the transaction intrinsic gas 𝑔₀ (Yellow Paper, 6.2) and minimal gas (EIP-7623).
+TransactionCost compute_tx_intrinsic_cost(evmc_revision rev, const Transaction& tx) noexcept
 {
     static constexpr auto TX_BASE_COST = 21000;
     static constexpr auto TX_CREATE_COST = 32000;
     static constexpr auto DATA_TOKEN_COST = 4;
     static constexpr auto INITCODE_WORD_COST = 2;
+    static constexpr auto TOTAL_COST_FLOOR_PER_TOKEN = 10;
 
     const auto is_create = !tx.to.has_value();  // Covers also EOF creation transactions.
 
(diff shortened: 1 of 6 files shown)


test/state/transaction.hpp
```diff
@@ -72,6 +72,8 @@ struct TransactionProperties
 {
     /// The amount of gas provided to the EVM for the transaction execution.
     int64_t execution_gas_limit = 0;
+
```
rodiazet, line 75: Add doc string

test/unittests/state_tx_test.cpp
rodiazet, line 1: I would add one more test which tests two different case:
1. Case in which tx gas cost is the intrinsic. So the ratio of calldata gas and execution gas is below threshold. 
2. Case where we are above threshold and we have tx cost calculated calculated based on `TOTAL_COST_FLOOR_PER_TOKEN`
It should be possible just by extending call data size until we get into second case.

test/state/state.cpp
```diff
@@ -497,6 +510,9 @@ TransactionReceipt transition(const StateView& state_view, const BlockInfo& bloc
     gas_used -= refund;
     assert(gas_used > 0);
 
+    // EIP-7623: The gas used by the transaction must be at least the min_gas_cost.
+    gas_used = std::max(gas_used, tx_props.min_gas_cost);
```
gumb0, line 514: So in the end gas used may be higher than transaction gas limit? wild

gumb0: Ah no, I guess not, because it's checked in transaction validity

test/unittests/state_tx_test.cpp
chfast: I added 3 execution test cases. Please check if these are what you requested.

test/state/transaction.hpp
chfast: Done.

test/unittests/state_tx_test.cpp
rodiazet: Yes. Thanks.

rodiazet (approved):

test/state/state.cpp
gumb0: Hm in EIP max operation left side is 
```STANDARD_TOKEN_COST * tokens_in_calldata
+ execution_gas_used
+ isContractCreation * (32000 + INITCODE_WORD_COST * words(calldata))
```
but here it is, I think
```
21000
+ STANDARD_TOKEN_COST * tokens_in_calldata
+ execution_gas_used
+ isContractCreation * (32000 + INITCODE_WORD_COST * words(calldata))
```
because `21000` was included in `intrinsic_cost`?

chfast: By the Yellow Paper and the names we used so far the intrinsic gas includes the base cost of 21000. The EIP extracts the base cost outside of the max function: `21000 + max(intrinsic_without_base, floor_gas)`.
Because also the EIP don't name any of these values, I decided to include the base cost in the `min_gas_cost`. It is easier to use.

gumb0: Oh I see

pdobacz (approved):

PR #1101, opened 2025-01-10
tidy: Remove redundant inline specifiers

chfast:
clang-tidy guarantees that the readability-redundant-inline-specifier check in the default non-strict mode will only report inline specifiers not affecting the generated machine code.

Enable this check and fix reported issues.


rodiazet (approved):
Looks good to me.

PR #1102, opened 2025-01-10
tidy: Use ranges version of C++ algorithms

chfast:
Apply suggestions by the clang-tidy's modernize-use-ranges check.

### DIFF
--- a/test/bench/bench.cpp
+++ b/test/bench/bench.cpp
@@ -58,7 +58,7 @@ std::vector<BenchmarkCase::Input> load_inputs(const StateTransitionTest& state_t
 BenchmarkCase load_benchmark(const fs::path& path, const std::string& name_prefix)
 {
     std::ifstream f{path};
-    auto state_test = std::move(evmone::test::load_state_tests(f).at(0));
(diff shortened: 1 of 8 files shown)


test/bench/bench.cpp
```diff
@@ -58,7 +58,7 @@ std::vector<BenchmarkCase::Input> load_inputs(const StateTransitionTest& state_t
 BenchmarkCase load_benchmark(const fs::path& path, const std::string& name_prefix)
 {
     std::ifstream f{path};
-    auto state_test = std::move(evmone::test::load_state_tests(f).at(0));
+    auto state_test = std::move(load_state_tests(f).at(0));
```
rodiazet, line 61: Unrelated change. I would move it to separated commit.

rodiazet (approved):
One nitpick only. I haven't checked if there are some more places in the code where this improvement can be applied.

test/bench/bench.cpp
chfast: This is almost like a typo, we are already in the `evmone::test` namespace.

PR #1105, opened 2025-01-13
ci: Add pectra-devnet-5 EEST

chfast:
Run EEST tests pectra-devnet-5@v1.2.0 alongside the devnet-4 tests. This allows migrating to the devnet-5 specs EIP-by-EIP. In this PR we just enable example devnet-5 tests which happen to pass in the devnet-4 implementation.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -370,14 +370,14 @@ jobs:
           release: pectra-devnet-4@v1.0.1
           fixtures_suffix: pectra-devnet-4
       - run:
-          name: "Execution spec tests (develop, state_tests)"
+          name: "Execution spec tests (develop, state_tests) - pectra-devnet-4"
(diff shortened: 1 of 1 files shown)


gumb0 (approved):

circle.yml
```diff
@@ -387,6 +387,25 @@ jobs:
             prague/eip7002_el_triggerable_withdrawals
             prague/eip7251_consolidations
             prague/eip7685_general_purpose_el_requests
+      - download_execution_spec_tests:
```
gumb0, line 390: I think this unpacks them into the same `fixtures` subfolder as previous `pectra-devnet-4` download?

chfast: Yes. If an issue clear the directory before unpacking in the download step

PR #1111, opened 2025-01-20
tests: Add option to filter state tests by name

rodiazet:
This PR improves state test functionality by allowing filtering by test case name defined inside json test file.

### DIFF
--- a/test/statetest/statetest.cpp
+++ b/test/statetest/statetest.cpp
@@ -16,12 +16,14 @@ namespace
 class StateTest : public testing::Test
 {
     fs::path m_json_test_file;
+    std::optional<std::string> m_filter;
     evmc::VM& m_vm;
     bool m_trace = false;
 
 public:
-    explicit StateTest(fs::path json_test_file, evmc::VM& vm, bool trace) noexcept
-      : m_json_test_file{std::move(json_test_file)}, m_vm{vm}, m_trace{trace}
+    explicit StateTest(fs::path json_test_file, const std::optional<std::string>& filter,
+        evmc::VM& vm, bool trace) noexcept
+      : m_json_test_file{std::move(json_test_file)}, m_filter{filter}, m_vm{vm}, m_trace{trace}
     {}
 
     void TestBody() final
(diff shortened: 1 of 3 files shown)


test/statetest/statetest.cpp
```diff
@@ -16,31 +16,42 @@ namespace
...
     {}
 
     void TestBody() final
     {
         std::ifstream f{m_json_test_file};
         const auto tests = evmone::test::load_state_tests(f);
         for (const auto& test : tests)
-            evmone::test::run_state_test(test, m_vm, m_trace);
+        {
+            if (!m_test_name.has_value() || test.name.find(*m_test_name) != std::string::npos)
```
chfast, line 38: Restructure this to use `continue`.

test/statetest/statetest.cpp
```diff
@@ -16,31 +16,42 @@ namespace
 class StateTest : public testing::Test
 {
     fs::path m_json_test_file;
+    std::optional<std::string> m_test_name;
```
chfast, line 19: This is rather a filter than a name. Maybe name it `m_name_filter` or just `m_filter`.

test/statetest/statetest.cpp
```diff
@@ -88,6 +99,10 @@ int main(int argc, char* argv[])
             ->required()
             ->check(CLI::ExistingPath);
 
+        std::optional<std::string> test_name;
+        app.add_option("--name", test_name,
```
chfast, line 103: The "name" is too generic. We should also suggest this is a test filter. Maybe actually go with pytest `-k` which stands for "keyword".

gumb0: In geth it is `--run <regexp>`, also quite generic

chfast: `--run` is fine to me. it comes from golang standard testing library.

chfast (commented):
This likely needs an integration test.

rodiazet:
Integration test added.

chfast (approved):

PR #1118, opened 2025-01-27
 Support EIP-7691: Blob throughput increase

pdobacz:
Implements [EIP-7691](https://eips.ethereum.org/EIPS/eip-7691). It seems only 7702 is missing (once spec stabilizes) from passing EEST devnet-5/6 tests

The parametrization of blob gas schedule with `rev` led to quite a bit of fallout...

### DIFF
--- a/test/blockchaintest/blockchaintest_loader.cpp
+++ b/test/blockchaintest/blockchaintest_loader.cpp
@@ -54,7 +54,7 @@ BlockHeader from_json<BlockHeader>(const json::json& j)
     };
 }
 
-static TestBlock load_test_block(const json::json& j, evmc_revision rev)
+static TestBlock load_test_block(const json::json& j, const RevisionSchedule& rev_schedule)
 {
     using namespace state;
     TestBlock tb;
@@ -64,6 +64,9 @@ static TestBlock load_test_block(const json::json& j, evmc_revision rev)
         tb.expected_block_header = from_json<BlockHeader>(*it);
         tb.block_info.number = tb.expected_block_header.block_number;
         tb.block_info.timestamp = tb.expected_block_header.timestamp;
+
+        const auto rev = rev_schedule.get_revision(tb.block_info.timestamp);
+
         tb.block_info.gas_limit = tb.expected_block_header.gas_limit;
         tb.block_info.coinbase = tb.expected_block_header.coinbase;
         tb.block_info.difficulty = tb.expected_block_header.difficulty;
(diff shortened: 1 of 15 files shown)


test/unittests/statetest_loader_test.cpp
```diff
@@ -117,7 +117,9 @@ TEST(statetest_loader, load_minimal_test)
                 "value": null,
                 "nonce" : "0"
             },
-            "post": {},
+            "post": {
+                "Cancun": []
```
chfast, line 121: Do EEST have multiple entries here?

chfast (commented):
Do you want to split out the BlockInfo change?

test/unittests/statetest_loader_test.cpp
pdobacz: Hm, no, it seems `"Cancun"` and `"Prague"` keys would get split into two tests. Do you think this warrants removing the [tests: Support varying BlockInfo in statetests](https://github.com/ethereum/evmone/pull/1118/commits/1ddc15395b0776132eaf0269f8ce9a01b78651e0)? And just using the first entry and asserting there is only one?

chfast: No, I think the change is fine. We rather should push for the test format update long term.

chfast: There is also one change I have in mind: we should split `BlockInfo` into multiple structs: data from tests, computed values, data needed for EVM execution.

pdobacz: > No, I think the change is fine. We rather should push for the test format update long term.

Meanwhile, I've tried it. It is only marginally smaller, not worth it

> There is also one change I have in mind: we should split BlockInfo into multiple structs: data from tests, computed values, data needed for EVM execution.

I presume not in this PR?

> Do you want to split out the BlockInfo change?

I was not planning to squash it, so it remains as a separate commit, ok?

test/blockchaintest/blockchaintest_loader.cpp
```diff
@@ -64,6 +64,9 @@ static TestBlock load_test_block(const json::json& j, evmc_revision rev)
         tb.expected_block_header = from_json<BlockHeader>(*it);
         tb.block_info.number = tb.expected_block_header.block_number;
         tb.block_info.timestamp = tb.expected_block_header.timestamp;
+
+        const evmc_revision rev = revision_schedule.get_revision(tb.block_info.timestamp);
```
chfast, line 68:
```suggestion
        const auto rev = revision_schedule.get_revision(tb.block_info.timestamp);
```

test/blockchaintest/blockchaintest_loader.cpp
```diff
@@ -54,7 +54,7 @@ BlockHeader from_json<BlockHeader>(const json::json& j)
     };
 }
 
-static TestBlock load_test_block(const json::json& j, evmc_revision rev)
+static TestBlock load_test_block(const json::json& j, const RevisionSchedule& revision_schedule)
```
chfast, line 57:
```suggestion
static TestBlock load_test_block(const json::json& j, const RevisionSchedule& rev_schedule)
```

chfast (approved):

PR #1120, opened 2025-01-30
evmmax: Implementation of vectorized version od EVMMAX opcodes (EIP-6690)

rodiazet:
This PR implements PoC version of EVMMAX according to EIP-6690 (called vectorized, SIMD or sequential version).
Gas model is not updated to how the EIP defines it.


pdobacz:
Nice! Would it be unreasonable to, instead of 2 branches, just have both options (vectorized and not) available on 1 branch, with opcodes assigned somewhere arbitrarily and with heavy copy-paste code? Or even same opcodes but a feature flag on the `libevmone.so`? I'm guessing there's going to be a lot of experimentation and measuring involved here, maybe it's better to plan for it from day 1?

rodiazet:
Good point. I can think of some c++ macro which enables requested version or maybe introduce a runtime flag.

PR #1122, opened 2025-02-04
eof: Detect unused inputs

chfast:
Add one more EOF validation rule: all code section inputs must be used. This is validated by tracking if any instruction in the code section reaches the stack bottom (height 0).

Spec issue: https://github.com/ipsilon/eof/issues/176.

### DIFF
--- a/test/unittests/eof_validation_stack_test.cpp
+++ b/test/unittests/eof_validation_stack_test.cpp
@@ -1241,11 +1241,11 @@ TEST_F(eof_validation, jumpf_to_nonreturning)
 
     // Exactly required inputs on stack at JUMPF
     add_test_case(eof_bytecode(3 * OP_PUSH0 + jumpf(1), 3).code(OP_STOP, 3, 0x80, 3),
-        EOFValidationError::success);
+        EOFValidationError::unused_input);
 
     // Extra items on stack at JUMPF
     add_test_case(eof_bytecode(4 * OP_PUSH0 + jumpf(1), 4).code(OP_STOP, 3, 0x80, 3),
-        EOFValidationError::success);
+        EOFValidationError::unused_input);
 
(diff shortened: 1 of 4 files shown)


test/unittests/eof_validation_stack_test.cpp
```diff
@@ -1241,11 +1241,11 @@ TEST_F(eof_validation, jumpf_to_nonreturning)
 
     // Exactly required inputs on stack at JUMPF
     add_test_case(eof_bytecode(3 * OP_PUSH0 + jumpf(1), 3).code(OP_STOP, 3, 0x80, 3),
-        EOFValidationError::success);
+        EOFValidationError::unused_input);
```
pdobacz, line 1244: not necessarily urgent, but leaving a comment to not miss - we may maybe want to move such `unused_input` cases to their separate batch of tests, and rewrite these variants to actually have them use the inputs.

Or, maybe actually writing these tests now could be useful to get more insight into what the change entails.

pdobacz:
another observation - did you maybe check if this breaks any RETF tests (i.e. in EEST?). If not, we're missing tests

chfast:
> another observation - ~did you maybe check if this breaks any RETF tests (i.e. in EEST?). If not, we're missing tests~
> 
> Ah, I see in the `eof` issue - EEST tests only broke with JUMPF, so we're missing tests.

No. It only breaks similar `JUMPF` tests.

test/unittests/eof_validation_stack_test.cpp
chfast: That's the plan. I only wanted to list all affected tests.

chfast:
We will not do this.

pdobacz:
@chfast did we settle on whether or not we need new RETF EEST tests. IIUC none broke when you implemented this?

PR #1134, opened 2025-02-17
Add prestate validation checks to state test loader

gumb0:
Idea originated in https://github.com/ethereum/evmone/pull/961#pullrequestreview-2606700539

### DIFF
--- a/test/statetest/statetest_loader.cpp
+++ b/test/statetest/statetest_loader.cpp
@@ -498,8 +498,9 @@ void validate_state(const TestState& state, evmc_revision rev)
 {
     for (const auto& [addr, acc] : state)
     {
-        // TODO: Check for empty accounts after Paris.
-        //       https://github.com/ethereum/tests/issues/1331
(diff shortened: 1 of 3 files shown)


test/statetest/statetest_loader.cpp
```diff
@@ -4,7 +4,10 @@
 
 #include "../utils/stdx/utility.hpp"
 #include "../utils/utils.hpp"
+#include "evmone/delegation.hpp"
```
chfast, line 7:
```suggestion
#include <evmone/delegation.hpp>
```

test/statetest/statetest_loader.cpp
```diff
@@ -4,7 +4,10 @@
...
 #include "statetest.hpp"
+#include "test/state/precompiles.hpp"
```
chfast, line 9:
```suggestion
#include <test/state/precompiles.hpp>
```

chfast (approved):
LGTM.

I think we should submit a PR to EEST with the test removal. This may start a longer discussion there, but this is fine.

Then we can wait until the discussion is resolved and merge this, or merge it before.

chfast (approved):

PR #1135, opened 2025-02-18
precompiles: Implement modexp (up to 1024 bytes)

rodiazet:
Using Montgomery modular multiplication from evmmax utils,
implement modular exponentiation (modexp) with unlimited exponent length
and base/mod up to 1024 bytes (EIP-7823 limits).


rodiazet:
NO EOF = NO EVMMAX

test/state/precompiles.cpp
```diff
@@ -363,11 +366,24 @@ ExecutionResult expmod_execute(
...
     const auto mod = mod_requires_padding ? bytes_view{output, mod_len} : mod_explicit;
 
-#ifdef EVMONE_PRECOMPILES_GMP
+    uint8_t evmone_output[1024];  // Use a separate buffer not to overwrite the truncated mod.
+    const auto in_range = crypto::modexp(base, exp, mod, evmone_output);
+    if (in_range)
+    {
+        uint8_t gmp_output[1024];
+        expmod_gmp(base, exp, mod, gmp_output);
+        if (std::memcmp(evmone_output, gmp_output, mod_len) != 0)
```
chfast, line 375: This is a temporary code for fuzzing. Maybe just move it under `#ifndef NDEBUG` (fuzzer compiles with asserts enabled).

test/state/precompiles.cpp
```diff
@@ -363,11 +366,24 @@ ExecutionResult expmod_execute(
...
+            std::cerr << "modexp output mismatch"
+                      << "\n  evmone: " << evmc::hex({output, mod_len})
+                      << "\n  gmp:    " << evmc::hex({gmp_output, mod_len}) << '\n';
+            assert(false);
+        }
+        std::copy_n(evmone_output, mod_len, output);
+        return {EVMC_SUCCESS, mod_len};
+    }
+
     expmod_gmp(base, exp, mod, output);
```
chfast, line 386: The new implementation is slower than GMP so it should replace the "stub", but GMP has to have the priority.

test/unittests/precompiles_expmod_test.cpp
```diff
@@ -22,14 +22,78 @@ TEST(expmod, test_vectors)
     const std::vector<TestCase> test_cases{
         {"", "", "", ""},
         {"", "", "00", "00"},
+        {"", "", "01", "00"},
```
chfast, line 25: I added some test case for debugging. You can set them to evmone as a separate PR.

chfast (approved):

Issue #1130, opened 2025-02-13
evmone t8n changes the state with --state.reward -1  on Prague


winsvega:
I am asking evmone to calculate genesis state hash on Prague. 
But it is changing the state even though reward -1 flag provided:

http://retesteth.ethdevops.io/temp/evmone/eels_Prague_block0_0x5c6f65/outAlloc.json

```
Error: Check State: Remote account '0x0000f90827f1c53a10cb7a02335b175320002935'  has more storage records than expected!
 [0x01ff] = 0x798272c22de7de1bbb41d9d76b5240e67bb83e9ece1afeb940834536b3646693(54960336831144048435103419112880773968606735434740154474392682579052779038355)
 ```

chfast:
Do you mean the hack is: don't execute system contracts with reward -1?

winsvega:
yes

PR #1136, opened 2025-02-18
t8n: Skip system contracts if --state.reward=-1

chfast:
Fixes https://github.com/ethereum/evmone/issues/1130.


winsvega (approved):
Verified, now it doesn't change the state with reward set to -1

chfast:
This should have an integration test, but I don't have time for it. Maybe later when we will improve t8n testability.

PR #1137, opened 2025-02-19
eof: Remove duplicated type section size check

chfast:
The check that type section size matches the number of code sections is done twice in EOF validation:
- in `validate_section_headers()`
- in `validate_header()` which always calls `validate_section_headers()` so this check never fails.

I decided to leave the check in `validate_header()` because it better fits there.


chfast:
> I was trying hard to remember the historical reasons behind this, but couldn't. lgtm

Added here: https://github.com/ethereum/evmone/pull/1034/files#diff-6c68f99275b828a78c2791fee776edf5c210a321c916e8afc58689b1d4d5762dR771

PR #1139, opened 2025-02-21
EIP-7873: EOF - TXCREATE and InitcodeTransaction type

gumb0:


pdobacz (approved):

lib/evmone/eof.hpp
```diff
@@ -192,7 +192,7 @@ enum class EOFValidationError
 
 enum class ContainerKind : uint8_t
 {
-    /// Container that uses RETURNCODE. Can be used by EOFCREATE/Creation transaction.
+    /// Container that uses RETURNCODE. Can be used by EOFCREATE/TXCREATE/Creation transaction.
```
chfast, line 195:
```suggestion
    /// Container that uses RETURNCODE. Can be used by EOFCREATE/TXCREATE.
```
The create transaction is on the way out?

lib/evmone/execution_state.hpp
```diff
@@ -172,6 +172,7 @@ class ExecutionState
 
 private:
     evmc_tx_context m_tx = {};
+    std::optional<std::unordered_map<evmc::bytes32, bytes_view>> m_initcodes;
```
chfast, line 175: Will it fly if we omit `std::optional` (empty map meaning no initcodes)?

lib/evmone/instructions_calls.cpp
```diff
@@ -398,11 +404,24 @@ Result eofcreate(
     if (!check_memory(gas_left, state.memory, input_offset_u256, input_size_u256))
```
chfast, line 404: This is weird that memory check happens _after_ we push 0 to the stack. Please check this.

test/state/state.cpp
```diff
@@ -450,11 +463,27 @@ std::variant<TransactionProperties, std::error_code> validate_transaction(
...
         break;
 
+    case Transaction::Type::initcodes:
+        if (rev < EVMC_OSAKA)
+            return make_error_code(TX_TYPE_NOT_SUPPORTED);
+        if (tx.initcodes.size() > MAX_INITCODE_COUNT)
+            return make_error_code(INIT_CODE_COUNT_LIMIT_EXCEEDED);
+        if (tx.initcodes.empty())
+            return make_error_code(INIT_CODE_COUNT_ZERO);
+        if (std::any_of(tx.initcodes.begin(), tx.initcodes.end(),
```
chfast, line 473: Try `std::ranges`

lib/evmone/execution_state.hpp
gumb0: I think this is optimization to avoid calling `get_tx_context().initcodes` on every `get_tx_initcode_by_hash()`.

So it will work without optional, but for no initcodes will check `tx_context.initcodes_count` on every TXCREATE (will not query the host though, thanks to cached `tx_context`).

lib/evmone/instructions_calls.cpp
gumb0: It's the same as what CREATE* is doing (stack is not observable after OOG)

lib/evmone/eof.hpp
gumb0: It is, not in this PR though.

gumb0: Removing it in https://github.com/ethereum/evmone/pull/1190

pdobacz (approved):

chfast (approved):

PR #1140, opened 2025-02-24
Change EOFCREATE new address hashing scheme

gumb0:
Implement https://github.com/ethereum/EIPs/pull/9298


pdobacz (commented):
this should also remove hashing cost from `lib/evmone/instructions_calls.cpp:408`

pdobacz (approved):

pdobacz (approved):

PR #1141, opened 2025-02-24
EIP-7702: Remove the check for empty in-state accounts

chfast:
In the original EIP-7702 implementation we were handling empty accounts in the state. Such accounts must not exist in the state, but some tests were disobeying this property.


gumb0:
Looks fine to me

gumb0 (approved):

PR #1145, opened 2025-02-26
eof: Clean up impossible EOF validation conditions

chfast:
- Remove the "impossible" validation error.
- Replace assert(false) with unreachable().
- Downgrade some checks to asserts.


chfast:
Comment about unreachable:

- the `std::unreachable()` traps in Debug builds, but `intx::unreachable()` don't. Upgrading to C++23 solves this.
- we can still detect reaching unreachable with `-fsanitize=unreachable`. Fuzzers do this.

gumb0 (approved):

PR #1150, opened 2025-03-03
precompiles: Fix BLS MSMs for inputs with points-at-infinity

rodiazet:
Fixes BLS MSM  implementations for inputs with points-at-infinity.

Grazie molto for @holiman for pointing this out. 🥇


chfast (changes requested):
This requires a regression test for each precompile.

chfast (approved):

PR #1152, opened 2025-03-05
Check for topic when parsing deposit contract logs

gumb0:
Turns out on Sepolia deposit contract is different and can emit other logs (ERC20's `Transfer`)

### DIFF
--- a/test/state/requests.cpp
+++ b/test/state/requests.cpp
@@ -36,7 +36,9 @@ Requests collect_deposit_requests(std::span<const TransactionReceipt> receipts)
     {
         for (const auto& log : receipt.logs)
         {
-            if (log.addr == DEPOSIT_CONTRACT_ADDRESS)
+            assert(log.addr != DEPOSIT_CONTRACT_ADDRESS || !log.topics.empty());
+            if (log.addr == DEPOSIT_CONTRACT_ADDRESS &&
+                log.topics[0] == DEPOSIT_EVENT_SIGNATURE_HASH)
             {
                 // Deposit log definition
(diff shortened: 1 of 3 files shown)


shemnon:
LGTM.  This is how all the other clients approach it.

chfast (commented):
Can we have a regression test for this?

test/state/requests.cpp
```diff
@@ -11,6 +11,9 @@ namespace
 {
 /// The address of the deposit contract.
 constexpr auto DEPOSIT_CONTRACT_ADDRESS = 0x00000000219ab540356cBB839Cbe05303d7705Fa_address;
+/// The topic of deposit log of the deposit contract.
+constexpr auto DEPOSIT_TOPIC =
```
chfast, line 15: This should go to the header file now (so we can use it in unit tests).

test/state/requests.cpp
```diff
@@ -42,7 +45,8 @@ Requests collect_deposit_requests(std::span<const TransactionReceipt> receipts)
     {
         for (const auto& log : receipt.logs)
         {
-            if (log.addr == DEPOSIT_CONTRACT_ADDRESS)
+            assert(log.addr != DEPOSIT_CONTRACT_ADDRESS || !log.topics.empty());
+            if (log.addr == DEPOSIT_CONTRACT_ADDRESS && log.topics[0] == DEPOSIT_TOPIC)
```
chfast, line 49: We usually don't do this, but I'm thinking about also checking if the `topics[0]` exists and if the length of the data is enough. But we can wait for the https://github.com/ethereum/EIPs/pull/9460.

chfast (approved):

PR #1154, opened 2025-03-06
Improve BLS precompiles implementation

chfast:
Most importantly, this removes double induction variables in loops in the pairing chack and MSMs. This allows us to use early breaks in the loop safely.

### DIFF
--- a/lib/evmone_precompiles/bls.cpp
+++ b/lib/evmone_precompiles/bls.cpp
@@ -196,7 +196,7 @@ void store(uint8_t _rx[128], const blst_fp2& _x) noexcept
 {
     constexpr auto SINGLE_ENTRY_SIZE = (64 * 2 + 32);
     assert(size % SINGLE_ENTRY_SIZE == 0);
-    auto npoints = size / SINGLE_ENTRY_SIZE;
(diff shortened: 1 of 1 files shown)


lib/evmone_precompiles/bls.cpp
```diff
@@ -365,13 +359,11 @@ void store(uint8_t _rx[128], const blst_fp2& _x) noexcept
     static constexpr auto PAIR_SIZE = P1_SIZE + P2_SIZE;
     assert(size % PAIR_SIZE == 0);
 
-    const auto npairs = size / PAIR_SIZE;
-    auto ptr = _pairs;
-    blst_fp12 acc = *blst_fp12_one();
-
     const auto Qlines = std::make_unique_for_overwrite<blst_fp6[]>(68);
```
rodiazet, line 362: I would remove `Qlines` and run regular pairing without lines precomputation.

rodiazet (approved):
OK with one comment regarding lines pre-computation.

lib/evmone_precompiles/bls.cpp
chfast: Done in https://github.com/ethereum/evmone/pull/1155.

PR #1159, opened 2025-03-10
Add basic unit tests for collecting deposit requests

chfast:


test/state/requests.hpp
```diff
@@ -8,10 +8,14 @@
...
 #include <evmc/evmc.hpp>
 #include <span>
-#include <vector>
 
 namespace evmone::state
 {
+/// The address of the deposit contract.
+///
+/// TODO: This address differs in different chains, so it should be configurable.
+constexpr auto DEPOSIT_CONTRACT_ADDRESS = 0x00000000219ab540356cBB839Cbe05303d7705Fa_address;
```
pdobacz, line 17: Would it be possible to add some kind of an assertion whenever that constant is accessed, e.g. check the chain_id? This TODO is likely to be missed.

chfast: It depends how the tests will evolve. It seems this should be a parameter of the JSON tests. In any case, this requires some research. As long as we pass all tests, this is not priority.

pdobacz: I only meant in the unlikely event someone integrates evmone into a full client and forgets this.

pdobacz: > integrates evmone into a full client and forgets
should have been "integrates before we fix the TODO"

chfast: Yes, this is likely to happen. We had similar issue in Silkworm where we had chain id hardcoded to 1. This was only noticed by syncing a testnet.

test/unittests/state_deposit_requests_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+    log_data.replace(6 * 32, 48, 48, 0x01);   // pubkey
+    log_data.replace(9 * 32, 32, 32, 0x02);   // withdrawal_credentials
+    log_data.replace(11 * 32, 8, 8, 0x03);    // amount
+    log_data.replace(13 * 32, 96, 96, 0x04);  // signature
+    log_data.replace(17 * 32, 8, 8, 0x05);    // index
+
+    const auto requests = collect_deposit_requests(receipts);
+    EXPECT_EQ(requests.type(), Requests::Type::deposit);
+    EXPECT_EQ(requests.data().size(), 192);
+    EXPECT_EQ(requests.data().substr(0, 48), bytes(48, 0x01));
```
gumb0, line 28: could be fewer offset calculations like
```suggestion
    EXPECT_EQ(requests.data(), bytes(48, 0x01) + bytes(32, 0x02) + ...);
```

gumb0 (approved):

test/unittests/state_deposit_requests_test.cpp
chfast: Make sense. I will change it.

PR #1163, opened 2025-03-11
precompiles: Handle trivial inputs to expmod

chfast:
Before looking up an answer for the expmod precompile in stub,
check if the input has at least the bytes encoding the value lengths.
If not or if the modulus length is 0 then the answer is trivially
the sequence of zero bytes of the length of the modulus.


pdobacz (approved):

rodiazet (approved):
Looks good. As a comment I would also add less trivial check. In case when input size is less or equal than `96+base_len+exp_len` it such a case modulus is also `0`. So the result is `0` too. This can be added in a separated PR.

chfast:
> Looks good. As a comment I would also add less trivial check. In case when input size is less or equal than `96+base_len+exp_len` it such a case modulus is also `0`. So the result is `0` too. This can be added in a separated PR.

This is correct, but I'm not sure we want to go this path. I.e. we don't want do handle degenerated cases separately unless necessary.

PR #1165, opened 2025-03-13
precompiles: Use classic EC point add formula

chfast:
This makes the BN254 `ecadd` precompile 30% faster.

### DIFF
--- a/lib/evmone_precompiles/ecc.hpp
+++ b/lib/evmone_precompiles/ecc.hpp
@@ -91,6 +91,43 @@ inline Point<IntT> to_affine(const ModArith<IntT>& s, const ProjPoint<IntT>& p) 
     return {s.from_mont(s.mul(p.x, z_inv)), s.from_mont(s.mul(p.y, z_inv))};
 }
 
+/// Adds two elliptic curve points in affine coordinates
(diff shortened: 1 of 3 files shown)


lib/evmone_precompiles/ecc.hpp
```diff
@@ -91,6 +91,43 @@ inline Point<IntT> to_affine(const ModArith<IntT>& s, const ProjPoint<IntT>& p)
...
+    const auto slope = m.mul(dy, m.inv(dx));
+
+    const auto xr = m.sub(m.sub(m.mul(slope, slope), x1), x2);
+    const auto yr = m.sub(m.mul(m.sub(x1, xr), slope), y1);
+    return {m.from_mont(xr), m.from_mont(yr)};
+}
+
 template <typename IntT, int A = 0>
 ProjPoint<IntT> add(const evmmax::ModArith<IntT>& s, const ProjPoint<IntT>& p,
     const ProjPoint<IntT>& q, const IntT& b3) noexcept
```
rodiazet, line 133: Why not drop it if we have general formula?

rodiazet (approved):
Looks good. I left one comment?

PR #1171, opened 2025-03-17
EOF: Make container sections size 4 bytes

gumb0:


lib/evmone/eof.hpp
```diff
@@ -80,11 +91,11 @@ struct EOF1Header
     /// @data_size.
     uint16_t data_size = 0;
     /// Offset of data container section start.
-    uint16_t data_offset = 0;
+    uint32_t data_offset = 0;
     /// Size of every container section.
-    std::vector<uint16_t> container_sizes;
+    std::vector<uint32_t> container_sizes;
```
gumb0, line 96: Strictly speaking we don't need the changes here yet, given current code size limit.

But during header validation we do need to parse full 4-byte sizes and then use them to compare declared body size against actual body size, so we have there a vector of `uint32_t` container sizes already, and it's simpler and more efficient to just move it into this vector here.

test/unittests/eof_example_test.cpp
```diff
@@ -168,42 +168,43 @@ TEST_F(state_transition, eof_examples_eofcreate)
...
-        //       |                    |            Header: data section 0 bytes long
-        //       |                    |                       |
-        //       Header: types section 4 bytes long           Types section: first code section
-        //                            |                       0 inputs, non-returning,
-        //                            |                       max stack height 2
+        //                                                                   |
+        //               Header: 1 code section 4 bytes long                 |
+        //               |                                                   |
+        //    version    |                                     Header terminator
+        //    |          |___________                          |                 |_________
```
pdobacz, line 200:
```suggestion
        //    |          |___________                          |             |_________
```

pdobacz (approved):

lib/evmone/eof.cpp
```diff
@@ -40,25 +35,33 @@ constexpr auto REL_OFFSET_SIZE = sizeof(int16_t);
...
     auto header_size = std::size(EOF_MAGIC) + 1 +  // 1 version byte
                        non_code_section_count * non_code_section_header_size +
-                       sizeof(CODE_SECTION) + 2 + code_section_count * section_size_size +
+                       sizeof(CODE_SECTION) + 2 + code_section_count * code_section_size_size +
                        sizeof(TERMINATOR);
 
     if (container_section_count != 0)
     {
-        header_size += sizeof(CONTAINER_SECTION) + 2 + container_section_count * section_size_size;
+        constexpr auto container_section_size_size = 4;
```
chfast, line 62:
```suggestion
        constexpr auto CONTAINER_SIZE_SIZE = sizeof(uint32_t);
```

lib/evmone/eof.cpp
```diff
@@ -204,12 +219,12 @@ std::variant<EOFSectionHeaders, EOFValidationError> validate_section_headers(byt
         return EOFValidationError::section_headers_not_terminated;
 
     const auto section_bodies_without_data =
-        section_headers[TYPE_SECTION].front() +
+        static_cast<uint64_t>(section_headers.type_size) +
```
chfast, line 222:
```suggestion
        uint64_t{section_headers.type_size} +
```

I'm not sure about uint64_t. Can't we keep it as uint32_t?

lib/evmone/eof.cpp
```diff
@@ -729,7 +744,7 @@ size_t EOF1Header::data_size_position() const noexcept
            3 +                          // type section kind + size
            3 + 2 * num_code_sections +  // code sections kind + count + sizes
            // container sections kind + count + sizes
-           (num_container_sections != 0 ? 3 + 2 * num_container_sections : 0) +
+           (num_container_sections != 0 ? 3 + 4 * num_container_sections : 0) +
```
chfast, line 747: Use the `CONTAINER_SIZE_SIZE` constant?

lib/evmone/eof.cpp
chfast: Note to myself: this accumulation can be replaced with prefix sum to compute offsets in the same time.

lib/evmone/eof.hpp
chfast: We actually should not copy the sizes to the separate vector but read these on demand. But this change is for some other time.

lib/evmone/eof.cpp
gumb0: > I'm not sure about uint64_t. Can't we keep it as uint32_t?

We cannot, actually. In case there's huge section sizes declared, but body is truncated, we might pass the contract limit size (before this function), and get 32-bit overflow here (before body truncation is checked below).

gumb0: I've added a unit test so far.

lib/evmone/eof.cpp
```diff
@@ -40,25 +37,31 @@ constexpr auto REL_OFFSET_SIZE = sizeof(int16_t);
 constexpr auto STACK_SIZE_LIMIT = 1024;
 constexpr uint8_t NON_RETURNING_FUNCTION = 0x80;
 
-using EOFSectionHeaders = std::array<std::vector<uint16_t>, NUM_SECTIONS + 1>;
+struct EOFSectionHeaders
+{
+    uint16_t type_size = 0;
+    std::vector<uint16_t> code_sizes;
+    std::vector<uint32_t> container_sizes;
+    uint16_t data_size = 0;
```
chfast, line 45: It's better for alignment to put `data_size` after `type_size`.

chfast (approved):

chfast (approved):

PR #1175, opened 2025-03-20
precompiles: Refactor BN254 ecpairing

chfast:
Refactor the loop preparing inputs for the BN254 pairing check.

### DIFF
--- a/test/state/precompiles.cpp
+++ b/test/state/precompiles.cpp
@@ -371,50 +371,37 @@ ExecutionResult ecmul_execute(const uint8_t* input, size_t input_size, uint8_t* 
 ExecutionResult ecpairing_execute(const uint8_t* input, size_t input_size, uint8_t* output,
     [[maybe_unused]] size_t output_size) noexcept
 {
-    assert(output_size >= 32);
-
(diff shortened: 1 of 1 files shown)


test/state/precompiles.cpp
```diff
@@ -371,50 +371,37 @@ ExecutionResult ecmul_execute(const uint8_t* input, size_t input_size, uint8_t*
 ExecutionResult ecpairing_execute(const uint8_t* input, size_t input_size, uint8_t* output,
     [[maybe_unused]] size_t output_size) noexcept
 {
-    assert(output_size >= 32);
-
+    static constexpr auto OUTPUT_SIZE = 32;
     static constexpr size_t PAIR_SIZE = 192;
+    assert(output_size >= OUTPUT_SIZE);
```
rodiazet, line 376: Why not just `output_size == OUTPUT_SIZE`?

rodiazet (approved):
Added one comment

test/state/precompiles.cpp
chfast: This is generic precompile's implementation framework, sometimes it passes bigger buffers, but I don't remember why currently.

PR #1177, opened 2025-03-21
eof: Change kind_data from 0x04 to 0xff - rough version + tests

pdobacz:
Align with eof-devnet-1 changes https://github.com/ipsilon/eof/issues/173. Goes in hand with https://github.com/ethereum/execution-spec-tests/pull/1339


lib/evmone/eof.cpp
```diff
@@ -26,8 +26,12 @@ constexpr uint8_t TERMINATOR = 0x00;
 constexpr uint8_t TYPE_SECTION = 0x01;
 constexpr uint8_t CODE_SECTION = 0x02;
 constexpr uint8_t CONTAINER_SECTION = 0x03;
+// TODO: kind_data is now 0xff, but we're still using a 4-element array to hold
```
pdobacz, line 29: note that this is resolved and the TODO removed in a subsequent PR

pdobacz:
While the tests are failing, I would prefer to force merge without silencing them. It is somewhat reassuring that at least some tests pass, which we can just eyeball when merging. If we silence the tests, we'll lose that possibility. But I can also silence tests if needed.

Doing an interim EEST release in the ipsilon fork would be an option, but unfortunately, this seems to not work anymore. Final alternative is to make a proper EEST release like `v3.0.0-rc1`, filling it with an off-of-a-branch `evmone`. probably tidiest but most legwork.

Maybe that is the way to go tho :/

pdobacz:
Merging to move fast. It fills for me locally, if I arrange the `evmone` - `EEST` branches to match

PR #1181, opened 2025-04-01
eof: Change max stack height to stack height increase

chfast:
This changes the meaning of the stack height parameter of the code type from "max stack height" (including code inputs)
to "stack height increase" (excluding code inputs).

The bytecode test unitilty keeps the `max_stack_height` to minimize changes to tests.


pdobacz (approved):
lgtm, I stopped for a moment at underflow_variable_stack, but makes sense

PR #1185, opened 2025-04-07
eof: Allow TXCREATE in legacy

pdobacz:


pdobacz:
> Can you add me as commit co-author or unsquash it?

Whoops sorry, I didn't mean to un-co-author (not sure how that is derived tbh), just to correct the message

pdobacz:
@gumb0 Ok I guess it works off the commit description? I amended it, pls take a look if that's that?

gumb0 (approved):

PR #1191, opened 2025-04-11
ci: Upgrade EEST to v4.2.0 develop

chfast:
This upgrades EEST to v4.2.0. It also only runs develop fixtures because they include the stable ones.


chfast:
Upgraded to v4.3.0 in #1199.

PR #1192, opened 2025-04-11
EOF: Fix rlp_encode of InitcodeTransaction

pdobacz:


test/state/transaction.cpp
```diff
@@ -69,6 +69,17 @@ namespace evmone::state
                    tx.to.has_value() ? tx.to.value() : bytes_view(), tx.value, tx.data,
                    tx.access_list, tx.authorization_list, tx.v, tx.r, tx.s);
     }
+    else  // Transaction::Type::initcodes
```
chfast, line 72: This is kinda weird pattern that for any new tx type we need to convert a comment to `if`. Maybe change it to proper `else if` and return something dummy for the default path (can be also `unreachable()`).

chfast (approved):

pdobacz:
We need to release and update EEST to tickle the rlp encoding lines for the new tx type. Making a self not to double check that is the case

PR #1193, opened 2025-04-11
eof: Add caching of TXCREATE validation results

gumb0:


test/unittests/state_transition_txcreate_test.cpp
```diff
@@ -1170,6 +1170,42 @@ TEST_F(state_transition, txcreate_from_blob_tx)
...
+
+    const auto deploy_container = eof_bytecode(bytecode(OP_INVALID));
+
+    const auto init_code = returncode(0, 0, 0);
+    const bytecode init_container = eof_bytecode(init_code, 2).container(deploy_container);
+
+    tx.type = Transaction::Type::initcodes;
+    tx.initcodes.push_back(init_container);
+
+    constexpr auto iterations = 500;
```
pdobacz, line 1188: as mentioned on the call, I'd also make a version with invalid EOF, and with a mix valid/invalid. But maybe enough to leave it for EEST.

There is also a twisted variant where the loop is handled by recursion, where an (erroneous) cache mechanism could clear the cache on every call depth level. Also possible better in EEST.

pdobacz (approved):
LGTM. Just some test ideas, but better to implement these in EEST

lib/evmone/execution_state.hpp
```diff
@@ -216,6 +217,7 @@ class ExecutionState
         deploy_container = {};
         m_tx = {};
         m_initcodes.reset();
+        m_initcodes_validity.clear();
```
chfast, line 220: I think it is better to combine this with `m_initcodes`: to have value as `struct {bytes_view, std::optional<bool>}` or `struct {bytes_view, bool}`.

lib/evmone/instructions_calls.cpp
```diff
@@ -435,8 +435,18 @@ Result create_eof_impl(
 
     if constexpr (Op == OP_TXCREATE)
     {
-        const auto error_subcont = validate_eof(state.rev, ContainerKind::initcode, initcontainer);
-        if (error_subcont != EOFValidationError::success)
+        const auto validity = state.get_initcode_validity(initcode_hash);
```
chfast, line 438: This section is unnecessarily separated from the other `TXCREATE` processing section above. I think we should reorder them and place the endowment and depth checks before the `get_tx_initcode_by_hash()` section.

Then we will be able to combine the `get_tx_initcode_by_hash()` and `set_initcode_validity`, potentially with having single `unordered_map`.

chfast (commented):
So

lib/evmone/instructions_calls.cpp
gumb0: Problem is `pos` has to be updated before light failures
(and differently for EOFCREATE and TXCREATE)

I pushed one possible version, without unifying initcode and validity query yet.

lib/evmone/execution_state.hpp
gumb0: Combined now.

chfast (approved):

PR #1195, opened 2025-04-14
precompiles: Validate and preprocess expmod input

chfast:
Decompose the expmod input into separate bytes_views for base, exp, mod. Handle arguments padding. Handle mod being zero.

Simplify and vastly reduce the expmod stub by using the preprocessed arguments and handling trivial cases before doing the result lookup.


test/state/precompiles.cpp
```diff
@@ -317,19 +317,57 @@ ExecutionResult expmod_execute(
...
+    // - bigint libraries don't like zero moduli because division by 0 is not well-defined,
+    // - having non-zero moduli guarantees that base and exp aren't out-of-bounds.
+    if (mod_explicit.find_first_not_of(uint8_t{0}) == bytes_view::npos) [[unlikely]]
     {
+        // The modulus is zero, so the result is zero.
         std::fill_n(output, output_size, 0);
         return {EVMC_SUCCESS, output_size};
     }
 
+    const auto mod_requires_padding = mod_explicit.size() != mod_len;
```
rodiazet, line 352: I would add here some comment to make it clear why mod has to be padded in some cases.

rodiazet (approved):
Overall ok, one comment added.

test/state/precompiles.cpp
chfast: Added comment inside `if`.

PR #1198, opened 2025-04-14
precompiles: Implement expmod with GMP

chfast:


test/state/CMakeLists.txt
```diff
@@ -72,3 +72,16 @@ if(EVMONE_PRECOMPILES_SILKPRE)
         precompiles_silkpre.cpp
     )
 endif()
+
+# This is done after Silkpre because it also tries to find GMP.
+option(EVMONE_PRECOMPILES_GMP "Enable precompiles implementations via the GMP/MPIR library" ON)
```
chfast, line 77: On by default seems to cause a lot of build issues. Maybe use the same strategy as with SILKPRE (OFF by default).

test/state/CMakeLists.txt
```diff
@@ -72,3 +72,16 @@ if(EVMONE_PRECOMPILES_SILKPRE)
         precompiles_silkpre.cpp
     )
 endif()
+
+# This is done after Silkpre because it also tries to find GMP.
```
chfast, line 76: SILKPRE option stays for benchmarking purposes only.

test/state/precompiles_gmp.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace evmone::state
+{
+void expmod_gmp(bytes_view base, bytes_view exp, bytes_view mod, uint8_t* output) noexcept
+{
+    mpz_t b, e, m, r;  // NOLINT(*-isolate-declaration)
+    mpz_inits(b, e, m, r, nullptr);
+    mpz_import(b, base.size(), 1, 1, 0, 0, base.data());
+    mpz_import(e, exp.size(), 1, 1, 0, 0, exp.data());
+    mpz_import(m, mod.size(), 1, 1, 0, 0, mod.data());
+    assert(mpz_sgn(m) != 0);
```
chfast, line 18: I should document in the `expmod_xxx` interface that the `mod` is expected non-zero.

rodiazet (approved):

PR #1199, opened 2025-04-15
Handle failing system contracts

chfast:
- The EIP-7002 (withdrawal requests) and EIP-7251 (consolidation requests) have been updated to specify that a failing system contract makes the block invalid.

  - https://github.com/ethereum/EIPs/pull/9508
  - https://github.com/ethereum/EIPs/pull/9582

- Add support for minimal EIP-6110 deposit log validation following the spec update. The invalid deposit logs invalidate the block.

  - https://github.com/ethereum/EIPs/pull/9460

- Upgrade EEST to v4.3.0

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -414,6 +414,12 @@ jobs:
       - download_execution_tests:
           rev: v17.0
       - run:
+          name: "Filter-out invalid Prague tests"
+          working_directory: ~/tests
+          command: |
+            sudo apt-get -q update && sudo apt-get -qy install jq
+            find BlockchainTests -name '*.json' -exec sh -c 'jq "with_entries(select(.key | test(\"Prague\") | not))" {} > {}.tmp && mv {}.tmp {}' \;
+      - run:
           name: "State tests"
           working_directory: ~/build
           command: >
--- a/circle.yml
+++ b/circle.yml
@@ -356,19 +356,8 @@ jobs:
     steps:
       - build
       - download_execution_spec_tests:
-          release: v4.1.0
-      - run:
-          name: "Execution spec tests (state_tests)"
-          working_directory: ~/build
-          command: >
-            bin/evmone-statetest ~/spec-tests/fixtures/state_tests
-      - run:
-          name: "Execution spec tests (blockchain_tests)"
-          working_directory: ~/build
-          command: >
-            bin/evmone-blockchaintest ~/spec-tests/fixtures/blockchain_tests
-      - download_execution_spec_tests:
-          release: v4.1.0
+          release: v4.3.0
+          # develop includes stable
           fixtures_suffix: develop
       - run:
           name: "Execution spec tests (develop, state_tests)"
@@ -379,6 +368,7 @@ jobs:
       - run:
           name: "Execution spec tests (develop, blockchain_tests)"
           # Tests for in-development EVM revision currently passing.
+          # - The block_hashes_history test is disabled because it takes too long.
           working_directory: ~/spec-tests/fixtures/blockchain_tests
           command: >
             ~/build/bin/evmone-blockchaintest ~/spec-tests/fixtures/blockchain_tests
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -84,7 +84,12 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
         std::vector<state::Requests> collected;
 
         if (rev >= EVMC_PRAGUE)
-            collected.emplace_back(collect_deposit_requests(receipts));
+        {
+            auto opt_deposits = collect_deposit_requests(receipts);
+            if (!opt_deposits.has_value())
+                return std::nullopt;
+            collected.emplace_back(std::move(*opt_deposits));
+        }
 
         auto requests_result = system_call_block_end(block_state, block, block_hashes, rev, vm);
         if (!requests_result.has_value())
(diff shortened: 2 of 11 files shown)


test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -29,6 +29,7 @@ struct TransitionResult
     state::BloomFilter bloom;
     int64_t blob_gas_left;
     TestState block_state;
+    bool system_contracts_valid = true;
```
chfast, line 32: We don't check this flag yet and tests pass... We should take a look how we check blockchain test results.

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -84,15 +85,17 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
...
+    if (system_contracts_valid)
+        std::ranges::move(*r, std::back_inserter(requests));
 
     finalize(block_state, rev, block.coinbase, block_reward, block.ommers, block.withdrawals);
 
     const auto bloom = compute_bloom_filter(receipts);
 
     return {std::move(receipts), std::move(rejected_txs), std::move(requests), cumulative_gas_used,
-        bloom, blob_gas_left, std::move(block_state)};
+        bloom, blob_gas_left, std::move(block_state), system_contracts_valid};
```
pdobacz, line 98: re: the question you've asked in a comment, you should unpack the new field returned in `run_blockchain_tests` where `apply_block` is called and act on it - either `EXPECT` this is `true` for valid blocks or `continue` if this is `false` for invalid blocks.

test/blockchaintest/blockchaintest_runner.cpp
pdobacz: I added a comment to another line explaining what needs to be done https://github.com/ethereum/evmone/pull/1199#discussion_r2044113080

test/t8n/t8n.cpp
```diff
@@ -230,11 +230,13 @@ int main(int argc, const char* argv[])
 
             if (!pre_state_only && rev >= EVMC_PRAGUE)
             {
-                requests.emplace_back(collect_deposit_requests(receipts));
-
-                auto system_call_requests =
-                    system_call_block_end(state, block, block_hashes, rev, vm);
-                std::ranges::move(system_call_requests, std::back_inserter(requests));
+                // FIXME: Report invalid block if system contracts execution fails.
```
pdobacz, line 233: if you don't intend to fix in this PR or immediately after consider renaming to a TODO

pdobacz (approved):

circle.yml
```diff
@@ -376,6 +365,7 @@ jobs:
       - run:
           name: "Execution spec tests (develop, blockchain_tests)"
           # Tests for in-development EVM revision currently passing.
+          # - the block_hashes_history test is disable because takes too long.
```
gumb0, line 368:
```suggestion
          # - the block_hashes_history test is disabled because takes too long.
```

circle.yml
```diff
@@ -420,6 +410,12 @@ jobs:
       - build
       - download_execution_tests:
           rev: v17.0
+      - run:
+          name: "Filter-out invalid Prague tests"
+          working_directory: ~/tests
+          command: |
+            sudo apt-get -q update && sudo apt-get -qy install jq
+            find BlockchainTests -name '*.json' -exec sh -c 'jq "with_entries(select(.key | test(\"Prague\") | not))" {} > {}.tmp && mv {}.tmp {}' \;
```
gumb0, line 418: what does this do? filters out all Prague tests?

chfast: Yes, removes all tests with "Prague" in the name. They are not updated and fail because they are missing system contracts.

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -81,18 +84,28 @@ TransitionResult apply_block(TestState& state, evmc::VM& vm, const state::BlockI
...
+            system_contracts_failed = true;
+        else
+            requests.emplace_back(std::move(*opt_deposits));
+    }
 
-    auto system_call_requests = system_call_block_end(block_state, block, block_hashes, rev, vm);
-    std::ranges::move(system_call_requests, std::back_inserter(requests));
+    auto requests_result = system_call_block_end(block_state, block, block_hashes, rev, vm);
+    system_contracts_failed |= !requests_result.has_value();
+    if (!system_contracts_failed)
```
gumb0, line 100: here if deposits collection failed, but `system_call_block_end()` succeeded you still will skip collecting `requests_result`.

Maybe not a big problem, but also maybe makes sense to skip `system_call_block_end()` altogether if deposits collection already failed?

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -29,6 +29,9 @@ struct TransitionResult
     state::BloomFilter bloom;
     int64_t blob_gas_left;
     TestState block_state;
+
+    /// The execution of system contracts has failed what makes the block invalid.
+    bool system_contracts_failed = false;
```
gumb0, line 34: We didn't support other invalid block tests before? Maybe makes sense to name this more general "invalid_block" ?

gumb0: Ah I see other invalid cases are handled differently, so this is fine I guess

chfast: It seems the core functionality of t8n is to partition transactions into valid / rejected...

test/blockchaintest/blockchaintest_runner.cpp
chfast: Refactored...

gumb0 (approved):

PR #1219, opened 2025-05-13
test: Fix parent header handling in blockchain tests

gumb0:
When interating over block in the test, update `parent_header` only when valid block was encountered.


pdobacz (approved):

chfast (approved):

chfast:
Can we enable any more tests?

gumb0:
> Can we enable any more tests?

I enable some in https://github.com/ethereum/evmone/pull/1220

PR #1220, opened 2025-05-13
test: Add base fee and gas limit checks to blockchain tests

gumb0:


test/state/block.cpp
```diff
@@ -16,6 +19,35 @@ static constexpr uint64_t MAX_BLOB_GAS_PER_BLOCK_PRAGUE = 1179648;
 static constexpr uint64_t BLOB_GASPRICE_UPDATE_FRACTION_CANCUN = 3338477;
 static constexpr uint64_t BLOB_GASPRICE_UPDATE_FRACTION_PRAGUE = 5007716;
 
+uint64_t calc_base_fee([[maybe_unused]] int64_t block_number, int64_t parent_gas_limit,
+    int64_t parent_gas_used, uint64_t parent_base_fee) noexcept
+{
+    auto parent_gas_target = parent_gas_limit / GAS_LIMIT_ELASTICITY_MULTIPLIER;
+
+    // Special logic for block activating EIP-1559 is not implemented, because test don't cover it.
+    assert(block_number != 12'965'000);
```
gumb0, line 28: not sure if good idea

chfast: Just leave the comment, but remove the assert.

test/state/block.cpp
```diff
@@ -7,6 +7,9 @@
 
 namespace evmone::state
 {
+static constexpr auto GAS_LIMIT_ELASTICITY_MULTIPLIER = 2;
```
chfast, line 10: Maybe move to the function body?

circle.yml
```diff
@@ -412,7 +412,7 @@ jobs:
     steps:
       - build
       - download_execution_tests:
-          rev: v17.0
+          rev: v17.1
```
chfast, line 415: Why?

test/state/block.cpp
gumb0: I followed the style of this file - other constants are global

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -112,6 +112,30 @@ bool validate_block(
...
+    const auto parent_header_gas_limit_u64 = static_cast<uint64_t>(parent_header.gas_limit);
+    const auto test_block_gas_limit_u64 = static_cast<uint64_t>(test_block.block_info.gas_limit);
+    if (test_block_gas_limit_u64 >=
+        parent_header_gas_limit_u64 + parent_header_gas_limit_u64 / 1024)
+        return false;
+    if (test_block_gas_limit_u64 <=
+        parent_header_gas_limit_u64 - parent_header_gas_limit_u64 / 1024)
+        return false;
+
+    if (test_block.block_info.gas_limit < 5000)
```
chfast, line 128: What a hell?

gumb0: It's requirement from yellow paper, also mentioned in EIP-1559
image

But tests don't cover it, so should I remove?
cc @winsvega

chfast: This is funny. Does it mean an empty block with 4999 gas limit is invalid?

Maybe it comes for the gas increase/decrease limits... For low values the increase may become 0 forever.

Usually in such case I use `assert()`. But fine to leave the check too (we can later add an EEST test). Just add the comment.

PR #1232, opened 2025-05-16
Refactor PUSH implementation

chfast:
Skip the push data full word loading loop with `if constexpr`. This improves source code coverage.


pdobacz (approved):

lib/evmone/instructions.hpp
```diff
@@ -878,32 +878,36 @@ inline uint64_t load_partial_push_data<4>(code_iterator pos) noexcept
...
+        pos += NUM_PARTIAL_BYTES;
     }
 
     // Load full words.
-    for (size_t i = 0; i < num_full_words; ++i)
+    if constexpr (NUM_FULL_WORDS != 0)
     {
-        r[num_full_words - 1 - i] = intx::be::unsafe::load<uint64_t>(data);
-        data += sizeof(uint64_t);
+        for (size_t i = 0; i != NUM_FULL_WORDS; ++i)
```
gumb0, line 903: Curios why the change from `i < num_full_words` to `i != NUM_FULL_WORDS`, seems non-conventional.

chfast: I think this is just pedantic iterator abstraction (some iterators are not less-than comparable). I will bring it back.

PR #1233, opened 2025-05-16
Mark the EVM stack pointer as 32-byte aligned

chfast:
This is guaranteed by the stack space allocation.

This allows the compiler to use aligned vector load instructions. Unfortunately, no performance gains on latest CPUs because aligned/unaligned instructions have the same cost there.


chfast:
> Why do this then?

It provides some additional information to the compiler. Maybe handy on other architectures.

gumb0 (approved):

gumb0 (approved):

chfast:
No changes on ARM64 either :cry:

PR #1241, opened 2025-05-21
test: Support withdrawals blockchain tests with fields wider than uint64

gumb0:
There are InvalidBlocks tests that check that Withdrawal's fields fit uint64.

Also note that the tests with fields wider than uint256 are being deleted: https://github.com/ethereum/legacytests/pull/14

### DIFF
--- a/test/statetest/statetest_loader.cpp
+++ b/test/statetest/statetest_loader.cpp
@@ -85,7 +85,10 @@ bytes from_json<bytes>(const json::json& j)
 template <>
 address from_json<address>(const json::json& j)
 {
-    return evmc::from_hex<address>(j.get<std::string>()).value();
+    const auto v = evmc::from_hex<address>(j.get<std::string>());
+    if (!v.has_value())
+        throw std::invalid_argument("from_json<address>: must be hexadecimal string");
+    return *v;
 }
 
 template <>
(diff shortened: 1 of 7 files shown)


test/statetest/statetest_loader.cpp
```diff
@@ -85,7 +102,10 @@ bytes from_json<bytes>(const json::json& j)
 template <>
 address from_json<address>(const json::json& j)
 {
-    return evmc::from_hex<address>(j.get<std::string>()).value();
+    const auto v = evmc::from_hex<address>(j.get<std::string>());
+    if (!v.has_value())
```
gumb0, line 106: This just improves the exception being thrown if address failed to parse.

test/state/block.hpp
```diff
@@ -18,15 +18,16 @@ struct Ommer
 
 struct Withdrawal
 {
-    uint64_t index = 0;
-    uint64_t validator_index = 0;
+    std::optional<uint64_t> index;
```
chfast, line 21: I don't like this change because it mixes the data de-serialization error with a logic error. I think, the `Withdrawal` struct represent the correctly de-serialized object. Otherwise, I cannot expose such API do end users (and I plan to). Is there any better option handle this?

PR #1242, opened 2025-05-28
precompiles: Add modexp changes from EIP-7823 and EIP-7883

rodiazet:
This PR implements two EIPs related expmod precompile changes in Osaka fork. [EIP-7823](https://eips.ethereum.org/EIPS/eip-7823) and [EIP-7883](https://eips.ethereum.org/EIPS/eip-7883)


chfast (commented):
You should enable EEST tests on CI. This may require moving EOF out of the way.

PR #1248, opened 2025-05-30
test: Add exhaustive benchmarks for modexp

chfast:


test/precompiles_bench/precompiles_bench.cpp
```diff
@@ -194,6 +195,57 @@ void precompile(benchmark::State& state)
...
+    const auto exp_len = static_cast<size_t>(state.range(1));
+
+    const auto payload_len = 2 * base_mod_len + exp_len;
+    const auto input_len = 3 * 32 + payload_len;
+    const auto input = std::make_unique_for_overwrite<uint8_t[]>(input_len);
+    intx::be::unsafe::store(&input[0], intx::uint256{base_mod_len});
+    intx::be::unsafe::store(&input[32], intx::uint256{exp_len});
+    intx::be::unsafe::store(&input[64], intx::uint256{base_mod_len});
+    std::fill_n(&input[96], payload_len, 0xff);
+    input[input_len - 1] = 0xfe;  // Make mod even.
```
chfast, line 210: This will make the base == mod + 1 so the answer should be 1 because we are constantly multiplying 1. Not sure this is good benchmark then. Maybe set the mod to 0xff..fc so that initial base % mod is 3.

test/precompiles_bench/precompiles_bench.cpp
```diff
@@ -194,6 +195,57 @@ void precompile(benchmark::State& state)
     state.counters["gas_rate"] = Counter(static_cast<double>(total_gas_used), Counter::kIsRate);
 }
 
+void modexp(benchmark::State& state)
+{
+    const auto base_mod_len = static_cast<size_t>(state.range(0));
+    const auto exp_len = static_cast<size_t>(state.range(1));
```
chfast, line 201: Maybe we want to have bit-level precision for exp length?

PR #1253, opened 2025-06-11
evmmax: BN254 point addition in Montgomery form.

rodiazet:
This PR extends bn254 point addition interface to be able to avoid unnecessary to/from Montgomery form conversions.

### DIFF
--- a/lib/evmone_precompiles/ecc.hpp
+++ b/lib/evmone_precompiles/ecc.hpp
@@ -26,6 +26,18 @@ struct Point
 static_assert(Point<unsigned>{}.is_inf());
 
 template <typename IntT>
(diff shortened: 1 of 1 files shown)


lib/evmone_precompiles/ecc.hpp
```diff
@@ -25,6 +25,18 @@ struct Point
 
 static_assert(Point<unsigned>{}.is_inf());
 
+template <typename IntT>
+inline Point<IntT> to_mont(ModArith<IntT> m, const Point<IntT>& p)
```
chfast:
```suggestion
inline Point<IntT> to_mont(const ModArith<IntT>& m, const Point<IntT>& p)
```

rodiazet:
No needed ATM.

PR #1274, opened 2025-07-21
t8n: Add `--opcode.count` option

chfast:
Closes https://github.com/ipsilon/evmone/issues/1256.


marioevz:
Giving it a try today 🫡

chfast:
No worries, take as much time as you need.

marioevz:
Works perfectly, I spent some time trying to think if there could be a better interface but I think this should work just fine.

Link to the EEST PR that adds support to this: https://github.com/ethereum/execution-spec-tests/pull/1956

PR #1292, opened 2025-08-13
modexp: Skip the leading zeros in the exponent

chfast:
Introduce the Exponent view type to represent the exponent bytes as an array of bits. Skip the leading exponent zero bytes and bits.

### DIFF
--- a/lib/evmone_precompiles/modexp.cpp
+++ b/lib/evmone_precompiles/modexp.cpp
@@ -34,41 +34,69 @@ constexpr unsigned ctz(const intx::uint<N>& x) noexcept
     return tz;
 }
 
+/// Represents the exponent value of the modular exponentiation operation.
+///
+/// This is a view type of the big-endian bytes representing the bits of the exponent.
+class Exponent
+{
+    const uint8_t* data_ = nullptr;
+    size_t bit_width_ = 0;
+
+public:
+    explicit Exponent(std::span<const uint8_t> bytes) noexcept
+    {
+        const auto it = std::ranges::find_if(bytes, [](auto x) { return x != 0; });
+        const auto trimmed_bytes = std::span{it, bytes.end()};
+        bit_width_ = trimmed_bytes.empty() ? 0 :
+                                             static_cast<size_t>(std::bit_width(trimmed_bytes[0])) +
+                                                 (trimmed_bytes.size() - 1) * 8;
+        data_ = trimmed_bytes.data();
+    }
+
+
+    [[nodiscard]] size_t bit_width() const noexcept { return bit_width_; }
+
+    /// Returns the bit value of the exponent at the given index, counting from the most significant
+    /// bit (e[0] is the top bit).
+    bool operator[](size_t index) const noexcept
+    {
+        // TODO: Replace this with a custom iterator type.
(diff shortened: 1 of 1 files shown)


kmk142789 (approved):

lib/evmone_precompiles/modexp.cpp
```diff
@@ -34,41 +34,63 @@ constexpr unsigned ctz(const intx::uint<N>& x) noexcept
     return tz;
 }
 
+class Exponent
```
rodiazet, line 40: Please add a doc comments.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -34,41 +34,63 @@ constexpr unsigned ctz(const intx::uint<N>& x) noexcept
...
+    }
+
+    [[nodiscard]] size_t num_bits() const noexcept { return num_bits_; }
+
+    bool operator[](size_t index) const noexcept
+    {
+        // TODO: Replace this with a custom iterator type.
+        const auto exp_size = (num_bits_ + 7) / 8;
+        const auto byte_index = index / 8;
+        const auto byte = data_[exp_size - 1 - byte_index];
```
rodiazet, line 66: Why not just this index is being counted from the end?

lib/evmone_precompiles/modexp.cpp
```diff
@@ -34,41 +34,63 @@ constexpr unsigned ctz(const intx::uint<N>& x) noexcept
...
     {
-        for (size_t i = 8; i != 0; --i)
-        {
-            ret = arith.mul(ret, ret);
-            const auto bit = (e >> (i - 1)) & 1;
-            if (bit != 0)
-                ret = arith.mul(ret, base_mont);
-        }
+        ret = arith.mul(ret, ret);
+        if (exp[i - 1])
```
rodiazet, line 83: Isn't the first always `1`? We can optimize this and make the loop shorter.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -34,41 +34,63 @@ constexpr unsigned ctz(const intx::uint<N>& x) noexcept
...
     {
-        for (size_t i = 8; i != 0; --i)
-        {
-            ret *= ret;
-            const auto bit = (e >> (i - 1)) & 1;
-            if (bit != 0)
-                ret *= base;
-        }
+        ret *= ret;
+        if (exp[i - 1])
```
rodiazet, line 98: The same optimization is possible here I think.

rodiazet (approved):
LGTM. Some nits found. I would document the the type and its member functions.

lib/evmone_precompiles/modexp.cpp
chfast: That's correct, but I don't like mixing optimization changes. This requires separate PR.

lib/evmone_precompiles/modexp.cpp
chfast: Do you mean why `exp[0]` means the highest bit instead the lowest bit? It's because the loop goes this way. The interface can be the opposite, but the in the loop you will have to invert it.

PR #1312, opened 2025-09-11
crypto: Switch to Jacobian coordinates in ECC

chfast:
Replace using of projective coordinates for elliptic curve point to Jacobian coordinates. Jacobian formulas are faster, easier to use and some of the furmulas are independent of the a parameter of a curve (should be handy for implementing secp256r1).

### DIFF
--- a/lib/evmone_precompiles/ecc.hpp
+++ b/lib/evmone_precompiles/ecc.hpp
@@ -72,6 +72,11 @@ struct FieldElement
         return wrap(Fp.sub(a.value_, b.value_));
     }
 
+    friend constexpr auto operator-(const FieldElement& a) noexcept
+    {
+        return wrap(Fp.sub(0, a.value_));
+    }
+
     friend constexpr auto operator/(one_t, const FieldElement& a) noexcept
     {
         return wrap(Fp.inv(a.value_));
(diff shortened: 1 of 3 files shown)


lib/evmone_precompiles/ecc.hpp
```diff
@@ -226,175 +255,143 @@ AffinePoint<Curve> add(const AffinePoint<Curve>& p, const AffinePoint<Curve>& q)
...
+    const auto z2z2 = z2 * z2;
+    const auto u1 = x1 * z2z2;
+    const auto u2 = x2 * z1z1;
+    const auto z1z1z1 = z1 * z1z1;
+    const auto z2z2z2 = z2 * z2z2;
+    const auto s1 = y1 * z2z2z2;
+    const auto s2 = y2 * z1z1z1;
+    const auto h = u2 - u1;
+    const auto r = s2 - s1;
+    assert(r != 0);            // TODO: Untested.
```
chfast: This is actually impossible. I will have a follow up PR which detects equal points in more efficient way.

lib/evmone_precompiles/ecc.hpp
```diff
@@ -226,175 +255,142 @@ AffinePoint<Curve> add(const AffinePoint<Curve>& p, const AffinePoint<Curve>& q)
...
-    x3 = t2 - x3;
-    y3 = y3 * t0;
-    t1 = t1 * z3;
-    y3 = t1 + y3;
-    t0 = t0 * t3;
-    z3 = z3 * t4;
-    z3 = z3 + t0;
+    const auto z1z1 = z1 * z1;
+    const auto u2 = x2 * z1z1;
+    const auto t0 = z1 * z1z1;
```
rodiazet: why not `z1z1z1` but `t0`?

chfast: I translated this "script": https://www.hyperelliptic.org/EFD/g1p/auto-code/shortw/jacobian/addition/madd.op3. But I'll change it.

PR #1322, opened 2025-09-24
tests: Fix blockchaintest EIP-7918

pdobacz:
as title, EIP-7918 formula was incorrectly implemented, should be:

```
... OB * get_base_fee_per_blob_gas(parent):
...  * (blobSchedule.max - blobSchedule.target) // blobSchedule.max      <==== not target
... 
...  - target_blob_gas
```

This activates one skipped test

### DIFF
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -166,6 +166,8 @@ bool validate_block(
             return false;
 
         // Check that the excess blob gas was updated correctly.
+        // According to EIP-7918 current blocks schedule (`rev`) should be used for parent base fee
+        // calculation.
         const auto parent_blob_base_fee =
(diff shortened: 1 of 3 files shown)


test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -166,6 +166,8 @@ bool validate_block(
             return false;
 
         // Check that the excess blob gas was updated correctly.
+        // According to EIP-7918 current blocks schedule (`rev`) should be used for parent base fee
```
pdobacz, line 169: took the liberty to add this, I've spent some time thinking this is the cause, as it is unintuitively the _correct_ revision used.

chfast: It is unfortunate we have this logic in `test/blockchaintest` which is filtered out in #1318.

PR #1329, opened 2025-10-01
tests: Support EIP-7934 - MAX_RLP_BLOCK_SIZE

pdobacz:
https://eips.ethereum.org/EIPS/eip-7934

Since `evmone` doesn't decode RLP blocks, but always uses `rlp_decoded`, the `rlp` bytes are not managed, just the size used for the limit check.

### DIFF
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -5,13 +5,19 @@
 #include "../state/mpt_hash.hpp"
 #include "../state/requests.hpp"
 #include "../state/rlp.hpp"
-#include "../state/system_contracts.hpp"
 #include "../test/statetest/statetest.hpp"
 #include "blockchaintest.hpp"
 #include <gtest/gtest.h>
 
 namespace evmone::test
 {
+
+/// The CL gossip protocol constraint of the maximum block size (EIP-7934).
+constexpr size_t MAX_BLOCK_SIZE = 10 * 1024 * 1024;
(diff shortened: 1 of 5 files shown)


test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -13,6 +13,10 @@
 namespace evmone::test
 {
 
+constexpr size_t MAX_BLOCK_SIZE = 10'485'760;  // EIP-7934
```
chfast: Explain these more. Maybe use format like `10*1024*1024`

pdobacz:
@chfast re the coverage miss - EEST tests do not count towards coverage of `blockchaintest` files, so the new is missed (along with many others in this file. How should we approach this?

chfast:
> @chfast re the coverage miss - EEST tests do not count towards coverage of `blockchaintest` files, so the new line is missed (along with many others in this file.) How should we approach this?

This logic should not be there, but it is fine for now. I will do manual review for the coverage report. This will also check the coverage of the edge values.

chfast (approved):

chfast:
> > This logic should not be there, but it is fine for now. I will do manual review for the coverage report. This will also check the coverage of the edge values.
> 
> Do you mean by that that we should look towards refactoring all the block/state logic into a new lib (not to `evmone` I guess, to keep it EVM only) or that that particular check should be elsewhere?

Yes. I'm thinking about this for a long time. It is mostly for the new API targeting full transactions. But we also have some validation code which we can expose (can be the same lib).

PR #1330, opened 2025-10-02
 tests: Support EIP-7892 - BPO forks 

### LINKED ISSUE #1325: Add support for EIP-7892 BPO forks
Conform to https://eips.ethereum.org/EIPS/eip-7892, in order to pass EEST v5.1.0.

Suggested logic:
1. EEST `blockchain_test` fixtures now come with a new `blobSchedule` (as specced out by the EIP) and a `network` field, which is sth like `OsakaToBPO1AtTime15k`
2. For tests which have `blobSchedule`, it will be parsed to produce a mapping `fork -> blob params`
3. The `network` field, if one of the predefined transitions like `OsakaToBPO1AtTime15k`, will be recognized. The transition time will be used to find blob params in that mapping, for each block.
4. If not found in the map, hardcoded blob params will be used as before (only up to Prague)
5. For purposes other than blob params, "forks" like `OsakaToBPO1AtTime15k` are just treated as `Osaka`, being the most recent non-BPO fork preceding the active fork for blob params

Beyond EEST v5.1.0, `network` field parsing (3.) may be substituted by parsing `<Fork>Time`  fields similar to format given in the EIP itself, if that is chosen to be put in the `blockchain_test` fixture format. (5.) would then become irrelevant as special `network` values would not be used anymore.

pdobacz:
https://eips.ethereum.org/EIPS/eip-7892

Closes #1325 and follows the proposal there.

Blob parameter only forks. This ended up a considerable revamp, as we've moved on from hardcoded blob schedules to ones coming from JSON config (post Osaka). Because of this, a bunch of internal APIs needed to be shuffled, I hope this all still makes sense.

To ease review, the "main" commit follows the old naming convention, which seems to have bin a misnomer - it took a single entry of a blob schedule for the blob schedule. The two follow-up commits, to be squashed away before merging, clean this up.

But, this PR should also be reviewable as a whole, just need to follow the new naming.


test/state/blob_schedule.cpp
```diff
@@ -0,0 +1,46 @@
+#include "blob_schedule.hpp"
+#include "../utils/utils.hpp"
+#include <cstdint>
+#include <string_view>
+
+namespace evmone::state
+{
+BlobParams get_blob_params(evmc_revision rev)
```
pdobacz, line 8: NOTE: this used to be `constexpr` but it cannot be anymore, since I've decided to overload `get_blob_params`. I don't think the constexprness has been leveraged anywhere, and now it also throws so probably doesn't even make sense to be constexpr.. But yea, lmk if this a problem

pdobacz:
I'm puzzled by why do things like `unittests` or `t8n` still come up on the coverage report, I left them on the ignore list. This is the last thing to tackle here I guess, but otherwise ready for review

test/state/blob_schedule.hpp
```diff
@@ -14,19 +14,25 @@ constexpr auto GAS_PER_BLOB = 0x20000;  // 2**17
...
 };
 
-/// Returns the blob schedule for the given EVM revision.
-constexpr BlobSchedule get_blob_schedule(evmc_revision rev) noexcept
-{
-    if (rev >= EVMC_PRAGUE)
-        return {6, 9, 5007716};
-    return {3, 6, 3338477};
-}
+using BlobSchedule = std::unordered_map<std::string, state::BlobParams>;
```
chfast, line 25: Focusing on the interface for EVM here, I think the mapping should be rather `timestamp => BlobParams` so this is independent of the `revision` and the network name.

So we should use something like `flat_map` to represent `BlobSchedule`.

Getting the `BlobSchedule` by name or from JSON should be the job for the EVM user. I imagine this will be hardcoded for Mainnet and constructed from JSON for tests.

I'm not sure how much refactoring this is, so we can leave it for some other task. Maybe when we will be putting better boundary between test and consensus code.

chfast: Now I realized that passing `BlobParams` instead of `BlobSchedule` is more in-line with passing `revision`.

pdobacz: > Focusing on the interface for EVM here, I think the mapping should be rather `timestamp => BlobParams` so this is independent of the `revision` and the network name.

So I thought about this today. What troubles me, is whether or not we'd still want to fetch `BlobParams` by fork explicitly, bypassing timestamp. More generally, I'm not 100% sure about the boundary between EEST and evmone here - in the contexts of state/blockchaintests, t8n, and potentially the future Engine API test consumption.

I think it's ok to merge as is to tick off coverage work, but I can try the refactor afterwards and we can decide the API considerations with the testing team.



> Now I realized that passing `BlobParams` instead of `BlobSchedule` is more in-line with passing `revision`.

Not sure what do you mean? passing where?

chfast: I meant the same as your concern.

chfast (approved):
You can rework commits and we can merge this.

test/integration/t8n/CMakeLists.txt
```diff
@@ -211,7 +211,7 @@ add_test(
 )
 string(
     JOIN ".*" EXPECTED_OUT
-    # config was read for Osaka fork
+    # resolved blob parameters were
```
chfast: typo?

test/statetest/statetest.hpp
```diff
@@ -91,7 +91,7 @@ template <>
 bytes from_json<bytes>(const json::json& j);
 
 state::BlockInfo from_json_with_rev(
-    const json::json& j, evmc_revision rev, const state::BlobSchedule& blob_schedule = {});
+    const json::json& j, evmc_revision rev, const state::BlobParams& blob_params);
```
chfast: You can pass `BlobParams` by value as it fits into single register.

PR #1331, opened 2025-10-07
tests: discover state/blockchaintests by individual test

pdobacz:
Tackles the point (3.) from github.com/ethereum/execution-spec-tests/pull/2243 (`consume direct` using `evmone` in EEST).

Here we switch statetest and blockchaintest runners to discover gtests by parsing the JSON files and recognizing each test.

The downside is that the invocation we use in CI (just `...evmone-statetest ... fixtures/state_tests`) became very chatty, so `--gtest_brief=1` was added. There is an option to only go inside the JSONs when a single `.json` file path is given (as `consume direct` does), but this would complicate things quite a bit, requiring us to have more code, but not terribly much more. 

Makes `consume direct` for evmone version in PR mentioned :point_down:  work correctly


pdobacz:
It seems that discovering tests individually has more problems than I thought initially, if you apply it always. Problems which arise:
1. When calling with a dir with many tests it becomes quite unresponsive, as it discovers the tests for a long time before giving any output
2. It becomes problematic to filter out tests

Would it work for us if discover was done as it used to be (per file) when provided with a dir path (e.g. like we do in our CI) and discover per test when provided with a single `.json` file path (e.g. like what `consume direct` uses)? @chfast

test/blockchaintest/blockchaintest.cpp
```diff
@@ -78,7 +117,10 @@ int main(int argc, char* argv[])
         app.set_version_flag("--version", "evmone-blockchaintest " EVMONE_VERSION);
 
         std::vector<std::string> paths;
-        app.add_option("path", paths, "Path to test file or directory")
+        app.add_option("path", paths,
+               "Path to test file or directory. For a directory, all .json "
+               "files (except index.json) are considered test files.and each file is treated as a "
```
chfast:
```suggestion
               "files (except index.json) are considered test files, and each file is treated as a "
```

circle.yml
```diff
@@ -399,7 +399,7 @@ jobs:
           working_directory: ~/build
           command: >
             LLVM_PROFILE_FILE=state_tests.profraw
-            bin/evmone-statetest ~/spec-tests/fixtures/state_tests
+            bin/evmone-statetest ~/spec-tests/fixtures/state_tests --gtest_brief=1
```
chfast: Why is this helpful now?

pdobacz: ah, it is not! it used to be when this printed all the tests, but I reverted that. Will revert here too.

test/statetest/statetest.cpp
```diff
@@ -93,7 +129,10 @@ int main(int argc, char* argv[])
         app.set_version_flag("--version", "evmone-statetest " EVMONE_VERSION);
 
         std::vector<std::string> paths;
-        app.add_option("path", paths, "Path to test file or directory")
+        app.add_option("path", paths,
+               "Path to test file or directory. For a directory, all .json "
+               "files (except index.json) are considered test files.and each file is treated as a "
```
chfast: Same typo here.

PR #1338, opened 2025-10-21
fix: Remove potentially dead, uncovered code in ExecutionState

pdobacz:
Coverage picks up the line in execution_state.hpp as uncovered, because it is used only in advanced execution. But maybe that can do away with the same method baseline uses?

Treat this more as a question - I'm unsure if these methods are actually fully equivalent. @chfast ?


pdobacz:
ah, the other way around would also seem to work and make the line covered properly - we can make baseline use the `.data` method which advanced uses. LMK

PR #1341, opened 2025-10-21
Refactor MPT implementation to get more precise code coverage

chfast:


hadlojen-dev:
It's hard, and I don't have the workers.

PR #1346, opened 2025-10-26
fix: remove dead code - secp256k1 affine point mul

pdobacz:
This PR is more of a question - is it OK to move this code into the unittest convenience wrapper? It is not used in any code covered by precompile tests.


test/unittests/evmmax_secp256k1_test.cpp
```diff
@@ -10,6 +10,18 @@ using namespace evmmax::secp256k1;
 using namespace evmc::literals;
 using namespace evmone::test;
 
+namespace
+{
+// Convenience wrapper for point multiplication test: Scalar multiplication in secp256k1.
```
chfast: Keep the /// comments.

PR #1347, opened 2025-10-26
fix: check block difficulty in blockchaintest

pdobacz:
Not sure if we still care for difficulty validation logic coverage, but I went down the rabbit hole and here it is :shrug:


chfast:
Oh, please compare with https://github.com/ipsilon/evmone/pull/1244.

PR #1348, opened 2025-10-28
refactor: filter relevant functionality in call_impl

pdobacz:
Some pieces of `call_impl` were missing coverage because they are never run for some `<Op>` instantiations of the `call_impl` template. This makes the coverage report harder to read.

Not 100% sure this is elegant, wdyt?

### DIFF
--- a/lib/evmone/instructions_calls.cpp
+++ b/lib/evmone/instructions_calls.cpp
@@ -80,9 +80,11 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
     static_assert(
         Op == OP_CALL || Op == OP_CALLCODE || Op == OP_DELEGATECALL || Op == OP_STATICCALL);
 
+    static constexpr bool HAS_VALUE_ARG = Op == OP_CALL || Op == OP_CALLCODE;
+
     const auto gas = stack.pop();
     const auto dst = intx::be::trunc<evmc::address>(stack.pop());
-    const auto value = (Op == OP_STATICCALL || Op == OP_DELEGATECALL) ? 0 : stack.pop();
+    const auto value = (!HAS_VALUE_ARG) ? 0 : stack.pop();
     const auto has_value = value != 0;
     const auto input_offset_u256 = stack.pop();
     const auto input_size_u256 = stack.pop();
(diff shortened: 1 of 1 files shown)


lib/evmone/instructions_calls.cpp
```diff
@@ -135,40 +137,57 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
         msg.input_size = input_size;
     }
 
-    auto cost = has_value ? CALL_VALUE_COST : 0;
-
-    if constexpr (Op == OP_CALL)
+    if constexpr (may_have_value)
```
pdobacz: the additional indents from the `if constexpr`s below cause a big diff, but the bodies were left untouched

lib/evmone/instructions_calls.cpp
```diff
@@ -80,9 +80,11 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
     static_assert(
         Op == OP_CALL || Op == OP_CALLCODE || Op == OP_DELEGATECALL || Op == OP_STATICCALL);
 
+    constexpr bool may_have_value = Op == OP_CALL || Op == OP_CALLCODE;
```
chfast:
```suggestion
    static constexpr bool HAS_VALUE_ARG = Op == OP_CALL || Op == OP_CALLCODE;
```

lib/evmone/instructions_calls.cpp
```diff
@@ -135,40 +137,57 @@ Result call_impl(StackTop stack, int64_t gas_left, ExecutionState& state) noexce
...
     }
 
     if (state.msg->depth >= 1024)
         return {EVMC_SUCCESS, gas_left};  // "Light" failure.
 
-    if (has_value && intx::be::load<uint256>(state.host.get_balance(state.msg->recipient)) < value)
-        return {EVMC_SUCCESS, gas_left};  // "Light" failure.
+    if constexpr (HAS_VALUE_ARG)
+    {
+        if (has_value &&
```
chfast: You can swap the balance check with the depth check to merge two `if constexpr`. The depth check was earlier because it is lighter, but it doesn't matter actually because the limit cannot be reached in practice.

PR #1353, opened 2025-10-29
Remove EOF implementation

chfast:
Complete removal of the EVM Object Format version 1 (EOFv1, the only version implemented) specified by the [EIP-7692](https://eips.ethereum.org/EIPS/eip-7692). This includes:
- implementation of EOF validation and execution,
- implementation of `DUPN`, `SWAPN` and `EXCHANGE` instructions,
- "initcodes" transaction type,
- additional testing tools: `eofparse`, `eoftest` and `eofparsefuzz`,
- unit tests, integration tests, benchmarks and test helpers.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -315,14 +315,9 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
 
     if (!code.empty())
     {
-        if (code[0] == 0xEF)
-        {
-            if (m_rev >= EVMC_LONDON)
-            {
-                // EIP-3541: Reject EF-prefixed code.
-                return evmc::Result{EVMC_CONTRACT_VALIDATION_FAILURE};
-            }
-        }
+        // EIP-3541: Reject new contract code starting with the 0xEF byte.
+        if (m_rev >= EVMC_LONDON && code[0] == 0xEF)
+            return evmc::Result{EVMC_CONTRACT_VALIDATION_FAILURE};
 
         new_acc->code_hash = keccak256(code);
         new_acc->code = code;


test/state/host.cpp
```diff
@@ -373,18 +317,9 @@ evmc::Result Host::create(const evmc_message& msg) noexcept
...
-            {
-                // Only EOFCREATE/TXCREATE is allowed to deploy code starting with
-                // EF. It must be valid EOF, which was validated before execution.
-                if (msg.kind != EVMC_EOFCREATE)
-                    return evmc::Result{EVMC_CONTRACT_VALIDATION_FAILURE};
-                assert(validate_eof(m_rev, ContainerKind::runtime, code) ==
-                       EOFValidationError::success);
-            }
-            else if (m_rev >= EVMC_LONDON)
+            if (m_rev >= EVMC_LONDON)
```
pdobacz: can collapse the 2 ifs into an `&&` one?

test/unittests/evm_eip663_dupn_swapn_test.cpp
```diff
@@ -1,188 +0,0 @@
...
-    execute(eof_bytecode(pushes + OP_SWAP16 + ret_top(), 33));
-    EXPECT_STATUS(EVMC_SUCCESS);
-    EXPECT_OUTPUT_INT(17);
-
-    execute(eof_bytecode(pushes + OP_SWAPN + "0f" + ret_top(), 33));
-    EXPECT_STATUS(EVMC_SUCCESS);
-    EXPECT_OUTPUT_INT(17);
-}
-
-TEST_P(evm, dupn_swapn_undefined_in_legacy)
```
pdobacz: do we need to keep these? or are they handled in some generic tests working off of the instruction definitions? If the latter then we can remove the `undefined_in_legacy` tests in other files too

pdobacz (approved):
Just 2 small nitpicks

test/unittests/evm_eip663_dupn_swapn_test.cpp
chfast: Yes, I want to keep these tests, so I will move them somewhere. Have you spot more of these?

I also wanted to keep "extcode" tests, but realized these are not valid as you cannot have code starting with `EF` without EOF.

pdobacz: yea, all named with `undefined_in_legacy` I believe, so easy to grep.

chfast: Tests moved in previous PR #1357.

PR #1355, opened 2025-10-30
refactor: move compute_mod_inv out of the template

pdobacz:
Pardon my c++ if this isn't the correct way.

Coverage is picking up the `static constepxr` method as not covered for uint::384, because the template is instantiated but never called (we're using a library for bls12-384, IIUC). 

This seems to fix the problem with coverage and compile fine, and also is the way to do such a thing I've found explained online, but please correct me if wrong.

### DIFF
--- a/include/evmmax/evmmax.hpp
+++ b/include/evmmax/evmmax.hpp
@@ -8,6 +8,22 @@
 namespace evmmax
 {
 
+/// Compute the modulus inverse for Montgomery multiplication, i.e. N': mod⋅N' = 2⁶⁴-1.
+///
+/// @param mod0  The least significant word of the modulus.
+constexpr uint64_t compute_mod_inv(uint64_t mod0) noexcept
+{
+    // TODO: Find what is this algorithm and why it works.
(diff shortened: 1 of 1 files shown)


include/evmmax/evmmax.hpp
```diff
@@ -8,6 +8,22 @@
 namespace evmmax
 {
 
+/// Compute the modulus inverse for Montgomery multiplication, i.e. N': mod⋅N' = 2⁶⁴-1.
+///
+/// @param mod0  The least significant word of the modulus.
+inline constexpr uint64_t compute_mod_inv(uint64_t mod0) noexcept
```
chfast:
```suggestion
constexpr uint64_t compute_mod_inv(uint64_t mod0) noexcept
```

I believe `constexpr` implies `inline`.

chfast (approved):
This seems good. Otherwise, we have multiple copies of the same method.

include/evmmax/evmmax.hpp
pdobacz: indeed, thx

PR #1358, opened 2025-10-30
fix: constexpr away impossible condition on gas_cost

pdobacz:
Coverage uncovered what seems to be an impossible condition - checking if instruction is defined for instructions which are known to be defined since Frontier.

My only doubt is - what if an instruction becomes undefined at some point - would this code fail to recognize it?

### DIFF
--- a/lib/evmone/baseline_execution.cpp
+++ b/lib/evmone/baseline_execution.cpp
@@ -51,16 +51,20 @@ inline evmc_status_code check_requirements(const CostTable& cost_table, int64_t&
     auto gas_cost = instr::gas_costs[EVMC_FRONTIER][Op];  // Init assuming const cost.
     if constexpr (!instr::has_const_gas_cost(Op))
     {
-        gas_cost = cost_table[Op];  // If not, load the cost from the table.
+        gas_cost = cost_table[Op];  // If not, load the cost from the current revision cost table.
 
         // Negative cost marks an undefined instruction.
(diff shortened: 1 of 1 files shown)


lib/evmone/baseline_execution.cpp
```diff
@@ -51,16 +51,20 @@ inline evmc_status_code check_requirements(const CostTable& cost_table, int64_t&
...
-        gas_cost = cost_table[Op];  // If not, load the cost from the table.
+        gas_cost = cost_table[Op];  // If not, load the cost from the current revision cost table.
 
         // Negative cost marks an undefined instruction.
-        // This check must be first to produce correct error code.
-        if (INTX_UNLIKELY(gas_cost < 0))
-            return EVMC_UNDEFINED_INSTRUCTION;
+        // This check must be the first to produce the correct error code.
+        // By definition not possible if defined since the first revision.
+        if constexpr (instr::traits[Op].since == EVMC_FRONTIER)
```
chfast: I used the opcode trait for the check. Please double check this is correct.

pdobacz: so I was pondering this, but opted for `gas_costs` since this is what the assertion above used, and also the entire check here. But if `traits` sounds good to you, I guess it is more natural, :+1:

chfast: Slightly more readable to me.

PR #1359, opened 2025-10-31
crypto: Optimize computing mod inversion for Montgomery

chfast:


include/evmmax/evmmax.hpp
```diff
@@ -8,20 +8,23 @@
 namespace evmmax
 {
 
-/// Compute the modulus inverse for Montgomery multiplication, i.e. N': mod⋅N' = 2⁶⁴-1.
+/// Compute the modulus inverse for Montgomery multiplication, i.e., N': mod⋅N' = 2⁶⁴-1.
```
rodiazet: I would change `mod` -> `M`. Now it's confusing with `K mod N'`

include/evmmax/evmmax.hpp
```diff
@@ -8,20 +8,23 @@
...
-    uint64_t base = 0 - mod0;
-    uint64_t result = 1;
-    for (auto i = 0; i < 64; ++i)
-    {
-        result *= base;
-        base *= base;
-    }
-    return result;
+    assert(mod0 % 2 == 1);
+    // Compute inversion mod 2⁶⁴ (inv⋅mod0 = 1) using the Newton–Raphson numeric method.
```
rodiazet: Can we link a reference with this algorithm?

include/evmmax/evmmax.hpp
```diff
@@ -8,20 +8,23 @@
...
+    // Each iteration doubles the number of correct bits: 2, 4, 8, ...
+    // so for 64-bit value we need 6 iterations.
+    // TODO(C++23): static
+    constexpr auto ITERATIONS = std::countr_zero(sizeof(mod0) * 8);
+    uint64_t inv = 1;  // Start with inversion mod 2.
+    for (auto i = 0; i < ITERATIONS; ++i)
+        inv *= 2 - mod0 * inv;
```
rodiazet: Doesn't this overflow or maybe it does and it's ok?

include/evmmax/evmmax.hpp
```diff
@@ -8,20 +8,23 @@
 namespace evmmax
 {
 
-/// Compute the modulus inverse for Montgomery multiplication, i.e. N': mod⋅N' = 2⁶⁴-1.
+/// Compute the modulus inverse for Montgomery multiplication, i.e., N': mod⋅N' = 2⁶⁴-1.
 ///
 /// @param mod0  The least significant word of the modulus.
 constexpr uint64_t compute_mod_inv(uint64_t mod0) noexcept
```
rodiazet: Why `mod0`?

include/evmmax/evmmax.hpp
chfast: Overflows are fine because this gives `mod 2^64`. I will add a comment.

include/evmmax/evmmax.hpp
chfast: The least significant word of `mod`. We can change this to something generic because this context belongs to the user, not here.

chfast: But also `mod0` is not the same as `mod` in the top comment.

include/evmmax/evmmax.hpp
chfast: But isn't the modulus named `N`?

include/evmmax/evmmax.hpp
```diff
@@ -7,21 +7,21 @@
...
-    for (auto i = 0; i < 64; ++i)
-    {
-        result *= base;
-        base *= base;
-    }
-    return result;
+    assert(a % 2 == 1);  // The argument must be odd, otherwise the inverse does not exist.
+
+    // Use the Newton–Raphson numeric method, see e.g.
+    // https://maths-people.anu.edu.au/~brent/pd/mca-cup-0.5.9.pdf#section.4.2
```
chfast: Better reference: https://gmplib.org/~tege/divcnst-pldi94.pdf#page=9, formula (9.2)

PR #1361, opened 2025-10-31
fix: constexpr away LOG0 topcis and gas==0 OOG

pdobacz:
I've started wondering if this isn't overdoing it. The gain is somewhat more readable coverage report - one doesn't have to stop by each uncovered line to "tick it off" as "yea, this cannot be covered, ever". I suspect the compiler handles this anyway and optimizes away, so there should be no performance gains (? not sure). The downside is just lines of code and source of obscure bugs in case our assumptions about opcodes somehow change (? unlikely?).

LMK WDYT

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -925,8 +925,11 @@ inline Result log(StackTop stack, int64_t gas_left, ExecutionState& state) noexc
         return {EVMC_OUT_OF_GAS, gas_left};
 
     std::array<evmc::bytes32, NumTopics> topics;  // NOLINT(cppcoreguidelines-pro-type-member-init)
-    for (auto& topic : topics)
-        topic = intx::be::store<evmc::bytes32>(stack.pop());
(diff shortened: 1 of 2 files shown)


lib/evmone/instructions.hpp
```diff
@@ -925,8 +925,9 @@ inline Result log(StackTop stack, int64_t gas_left, ExecutionState& state) noexc
         return {EVMC_OUT_OF_GAS, gas_left};
 
     std::array<evmc::bytes32, NumTopics> topics;  // NOLINT(cppcoreguidelines-pro-type-member-init)
-    for (auto& topic : topics)
-        topic = intx::be::store<evmc::bytes32>(stack.pop());
+    if constexpr (NumTopics > 0)
```
chfast, line 928: Add `{}`.

chfast: Alternatively, we could use `std::transform` here, this will hide the loop. But while you are working on coverage this is fine.

PR #1367, opened 2025-11-12
modexp: Optimize the exponentiation loop

chfast:
Use the information that in the first iteration of the exponentiation loop the bit is aways 1.


chfast:
All of these is superseded by the current master branch.

PR #1404, opened 2025-12-18
crypto: Rename ecc::add() → add_affine()

chfast:
Make it explicit in the prodedure name that this addition includes a finite field inversion on the way to produce the result in affine coordinates.

### DIFF
--- a/lib/evmone_precompiles/ecc.hpp
+++ b/lib/evmone_precompiles/ecc.hpp
@@ -236,8 +236,11 @@ inline AffinePoint<Curve> to_affine(const ProjPoint<Curve>& p) noexcept
 /// Elliptic curve point addition in affine coordinates.
 ///
 /// Computes P ⊕ Q for two points in affine coordinates on the elliptic curve.
+/// This procedure handles all inputs (e.g. doubling or points at infinity).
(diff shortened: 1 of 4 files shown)


lib/evmone_precompiles/ecc.hpp
```diff
@@ -236,8 +236,11 @@ inline AffinePoint<Curve> to_affine(const ProjPoint<Curve>& p) noexcept
 /// Elliptic curve point addition in affine coordinates.
 ///
 /// Computes P ⊕ Q for two points in affine coordinates on the elliptic curve.
+/// This procedure handles all inputs (e.g. doubling or points at infinity).
+/// The produced result is also in affine coordinates. Therefore, this is useful only for one-off
+/// additions, as otherwise multiple additions would be inefficient due to repeated inversions.
 template <typename Curve>
-AffinePoint<Curve> add(const AffinePoint<Curve>& p, const AffinePoint<Curve>& q) noexcept
+AffinePoint<Curve> add_to_affine(const AffinePoint<Curve>& p, const AffinePoint<Curve>& q) noexcept
```
chfast: Maybe `add_as_affine` or `affine_add`?

chfast:
@copilot, don't you think the name `add_affine` would be better?

PR #1429, opened 2026-01-30
Implement DUPN/SWAPN/EXCHANGE (EIP-8024)

frangio:
Implements the DUPN (`0xe6`), SWAPN (`0xe7`), and EXCHANGE (`0xe8`) stack instructions for the Amsterdam revision as specified in [EIP-8024](https://eips.ethereum.org/EIPS/eip-8024) ([`c0aa8d7`](https://github.com/ethereum/EIPs/commit/c0aa8d79038e0bf60ab8f15ff32ef17ba3b80882)).

### Changes

**Opcodes & traits** (`instructions_opcodes.hpp`, `instructions_traits.hpp`, `instructions_xmacro.hpp`):
- Register `OP_DUPN`, `OP_SWAPN`, `OP_EXCHANGE` with gas cost 3, activation at `EVMC_AMSTERDAM`.
- `stack_height_change=+1` for DUPN enables the overflow check in `check_requirements`.
- `stack_height_required=0` for all three — underflow depends on the decoded immediate and is checked in each handler.

**Immediate decoding** (`instructions.hpp`):
- `decode_dupn_swapn_imm()` → `std::optional<int>`: validates the forbidden range (`0x5b..0x7f`) and decodes `n = (imm + 0x91) mod 256` in one step.
- `decode_exchange_imm()` → `std::optional<std::pair<int, int>>`: validates the forbidden range (`0x52..0x7f`) and decodes `(n, m)` via XOR + divmod-16 grid.

**Handlers** (`instructions.hpp`):
- `dupn`: decode, check underflow (`n > stack_size`), push `stack[n-1]`.
- `swapn`: decode, check underflow (`n >= stack_size`), `fast_swap(top, stack[n])`.
- `exchange`: decode, check underflow (`m >= stack_size`), `fast_swap(stack[n], stack[m])`.
- `fast_swap` uses the manual 4-limb copy to match the existing `swap<N>` workaround for clang ([llvm#59116](https://github.com/llvm/llvm-project/issues/59116)).

**Advanced interpreter** (`advanced_instructions.cpp`):
- Opcodes mapped to `op_undefined` — not implemented in advanced (per maintainer decision).

**JUMPDEST analysis**: Unchanged. The EIP's forbidden immediate range ensures `0x5b` after these opcodes always causes an invalid-immediate halt, preserving `0x5b` as a valid JUMPDEST.

### Tests

- Basic execution, end-of-code (implicit `0x00` immediate), invalid immediates (single + boundary sweep), stack overflow/underflow for all three opcodes, gas boundary (OOG at gas-1, success at exact gas).
- `exchange_max_m`: regression for decode off-by-one at `(n=1, m=29)` — the same bug that caused Nethermind's consensus failure on bal-devnet-2.
- `dupn_immediate_0x5b_is_jumpdest`: JUMPDEST compatibility — `0x5b` in the immediate position is a valid jump target in both Amsterdam and pre-Amsterdam.
- `push1_stack_overflow`: regression ensuring the stack overflow check is not bypassed for PUSH instructions.
- Pre-Amsterdam undefined instruction tests (`evm_undefined_instructions_test.cpp`) updated from `EVMC_MAX_REVISION` to `EVMC_OSAKA`.

### DIFF
--- a/lib/evmone/instructions.hpp
+++ b/lib/evmone/instructions.hpp
@@ -74,8 +74,8 @@ constexpr void fast_swap(uint256& x, uint256& y) noexcept
     y[3] = t3;
 }
 
-/// Decode DUPN/SWAPN immediate. Returns the stack depth n (17..235),
-/// or std::nullopt if the immediate is in the forbidden range (0x5b..0x7f).
+/// Decode DUPN/SWAPN immediate. Returns the stack depth n [17–235],
+/// or std::nullopt if the immediate is in the forbidden range [0x5b–0x7f].
 constexpr std::optional<int> decode_dupn_swapn_imm(uint8_t imm) noexcept
 {
     if (imm >= 0x5b && imm <= 0x7f)
@@ -84,7 +84,7 @@ constexpr std::optional<int> decode_dupn_swapn_imm(uint8_t imm) noexcept
 }
 
 /// Decode EXCHANGE immediate. Returns the pair (n, m) with 1 <= n < m and n + m <= 30,
-/// or std::nullopt if the immediate is in the forbidden range (0x52..0x7f).
+/// or std::nullopt if the immediate is in the forbidden range [0x52–0x7f].
 constexpr std::optional<std::pair<int, int>> decode_exchange_imm(uint8_t imm) noexcept
 {
     if (imm >= 0x52 && imm <= 0x7f)
--- a/test/unittests/evm_eip8024_swapn_dupn_exchange_test.cpp
+++ b/test/unittests/evm_eip8024_swapn_dupn_exchange_test.cpp
@@ -40,7 +40,7 @@ TEST_P(evm, dupn_invalid_immediate)
         return;
 
     rev = EVMC_AMSTERDAM;
-    // 0x5b is in the forbidden range (0x5b..0x7f).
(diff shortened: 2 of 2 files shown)


lib/evmone/advanced_analysis.cpp
```diff
@@ -193,6 +202,27 @@ AdvancedCodeAnalysis analyze(evmc_revision rev, bytes_view code) noexcept
...
+            const auto valid = (opcode == OP_EXCHANGE) ?
+                                   evmone::instr::imm::is_valid_exchange(imm) :
+                                   evmone::instr::imm::is_valid_dupn_swapn(imm);
+            if (valid)
+            {
+                if (code_pos < code_end)
+                    ++code_pos;
+            }
+            else
+                rollback_block();
```
frangio: This rollback approach doesn't look great to me.

@chfast What do you think? How should "conditional" stack change be implemented?

It's not exactly conditional either because either the stack height changes (on DUPN) or the frame reverts due to invalid instruction.

chfast: You can ignore the advanced interpreter (disable it in relevant unit tests).

frangio: Removed changes to the advanced interpreter.

PR #1458, opened 2026-02-24
precompiles: Implement expmod via OpenSSL

chfast:
The modexp precompile implementation using OpenSSL.


chfast:
This is nice if you already have OpenSSL or similar as a dependency. But this was ~2x slower than GMP implementation.

PR #1536, opened 2026-05-22
precompiles: Optimize KZG point evaluation via joint G1 MSM

chfast:
Rearrange the standard KZG verification equation

    e(C - [y]₁, [1]₂) =? e(π, [s - z]₂)

into

    e(C + [z]π - [y]₁, [1]₂) =? e(π, [s]₂)

by moving the [z] factor from G2 to G1 via the pairing's bilinearity.

The new form:
  - eliminates the G2 scalar multiplication entirely (was [z](-G2_gen)),
  - eliminates the G2 point addition (was [s]G2 - [z]G2),
  - merges the two remaining G1 scalar multiplications ([z]π and [y]₁)
    into a single 2-point Pippenger MSM,
  - keeps both G2 pairing arguments as fixed constants ([1]₂ via
    blst_aggregated_in_g1's precomputed lines and the [s]₂ point from
    the trusted setup).

G2_GENERATOR_NEGATIVE and the blst_p2 mult/add helpers are no longer
needed. G1_GENERATOR_NEGATIVE is redeclared as blst_p1_affine since
the Pippenger MSM consumes affine points.

### DIFF
--- a/lib/evmone_precompiles/kzg.cpp
+++ b/lib/evmone_precompiles/kzg.cpp
@@ -12,29 +12,12 @@ namespace evmone::crypto
 {
 namespace
 {
-/// The field element 1 in Montgomery form.
-constexpr blst_fp ONE = {0x760900000002fffd, 0xebf4000bc40c0002, 0x5f48985753c758ba,
-    0x77ce585370525745, 0x5c071a97a256ec6d, 0x15f65ec3fa80e493};
-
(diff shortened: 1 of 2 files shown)


chfast:
```
precompile<PrecompileId::point_evaluation,_evmone_blst>-14        46.63M ± 0%   50.75M ± 0%  +8.85% (p=0.000 n=11)
```

PR #1537, opened 2026-05-22
precompiles: Precompute Miller loop lines for KZG trusted setup point

chfast:
Precompute the Miller-loop lines for the trusted-setup point [s]₂
and use these (instead of the point) for faster KZG pairing check.
The precomputation has been done outside and only the result is stored.
The static data size is ~19 KB.


chfast:
```
precompile<PrecompileId::point_evaluation,_evmone_blst>-14        50.75M ± 0%   54.25M ± 1%  +6.89% (p=0.000 n=11)
```

PR #1539, opened 2026-05-22
precompiles: Add Fq2 sqr() and use it in BN254 pairing helpers

chfast:


chfast:
```
precompile<PrecompileId::ecpairing,_evmmax_cpp>-14        86.92M ± 0%   92.11M ± 0%  +5.98% (p=0.000 n=11)
```

PR #1541, opened 2026-05-25
crypto: Optimize line evaluation in BN254 pairings

chfast:
Algebraic simplifications in the Jacobian add + line evaluation,
all recovering an intermediate reuse that was implicit in the original
reference but lost when the procedure was re-derived to fix the paper's
typos.


rodiazet (approved):
LGTM. Quite obvious simplification. Wondering why these papers don’t mention this. :)

chfast:
```
precompile<PrecompileId::ecpairing,_evmmax_cpp>-14        92.18M ± 1%   92.73M ± 0%  +0.59% (p=0.001 n=11)
```

PR #1544, opened 2026-05-26
crypto: Share Fq12 squaring across pairs in BN254 Miller loop

chfast:
pairing_check ran an independent Miller loop per pair, each paying 64 Fq12
squarings, and multiplied the results. The f <- f^2 * line recurrence is
multiplicative over the pairs, so a single accumulator serves them all,
dropping (N-1)*64 squarings and the N-1 Fq12 multiplications that combined
the per-pair results.

Measured on the ECPAIRING benchmark inputs, which carry 2 to 4 pairs:
instructions -13.4%, cycles -13.7%, IPC unchanged at 1.80. The saving grows
with the pair count, towards about a quarter of the cycles.

The only per-pair state the loop needs is the running point, so those are
held in a vector parallel to the input and the pairs with a point at
infinity are skipped in place. A single pair breaks even: the allocation
replaces the Fq12 multiplication that combined the per-pair results.
Building the running points up front also defers the Miller loop until
every pair has been validated, so an invalid pair no longer costs the
Miller loops of the pairs preceding it.

### DIFF
--- a/lib/evmone_precompiles/pairing/bn254/pairing.cpp
+++ b/lib/evmone_precompiles/pairing/bn254/pairing.cpp
@@ -5,6 +5,7 @@
 #include "../../bn254.hpp"
 #include "fields.hpp"
 #include "utils.hpp"
+#include <span>
 #include <vector>
 
 namespace evmmax::bn254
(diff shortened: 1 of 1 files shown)


chfast:
```
                                                   │ /proc/self/fd/11 │          /proc/self/fd/16           │
                                                   │      gas/s       │    gas/s     vs base                │
precompile<PrecompileId::ecpairing,_evmmax_cpp>-14        106.0M ± 0%   122.9M ± 1%  +16.02% (p=0.000 n=11)
```

PR #1547, opened 2026-05-26
crypto: Migrate ExtPoint to AffinePoint<E2> 

chfast:
Use `AffinePoint<ExtFieldElem>` instead of `ExtPoint` (now removed). This consolidates and de-duplicates code.

### DIFF
--- a/test/state/precompiles.cpp
+++ b/test/state/precompiles.cpp
@@ -551,7 +551,7 @@ ExecutionResult ecpairing_execute(const uint8_t* input, size_t input_size, uint8
     for (auto input_ptr = input; input_ptr != input + input_size; input_ptr += PAIR_SIZE)
     {
         namespace bn = evmmax::bn254;
-        auto p = bn::AffinePoint::from_bytes(std::span<const uint8_t, 64>{input_ptr, 64});
+        const auto p = bn::AffinePoint::from_bytes(std::span<const uint8_t, 64>{input_ptr, 64});
         if (!p.has_value()) [[unlikely]]
             return {EVMC_PRECOMPILE_FAILURE, 0};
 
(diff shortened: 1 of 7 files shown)


test/state/precompiles.cpp
```diff
@@ -550,18 +550,21 @@ ExecutionResult ecpairing_execute(const uint8_t* input, size_t input_size, uint8
...
+        // G2 EVM ABI feeds the imaginary coefficient before the real one for each Fq²,
+        // so swap the offsets when reading into (real, imaginary) order.
+        const auto qx_real = bn::Fq::from_bytes(std::span<const uint8_t, 32>{input_ptr + 96, 32});
+        const auto qx_imag = bn::Fq::from_bytes(std::span<const uint8_t, 32>{input_ptr + 64, 32});
+        const auto qy_real = bn::Fq::from_bytes(std::span<const uint8_t, 32>{input_ptr + 160, 32});
+        const auto qy_imag = bn::Fq::from_bytes(std::span<const uint8_t, 32>{input_ptr + 128, 32});
+        if (!qx_real || !qx_imag || !qy_real || !qy_imag) [[unlikely]]
+            return {EVMC_PRECOMPILE_FAILURE, 0};
+        bn::ExtPoint q{bn::Fq2({*qx_real, *qx_imag}), bn::Fq2({*qy_real, *qy_imag})};
+        pairs.emplace_back(std::move(*p), std::move(q));
```
chfast: Move on p and q doesn't make sense. Drop it and use `const auto` as before.

test/unittests/evmmax_bn254_pairing_test.cpp
```diff
@@ -24,39 +24,26 @@ TEST(evmmax, bn254_pairing)
...
     const auto Q1 = ExtPoint{
-        {
-            0x04bf11ca01483bfa8b34b43561848d28905960114c8ac04049af4b6315a41678_u256,
-            0x209dd15ebff5d46c4bd888e51a93cf99a7329636c63514396b4a452003a35bf7_u256,
-        },
-        {
-            0x120a2a4cf30c1bf9845f20c6fe39e07ea2cce61f0c9bb048165fe5e4de877550_u256,
-            0x2bb8324af6cfc93537a2ad1a445cfd0ca2a71acd7ac41fadbf933c2a51be344d_u256,
-        },
+        Fq2({Fq(0x04bf11ca01483bfa8b34b43561848d28905960114c8ac04049af4b6315a41678_u256),
```
chfast: I don't like we have to modify the constructor. Can we fix it as in AffinePoint where you can pass compile-time literals to constructor (they are going to be converted to Montgomery at compile-time).

PR #1549, opened 2026-05-27
crypto: Precompute Miller-loop lines for G2 generator in KZG verify

chfast:
It turned out the `blst_aggregated_in_g1` doesn't use precomputed lines for G2 generator point. Do it ourselves.


chfast:
```
precompile<PrecompileId::point_evaluation,_evmone_blst>-14        54.25M ± 1%   58.62M ± 0%  +8.06% (p=0.000 n=11)
```

PR #1551, opened 2026-05-27
tools: Fold evmone-t8n into the evmone CLI as a subcommand

chfast:
Replace the standalone evmone-t8n binary with `evmone t8n`. The
standalone executable is removed.

The command line options are the same, but now parsed by CLI11,
which makes validation strict at parse time:

- --state.reward range-checked via CLI::Range(-1, INT64_MAX); -1
  remains the "emit pre-state only" sentinel.
- --input.alloc/env/txs/blobParams must exist (CLI::ExistingFile);
  previously the standalone binary would open and read an empty/
  failed stream.
- Unknown options now error; previously the hand-rolled if-else
  chain silently ignored them.


test/integration/t8n/CMakeLists.txt
```diff
@@ -254,15 +253,29 @@ set_tests_properties(
...
-    ${PREFIX}/version PROPERTIES PASS_REGULAR_EXPRESSION "evmone-t8n")
+    ${PREFIX}/version PROPERTIES PASS_REGULAR_EXPRESSION "evmone")
 
-add_test(NAME ${PREFIX}/bad_fork COMMAND evmone-t8n --state.fork NoSuchRev)
+add_test(NAME ${PREFIX}/bad_fork COMMAND evmone-cli t8n --state.fork NoSuchRev)
 set_tests_properties(
     ${PREFIX}/bad_fork PROPERTIES PASS_REGULAR_EXPRESSION "unknown revision")
 
 
+# Exercises CLI11 integer parsing on the upper bound and 0x-hex form:
```
chfast:
```suggestion
# Check if chainid max value is accepted.
```

PR #1556, opened 2026-05-29
test: Match blockchain test exceptions

chfast:
The blockchain test runner treated block validity as a bool: for a block expected to be invalid it accepted ANY rejection and never checked that the reason matched the fixture's `expectException`. So a block rejected for the wrong reason still passed.

Make `validate_block` return a `std::error_code` instead of `bool`, with new block-level entries appended (at the back, keeping existing values stable) to `state::ErrorCode`. Each entry's message is the matching execution-spec-tests `BlockException` constant (INVALID_GASLIMIT, INVALID_BASEFEE_PER_GAS, INCORRECT_EXCESS_BLOB_GAS, RLP_BLOCK_LIMIT_EXCEEDED, INVALID_BLOCK_TIMESTAMP_OLDER_THAN_PARENT, INCORRECT_BLOCK_FORMAT for the rest). The loader captures the `expectException` string; the runner substring-matches the actual error's constant against it, which also handles the fixtures' `|`-separated alternatives.

Legacy ethereum/tests fixtures use older, sometimes finer-grained exception names (e.g. uncles after Paris, a number gap, a wrong base fee). The loader maps those (many-to-one) onto the `BlockException` constant evmone reports, so they match by reason as well.

### DIFF
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -4,6 +4,7 @@
 
 #include "blockchaintest_runner.hpp"
 #include <gtest/gtest.h>
+#include <test/state/errors.hpp>
 #include <test/state/ethash_difficulty.hpp>
 #include <test/state/requests.hpp>
 #include <test/utils/mpt_hash.hpp>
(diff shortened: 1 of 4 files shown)


test/utils/blockchaintest_loader.cpp
```diff
@@ -166,6 +166,7 @@ BlockchainTest load_blockchain_test_case(const std::string& name, const json::js
 
             auto test_block = load_test_block(el.at("rlp_decoded"), bt.network, bt.blob_schedule);
             test_block.valid = false;
+            test_block.expected_exception = it->get<std::string>();
```
chfast: This can replace valid flag.

test/utils/blockchaintest.hpp
```diff
@@ -52,6 +52,9 @@ struct TestBlock
     size_t rlp_size = 0;
     bool withdrawals_parse_success = true;
     bool valid = true;
+    /// The fixture's `expectException` string for an invalid block (empty if valid).
+    /// Used by the runner to match against the actual validation error.
+    std::string expected_exception;
```
chfast: Shorten to one line `///< Empty for valid blocks.`

test/state/errors.hpp
```diff
@@ -32,6 +32,17 @@ enum ErrorCode : int  // NOLINT(*-use-enum-class)
     EMPTY_AUTHORIZATION_LIST,
     MAX_GAS_LIMIT_EXCEEDED,
     UNKNOWN_ERROR,
+
+    // Block-level validation errors (not tied to a single transaction). Appended
```
chfast: One line should be enough: // Block-level validation.

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -317,9 +321,10 @@ void run_blockchain_tests(std::span<const BlockchainTest> tests, evmc::VM& vm)
 
             if (test_block.valid)
             {
-                ASSERT_TRUE(
-                    validate_block(rev, blob_params, test_block, parent_header, parent_has_ommers))
-                    << "Expected block to be valid (validate_block)";
+                const auto block_error =
```
chfast: extract block_error out of if-else (don't repeat).

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -125,75 +126,78 @@ TransitionResult apply_block(const TestState& state, evmc::VM& vm, const state::
         bloom, blob_gas_left, std::move(block_state)};
 }
 
-bool validate_block(evmc_revision rev, state::BlobParams blob_params, const TestBlock& test_block,
-    const BlockHeader* parent_header, bool parent_has_ommers) noexcept
+/// Validates block-level validity unrelated to individual txs (see `apply_block`).
```
chfast:
```suggestion
/// Validates block-level validity unrelated to individual transactions.
///
```

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -125,75 +126,78 @@ TransitionResult apply_block(const TestState& state, evmc::VM& vm, const state::
...
+/// Returns an empty error_code if the block is valid, otherwise the specific
```
chfast:
```suggestion
/// Returns an empty error_code if the block is valid, otherwise the specific validation error.
```

Don't mention matching EEST. This is bad place for this.

test/blockchaintest/blockchaintest_runner.cpp
```diff
@@ -377,8 +382,19 @@ void run_blockchain_tests(std::span<const BlockchainTest> tests, evmc::VM& vm)
...
+                    // Block correctly rejected at validation; verify the reason matches the
+                    // fixture's expected exception. The error message is the `BlockException`
+                    // constant; `expected_exception` may list `|`-separated alternatives, so a
+                    // substring search suffices as long as no constant name is a substring of
+                    // another (true for the constants evmone produces).
+                    EXPECT_NE(test_block.expected_exception.find(block_error.message()),
+                        std::string::npos)
+                        << "Block invalidity reason mismatch: got " << block_error.message()
+                        << ", expected " << test_block.expected_exception;
                     continue;
```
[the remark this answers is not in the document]
chfast: https://github.com/ipsilon/evmone/issues/1557

PR #1580, opened 2026-07-06
test: Name the shared RLP framing constants

chfast:
Introduce the RLP prefix bases (0x80 for strings, 0xc0 for lists) and the short-form length limit (Yellow Paper Appendix B) as named constants, replacing the magic numbers in the length encoder. A short item's prefix is base + length, matching the encoder's ShortBase template parameter and the spec's "value 0x80 plus the length" wording; the long-form base (base + SHORT_LENGTH_LIMIT) is derived where used, so only the two short bases need naming. The transaction decoder added next reuses these.

The constants live in test/state/rlp_common.hpp because later we will introduce RLP decoder there.

### DIFF
--- a/test/state/rlp_common.hpp
+++ b/test/state/rlp_common.hpp
@@ -1,0 +1,20 @@
+// evmone: Fast Ethereum Virtual Machine implementation
+// Copyright 2026 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+/// @file
+/// RLP prefix encoding constants (Yellow Paper Appendix B), shared by the encoder and the decoder.
+
+#include <cstddef>
+#include <cstdint>
+
+namespace evmone::rlp
+{
+/// The largest payload encoded in the short form; longer payloads use the long form.
+constexpr size_t SHORT_LENGTH_LIMIT = 55;
+/// Base of a byte-string prefix: a short string is this byte plus its length (0x80..0xb7).
+constexpr uint8_t SHORT_STRING_BASE = 0x80;
+/// Base of a list prefix: a short list is this byte plus its payload length (0xc0..0xf7).
+constexpr uint8_t SHORT_LIST_BASE = 0xc0;
+}  // namespace evmone::rlp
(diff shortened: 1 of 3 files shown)


test/utils/rlp.hpp
```diff
@@ -19,26 +20,26 @@ using evmc::bytes_view;
 
 namespace internal
 {
-template <uint8_t ShortBase, uint8_t LongBase>
+template <uint8_t ShortBase>
 inline bytes encode_length(size_t l)
 {
-    static constexpr uint8_t short_cutoff = 55;
-    static_assert(ShortBase + short_cutoff <= 0xff);
+    constexpr uint8_t LONG_BASE = ShortBase + SHORT_LENGTH_LIMIT;
```
chfast:
```suggestion
    static constexpr uint8_t LONG_BASE = ShortBase + SHORT_LENGTH_LIMIT;
```

test/state/rlp_common.hpp
```diff
@@ -0,0 +1,19 @@
...
+// Copyright 2026 The evmone Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace evmone::rlp
+{
+// RLP prefix encoding (Yellow Paper Appendix B), shared by the encoder and the decoder.
```
chfast: Make it proper file-level doxygen.

test/utils/rlp.hpp
```diff
@@ -59,11 +60,11 @@ inline decltype(rlp_encode(std::declval<T>())) encode(const T& v)
 
 inline bytes encode(bytes_view data)
 {
-    static constexpr uint8_t short_base = 128;
-    if (data.size() == 1 && data[0] < short_base)
+    if (data.size() == 1 && data[0] < SHORT_STRING_BASE)
         return {data[0]};
 
-    return internal::encode_length<short_base, 183>(data.size()) += data;  // Op + not available.
+    // Op + not available.
```
chfast: Don't move this comment.

PR #1582, opened 2026-07-08
precompiles: Handle allocation failure in call_precompile

ywy2090:
The result of the nothrow output buffer allocation in `call_precompile()` was not checked (noted by the existing `TODO: handle nullptr`). On allocation failure the precompile execution would write to a null pointer. Return `EVMC_INTERNAL_ERROR` instead.


chfast (changes requested):
Not a good fix because this is not certain anyone will handle `EVMC_INTERNAL_ERROR`. We should redesign memory allocation for precompiles instead.

PR #1583, opened 2026-07-08
state: Cache repeated and negative account lookups

ywy2090:
This resolves the TODO in `State::find()`.

Host callbacks look up the same account repeatedly: SLOAD/SSTORE perform two Host callbacks per instruction (`access_storage` followed by `get_storage`/`set_storage`), each doing a full hash-map lookup of the same address. This adds a single-entry cache of the last successful lookup so the second callback only compares the address. The cached pointer is stable (`unordered_map` references) and is invalidated on the account erase in `rollback()`.

Lookups of non-existent accounts (e.g. BALANCE/EXTCODESIZE of an empty address, repeated in a loop) previously hit the initial `StateView` on every query — with a real database backend each of those is a repeated trie/DB query. Cache the negative results in a set; `insert()` removes the entry when the account comes into existence.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -272,6 +272,7 @@ StateDiff State::build_diff(evmc_revision rev) const
 
 Account& State::insert(const address& addr, Account account)
 {
+    m_absent.erase(addr);  // The account exists now.
     const auto r = m_modified.insert({addr, std::move(account)});
     assert(r.second);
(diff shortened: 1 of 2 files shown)


chfast (changes requested):
You missed "If we want to cache non-existent account we need a proper flag for it.".

PR #1584, opened 2026-07-08
state: Avoid redundant lookups and allocations

ywy2090:
A collection of small optimizations on the state transition hot paths:

- Skip the initial state storage query for accounts without initial storage (e.g. newly created contracts): the value is guaranteed to be zero. A constructor storing N slots previously made N pointless `StateView::get_storage()` queries.
- Bind the sender account reference once in `execute_message()` instead of three `get()` lookups per value-transferring call (references to `unordered_map` elements are stable, inserting the recipient doesn't invalidate them).
- Look up the 0x03 quirk account on the revert path only for revisions where the quirk applies (Spurious Dragon and later).
- Return early from `access_account()` for precompiles: they are always warm (EIP-2929) and no state entry is needed for them. Note a small observable difference: untouched precompile addresses no longer appear in the state diff as no-op modified/deleted entries (the resulting post-state is identical).
- Build the EIP-7702 delegation designation in a stack buffer instead of concatenating temporary heap allocations per authorization.
- Reserve `StateDiff::modified_accounts`, pass `bytes_view` by value.

### DIFF
--- a/test/state/host.cpp
+++ b/test/state/host.cpp
@@ -364,11 +364,13 @@ evmc::Result Host::execute_message(const evmc_message& msg) noexcept
 
             // Transfer value: sender → recipient.
             // The sender's balance is already checked therefore the sender account must exist.
+            // References to accounts are stable, inserting the recipient doesn't affect them.
             const auto value = intx::be::load<intx::uint256>(msg.value);
-            assert(m_state.get(msg.sender).balance >= value);
-            m_state.journal_balance_change(msg.sender, m_state.get(msg.sender).balance);
+            auto& sender_acc = m_state.get(msg.sender);
+            assert(sender_acc.balance >= value);
+            m_state.journal_balance_change(msg.sender, sender_acc.balance);
             m_state.journal_balance_change(msg.recipient, dst_acc.balance);
-            m_state.get(msg.sender).balance -= value;
+            sender_acc.balance -= value;
             dst_acc.balance += value;
 
             if (m_rev >= EVMC_AMSTERDAM)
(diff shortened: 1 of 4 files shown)


ywy2090:
The execution-spec-tests / precompiles-gmp / precompiles-libsecp256k1 failures were a real bug in the first push: the access-list warming loop in `transition()` relied on `access_account()` materializing the account before `get_storage()` is called, which the precompile early-return broke — an access list naming a precompile **with storage keys** (EEST `eip2930_access_list.test_transaction_intrinsic_gas_cost`) then dereferenced a null account and crashed.

Fixed by materializing the account in the warming loop when storage keys are present, and added a unit test (`state_transition.access_list_precompile_with_storage_keys`) reproducing the crash (verified under ASan+UBSan: SEGV before the fix, 1130/1130 tests pass after).

test/state/host.cpp
```diff
@@ -402,15 +404,20 @@ evmc::Result Host::call(const evmc_message& orig_msg) noexcept
     if (result.status_code != EVMC_SUCCESS)
     {
         static constexpr auto addr_03 = 0x03_address;
-        auto* const acc_03 = m_state.find(addr_03);
-        const auto is_03_touched = acc_03 != nullptr && acc_03->erase_if_empty;
+        // The 0x03 quirk applies only from Spurious Dragon, don't look the account up earlier.
+        bool is_03_touched = false;
+        if (m_rev >= EVMC_SPURIOUS_DRAGON)
```
chfast, line 409: This is not worth it because we mostly care about the performance in the recent forks. However, you can check if it is enough to apply the quirk only in the single fork (where it happen on Mainnet).

test/state/state.cpp
```diff
@@ -333,7 +337,9 @@ StorageValue& State::get_storage(const address& addr, const bytes32& key)
     // TODO: Avoid account lookup by giving the reference to the account's storage to Host.
     auto& acc = get(addr);
     const auto [it, missing] = acc.storage.try_emplace(key);
-    if (missing)
+    // Query the initial state only if the account may have storage there;
+    // otherwise the initial value is guaranteed to be zero (the default).
+    if (missing && acc.has_initial_storage)
```
chfast, line 342: This looks like edge case which in unlikely in practice and we don't use `.has_initial_storage` as it is only needed for CREATE collisions (another edge case).

chfast (commented):
Generally these look good (except for inline comments), but please submit them as separate PRs (you can reuse this PR for one of them).

chfast:
Split into multiple separate changes: #1601, #1602, #1603, #1605, #1606, #1607.

chfast:
Big thanks!

PR #1608, opened 2026-07-21
test: Regression-test top-level tx execution at sender nonce 2^64-2

cavdarahmet:
Closes #1495 (I think — see note below).

A top-level CALL or CREATE transaction from a sender at nonce `2^64-2` must
execute normally; only `nonce == 2^64-1` is invalid per EIP-2681. #1495
reported that evmone incorrectly aborted such transactions as a "light
failure" after bumping the sender's nonce.

I could not reproduce the bug on current `master`. It looks like it was
fixed incidentally by #1589 ("Compute the create address in the VM"),
which moved the EIP-2681 nonce check out of the path shared by all
depth-0 messages and into `create_impl`, which only runs for nested,
bytecode-triggered `CREATE`/`CREATE2`. That PR doesn't mention #1495 or
EIP-2681, so this may be worth double-checking on your end.

Since there was no test covering this scenario, I added two (`CALL` and
`CREATE`, both at sender nonce `2^64-2`) so a future refactor can't
silently reintroduce the bug.

Tested with `ctest -R "evmone/unittests"`: 1133/1133 passed.

### DIFF
--- a/test/unittests/state_transition_eip7702_test.cpp
+++ b/test/unittests/state_transition_eip7702_test.cpp
@@ -4,6 +4,7 @@
 
 #include "../utils/bytecode.hpp"
 #include "state_transition.hpp"
+#include <evmone/constants.hpp>
 
 using namespace evmc::literals;
 using namespace evmone::test;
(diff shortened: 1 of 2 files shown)


chfast:
Thanks for the report and test. I'm aware of the issue, but it may take some time to fix it (it has lower priority due to unrachability in real networks).

cavdarahmet:
Thanks for the reply! Just to clarify — #1495 itself already looks resolved (fixed by #1589, confirmed by bisecting with evmone-cli t8n), so it's probably fine to close that one. Thanks for looking into it!

PR #1616, opened 2026-07-27
state: Recognize chain-id protected legacy transaction

chfast:
The chain id check exempted every legacy transaction with chain_id 0, because the decoder folded the wire v into (chain_id, y_parity), and that makes v = 35/36 -- protected for chain 0 (EIP-155) -- indistinguishable from the unprotected v = 27/28. A transaction signed for chain 0 was therefore accepted on any chain.

Keep v verbatim instead, the way rlp_encode() writes it and the JSON transaction loader reads it, and derive chain_id next to it: v >= 35 is exactly the EIP-155 form, so Transaction::chain_id_protected() decides the check, and decoding a transaction becomes the inverse of encoding it. Requiring the whole v to fit uint64_t bounds a legacy chain id to 2**63 - 18, the limit the loader already had.

A state test identifies its sender directly and usually carries no signature, so the exported transaction now also carries v; without it a re-executed test would take every legacy transaction for an unprotected one.

### DIFF
--- a/test/state/state.cpp
+++ b/test/state/state.cpp
@@ -438,10 +438,7 @@ std::variant<TransactionProperties, std::error_code> validate_transaction(
     const StateView& state_view, const BlockInfo& block, const Transaction& tx, evmc_revision rev,
     int64_t block_gas_left, int64_t blob_gas_left) noexcept
 {
-    // chain_id must match the chain; a legacy tx may use 0 (unspecified, EIP-155).
-    // TODO: chain_id 0 is not "unspecified" for a legacy tx protected for chain 0 (v 35/36);
-    //   telling it apart from an unprotected tx (v 27/28) needs the signature v from txbytes.
-    if (tx.chain_id != block.chain_id && (tx.type != Transaction::Type::legacy || tx.chain_id != 0))
+    if (tx.chain_id_protected() && tx.chain_id != block.chain_id)
         return make_error_code(INVALID_CHAIN_ID);
 
     switch (tx.type)  // Validate "special" transaction types.
--- a/test/state/transaction.cpp
+++ b/test/state/transaction.cpp
@@ -8,8 +8,6 @@
 
 namespace evmone::state
 {
-using intx::uint256;
-
 bool decode(bytes_view& from, Authorization& to) noexcept
 {
     bytes_view payload;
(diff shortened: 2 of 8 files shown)


test/state/state.cpp
```diff
@@ -438,10 +438,9 @@ std::variant<TransactionProperties, std::error_code> validate_transaction(
     const StateView& state_view, const BlockInfo& block, const Transaction& tx, evmc_revision rev,
     int64_t block_gas_left, int64_t blob_gas_left) noexcept
 {
-    // chain_id must match the chain; a legacy tx may use 0 (unspecified, EIP-155).
-    // TODO: chain_id 0 is not "unspecified" for a legacy tx protected for chain 0 (v 35/36);
-    //   telling it apart from an unprotected tx (v 27/28) needs the signature v from txbytes.
-    if (tx.chain_id != block.chain_id && (tx.type != Transaction::Type::legacy || tx.chain_id != 0))
+    // chain_id must match the chain, unless the transaction is not signed for any chain in
```
chfast: Remove this comment.

test/state/transaction.cpp
```diff
@@ -97,34 +95,19 @@ namespace
...
+        // the way rlp_encode() writes it, and the chain id derived alongside. Requiring the whole
+        // v to fit uint64_t bounds the chain id to 2**63 - 18, the same limit the JSON transaction
+        // loader has.
+        if (!rlp::decode(body, to.v))
             return false;
 
-        if (v_u256 == 27 || v_u256 == 28)  // Pre-EIP-155: no chain id.
-        {
-            to.v = (v_u256 == 28) ? 1 : 0;
+        if (to.v == 27 || to.v == 28)  // Pre-EIP-155: not bound to any chain.
```
chfast: Do we actually need to set chain_id to 0 here? this transaction is not protected so tx.chain_id should be ignored.

test/state/transaction.hpp
```diff
@@ -55,6 +55,13 @@ struct Transaction
     /// Returns amount of blob gas used by this transaction
     [[nodiscard]] uint64_t blob_gas_used() const { return GAS_PER_BLOB * blob_hashes.size(); }
 
+    /// Whether the signature covers the chain id: always for a typed transaction, which has an
```
chfast:
```suggestion
    /// Whether the transaction specifies expected chain id. Always true for typed transactions.
```

test/state/transaction.hpp
```diff
@@ -70,15 +77,20 @@ struct Transaction
     uint64_t nonce = 0;
     intx::uint256 r;
     intx::uint256 s;
+
+    /// The signature v as it appears on the wire: the y parity for a typed transaction, and for a
```
chfast:
```suggestion
    /// The verbatim v value of the signature. It encodes y_parity and for legacy transactions chain id.
```

test/state/transaction.hpp
```diff
@@ -70,15 +77,20 @@ struct Transaction
...
     uint64_t v = 0;
     AuthorizationList authorization_list;
 };
 
 /// Decodes a transaction from its complete serialization @p data.
 ///
 /// Handles the legacy RLP list and the EIP-2718 typed envelope (type byte followed by an RLP list).
-/// TODO: Not a strict inverse of rlp::encode: a legacy transaction's wire v is normalized into
-///   (chain_id, y_parity), so re-encoding a decoded legacy tx need not reproduce its bytes.
+/// The inverse of rlp_encode(), except that a legacy or access-list transaction has a single wire
```
chfast: This comment doesn't make sense: the exact representation in Transaction is not relevant here.

test/utils/rlp_encode.cpp
```diff
@@ -22,8 +22,8 @@ namespace evmone::state
     if (tx.type == Transaction::Type::legacy)
     {
         // rlp [nonce, gas_price, gas_limit, to, value, data, v, r, s];
-        // v is written verbatim. TODO: chain_id 0 is ambiguous — unprotected (v 27/28) vs
-        //   protected for chain 0 (v 35/36) — so v must not be re-derived from chain_id.
+        // v is written verbatim: it carries the chain id since EIP-155, and chain id 0 alone is
```
chfast: Remove this comment.

test/utils/statetest_export.cpp
```diff
@@ -122,6 +122,7 @@ json::json to_state_test(std::string_view test_name, const state::BlockInfo& blo
     jtx["secretKey"] = hex0x(SenderSecretKey);
     jtx["chainId"] = hex0x(tx.chain_id);
     jtx["nonce"] = hex0x(tx.nonce);
+    jtx["v"] = hex0x(tx.v);  // Not signed, but v tells whether a legacy tx is bound to a chain.
```
chfast: Remove comment.

test/utils/statetest_loader.cpp
```diff
@@ -479,6 +479,10 @@ static void from_json(const json::json& j, TestMultiTransaction& o)
 
     for (const auto& j_value : j.at("value"))
         o.values.emplace_back(from_json<intx::uint256>(j_value));
+
+    // A state test identifies the sender directly, so the signature is usually absent; when it is
```
chfast: Remove comment.

PR #1618, opened 2026-08-01
crypto: Use fixed-window exponentiation in modexp

AskAlexSharov:
## Summary

`modexp_odd` uses binary square-and-multiply — one Montgomery multiply per set exponent bit. For large exponents that is roughly twice the multiplies a windowed method needs.

This precomputes a small table of base powers (`b^1 .. b^(2^w - 1)` in Montgomery form) and consumes `w` exponent bits per multiply. The window width scales with the exponent size so the table cost stays amortized even for a sparse exponent:

- `w = 1` (plain binary, no table) for exponents ≤ 16 bits;
- `w = 2` up to 48 bits, `w = 3` up to 144 bits, `w = 4` above.

With `w = 1` the loop is byte-for-byte the previous binary square-and-multiply.

## Benchmarks

`evmone-precompiles-bench --benchmark_filter='modexp<expmod_execute_evmone>'`, AMD EPYC 4344P, gcc 15.2.0, Release:

| case | before | after | speedup |
|---|--:|--:|--:|
| mod_len:32 / exp_bits:256 | 19,604 ns | 12,904 ns | 1.52x |
| mod_len:32 / exp_bits:8192 | 650,488 ns | 419,503 ns | 1.55x |
| mod_len:504 / exp_bits:255 | 2,868,036 ns | 1,869,504 ns | 1.53x |
| mod_len:512 / exp_bits:8192 | 96,520,499 ns | 60,619,147 ns | 1.59x |
| mod_len:32 / exp_bits:33 | 2,582 ns | 2,089 ns | 1.24x |

Small exponents (≤ 16 bits) are unchanged; none regress.

## Cost

The power table adds `MODEXP_TABLE_MAX * n` words to the stack scratch buffer (`STACK_CAPACITY` and the `modexp_odd` scratch requirement are updated). At the EIP-7823 limit (`n = 128` words) that is ~15 KB of additional stack in the single modexp frame.

The window widths and thresholds are simple, conservative choices; happy to tune them or switch to a sliding window (odd-power table, ~half the entries) if preferred.

## Correctness

Adds `expmod.windowing_vs_gmp`: a differential test comparing evmone against GMP across many exponent bit-lengths (crossing the window-width thresholds) and bit patterns (all window values), for odd and even moduli. Existing `expmod` vectors and `large_inputs` continue to pass.

### DIFF
--- a/lib/evmone_precompiles/modexp.cpp
+++ b/lib/evmone_precompiles/modexp.cpp
@@ -431,26 +431,25 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
     const auto exp_loop = [&]<size_t N>() {
         auto r_cur = std::span<uint64_t, N>{result};
         auto r_tmp = std::span<uint64_t, N>{u.first(n)};
-        const auto bm = std::span<const uint64_t, N>{base_mont};
         const auto m = std::span<const uint64_t, N>{mod};
 
-        // The j-th precomputed value.
-        const auto precomputed = [table, n](size_t j) noexcept {
-            return std::span<uint64_t, N>{table.subspan(j * n, n)};
+        // base_mont^k, for k in 1..table_size.
+        const auto precomputed = [table, n](size_t k) noexcept {
+            return std::span<uint64_t, N>{table.subspan((k - 1) * n, n)};
         };
 
-        // precomputed[j] = base_mont^(j+1); precomputed[0] = base_mont is already set.
-        for (size_t j = 1; j < table_size; ++j)
-            mul_amm<N>(precomputed(j), precomputed(j - 1), bm, m, mod_inv);
+        // precomputed(1) = base_mont is already set.
+        for (size_t k = 2; k <= table_size; ++k)
+            mul_amm<N>(precomputed(k), precomputed(k - 1), precomputed(1), m, mod_inv);
 
-        // Reads the w-bit (or fewer) window whose most-significant bit is at index `hi`.
+        // Reads the `width` exponent bits starting at index `lo`.
         // TODO: A window spans at most two adjacent bytes, so it could be read with one
         //   two-byte load, a shift and a mask. Est. 1-3%, and only for a 4-word modulus
         //   with a very long exponent; measure before doing it.
-        const auto window = [&](size_t hi, size_t width) noexcept {
+        const auto window = [&](size_t lo, size_t width) noexcept {
             size_t v = 0;
             for (size_t b = 0; b < width; ++b)
-                v = (v << 1) | (exp[hi - b] ? size_t{1} : size_t{0});
+                v |= size_t{exp[lo + b]} << b;
             return v;
         };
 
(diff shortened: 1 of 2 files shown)


lib/evmone_precompiles/modexp.cpp
```diff
@@ -368,8 +368,37 @@ template <>
...
+/// bit, at the cost of a 2^w - 1 entry table. w == 1 is plain square-and-multiply.
+/// The thresholds are the break-even points for a random exponent,
+/// exp_bits = 2^w / ((1-2^-w)/w - (1-2^-(w+1))/(w+1)) ≈ 16, 48, 140 (rounded up to 144).
+///
+/// TODO: Switch to a sliding window, as GMP's mpn_powm and OpenSSL's BN_mod_exp_mont do:
+///   the table then holds only odd powers, half the entries per width. Measured ~3-7%.
+/// TODO: Tune for the densest exponent instead of the average, because gas is charged on
+///   exponent bit length and ignores Hamming weight. The width then collapses to
+///   min(MAX_WINDOW_WIDTH, (bit_width(exp_bits) + 1) / 2). Measured +10.7% worst case.
+constexpr unsigned window_width(size_t exp_bits) noexcept
```
chfast, line 383: Reverse if order.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -368,8 +368,37 @@ template <>
...
+        return 3;
+    if (exp_bits > 16)
+        return 2;
+    return 1;
+}
+
 /// Computes result[] = base[]^exp % mod[] for odd mod[] (mod[0] % 2 != 0).
-/// Scratch space required: 4n + 3*base.size() + 2 words, where n = mod.size().
+/// Scratch space required: (MAX_PRECOMPUTED + 3)*n + 3*base.size() + 2 words,
+/// where n = mod.size().
```
chfast: No point for n var if used once. Move formula to new line if doesn't fit

lib/evmone_precompiles/modexp.cpp
```diff
@@ -380,17 +409,24 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
 
     const auto n = mod.size();
     const auto mod_inv = -evmmax::modinv(mod[0]);
+    const auto exp_bits = exp.bit_width();
 
-    // Layout: u[n+base.size()] | base_mont[n] | t/rem_scratch[max(n, 2*(n+base.size())+2)]
-    // t and rem_scratch share the same region (exclusive lifetimes).
-    assert(scratch.size() >= 4 * n + 3 * base.size() + 2);
+    const auto w = window_width(exp_bits);
+    const size_t table_size = (size_t{1} << w) - 1;
```
chfast: auto

lib/evmone_precompiles/modexp.cpp
chfast: Done in 3c15cf85 — conditions now ascend (`<= 16` → 1, `<= 48` → 2, `<= 144` → 3, else `MAX_WINDOW_WIDTH`). Verified equivalent to the old chain for every `exp_bits` in 0..19999.

lib/evmone_precompiles/modexp.cpp
chfast: Done — inlined as `(MAX_PRECOMPUTED + 3)*mod.size() + 3*base.size() + 2 words.` It fits on one line (87 cols), so no wrap needed.

lib/evmone_precompiles/modexp.cpp
chfast: Done — `const auto table_size = (size_t{1} << w) - 1;`

chfast (approved):

PR #1622, opened 2026-08-09
test: Unify the JSON loaders' optional-key helper

chfast:
The state test and blockchain test loaders each carried their own `load_if_exists()`, which returned a default-constructed value for an absent key and so conflated "absent" with "zero": `BlockInfo::blob_gas_used` is a `std::optional<uint64_t>`, so an env without `blobGasUsed` ended up with an engaged optional holding 0. Both copies are replaced by a shared `load_optional()` in `statetest.hpp` for the genuinely optional fields and `load_or(j, key, default)` where the default belongs at the call site (`{}` for hashes and addresses, `0`/`1` for numbers). Single-value keys previously read through a `find()`/`end()` pair go through them as well, which drops the `contains()`+`at()` double lookups and the hand-rolled hex decoding of the t8n transaction hash.

### DIFF
--- a/test/utils/statetest_loader.cpp
+++ b/test/utils/statetest_loader.cpp
@@ -14,17 +14,6 @@ namespace evmone::test
 namespace json = nlohmann;
 using evmc::from_hex;
 
-namespace
-{
-template <typename T>
-T load_if_exists(const json::json& j, std::string_view key)
-{
-    if (const auto it = j.find(key); it != j.end())
-        return from_json<T>(*it);
-    return {};
-}
-}  // namespace
-
 template <>
 uint8_t from_json<uint8_t>(const json::json& j)
 {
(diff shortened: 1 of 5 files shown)


test/utils/statetest_loader.cpp
```diff
@@ -287,43 +258,35 @@ state::BlockInfo from_json_with_rev(
...
         excess_blob_gas = state::calc_excess_blob_gas(rev, blob_params, parent_blob_gas_used,
-            parent_excess_blob_gas, parent_base_fee, parent_blob_base_fee);
-    }
-    else if (const auto it2 = j.find("currentExcessBlobGas"); it2 != j.end())
-    {
-        excess_blob_gas = from_json<uint64_t>(*it2);
+            *parent_excess_blob_gas, parent_base_fee, parent_blob_base_fee);
     }
+    else
+        excess_blob_gas = load_optional<uint64_t>(j, "currentExcessBlobGas").value_or(0);
```
chfast: also use {} here.

test/utils/statetest_loader.cpp
```diff
@@ -526,10 +483,9 @@ static void from_json(const json::json& j_t, StateTransitionTest& o)
     uint64_t chain_id = 1;
     if (const auto config_it = j_t.find("config"); config_it != j_t.end())
     {
-        if (const auto bs_it = config_it->find("blobSchedule"); bs_it != config_it->end())
-            o.blob_schedule = from_json<BlobSchedule>(*bs_it);
-        if (const auto cid_it = config_it->find("chainid"); cid_it != config_it->end())
-            chain_id = from_json<uint64_t>(*cid_it);
+        if (auto blob_schedule = load_optional<BlobSchedule>(*config_it, "blobSchedule"))
+            o.blob_schedule = *std::move(blob_schedule);
```
chfast: Is this move variant necessary for blob_schedule?

test/utils/statetest_loader.cpp
chfast: Both of these turned out to be the same problem. `value_or({})` does not compile on the minimum toolchains (gcc-13's libstdc++): `value_or`'s parameter is a deduced `U&&`, so a braced initializer is a non-deduced context — the defaulted `U = remove_cv_t<T>` that makes it work only arrived in newer standard libraries, which is why it built locally and on clang-latest but broke clang-min/gcc-min/macos/riscv32.

So the defaults now go through `load_or(j, key, default)`, which takes the default by value and accepts `{}` everywhere. This line is numeric, so per your other message it stays `load_or<uint64_t>(j, "currentExcessBlobGas", 0)`; the hash/address/bytes32 ones read `load_or<hash256>(j, "parentUncleHash", {})`. Say the word if you want `{}` for the numerics too.

test/utils/statetest_loader.cpp
chfast: No — dropped. It saved a copy of a 1-3 entry map once per test file, which is not worth the odd spelling. Now plain `*blob_schedule`. (The one `std::move` left is for `postState`, a whole TestState, and it is spelled `std::move(*post_state)`.)

test/utils/statetest_loader.cpp
```diff
@@ -287,43 +261,35 @@ state::BlockInfo from_json_with_rev(
...
-        .parent_ommers_hash = parent_uncle_hash,
+        .difficulty = load_or<int64_t>(j, "currentDifficulty", 0),
+        .parent_difficulty = load_or<int64_t>(j, "parentDifficulty", 0),
+        .parent_ommers_hash = load_or<hash256>(j, "parentUncleHash", {}),
         .prev_randao = prev_randao,
-        .parent_beacon_block_root = load_if_exists<hash256>(j, "parentBeaconBlockRoot"),
+        .parent_beacon_block_root = load_or<hash256>(j, "parentBeaconBlockRoot", {}),
         .base_fee = base_fee,
-        .blob_gas_used = load_if_exists<uint64_t>(j, "blobGasUsed"),
+        .blob_gas_used = load_optional<uint64_t>(j, "blobGasUsed"),
```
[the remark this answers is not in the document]
chfast: Done, `statetest_loader.block_info_blob_gas_used` asserts both directions: absent leaves the optional disengaged, `"blobGasUsed": "0"` engages it with zero. Confirmed it is a real guard by restoring the old `load_or<uint64_t>(j, "blobGasUsed", 0)` behaviour, which makes it fail.

PR #1631, opened 2026-08-10
crypto: Use a sliding window in modexp

chfast:
Follow-up to #1618.

The fixed window precomputes all powers `b^1..b^(2^w-1)` — the layout a
constant-time implementation needs, because sliding windows leak through
data-dependent access. This code is not constant-time and EVM exponents are public
calldata, so the table was twice the size it had to be for no return.

Now only the odd powers are stored, and each window slides onto a set bit with its
trailing zeros trimmed so the value stays odd. Zero runs then cost one squaring
each, and half the table per width buys one more width within the same scratch.
Windows are read with a single two-byte access instead of one call per bit, then
trimmed with `countr_zero`.

A few percent fewer instructions over the benchmark matrix and over a corpus of
unique mainnet inputs, up to about 17% on small exponents. One case regresses by
about 1%, `mod_len=32 / mod_tz=254 / exp_bits=256`: the odd part of its modulus is a
single word, so the wider table is not repaid. That is a width-selection limit —
`window_width()` sees only `exp_bits`, not how cheap a multiply is for this modulus.

### DIFF
--- a/lib/evmone_precompiles/modexp.cpp
+++ b/lib/evmone_precompiles/modexp.cpp
@@ -292,6 +292,19 @@ public:
         const auto bit = (byte >> bit_index) & 1;
         return bit != 0;
     }
+
+    /// Returns bits [lo, hi] as an integer, the bit at hi being the most significant.
+    /// The range must span at most 8 bits, so it covers at most two adjacent bytes.
+    [[nodiscard]] size_t window(size_t lo, size_t hi) const noexcept
+    {
+        assert(lo <= hi && hi - lo < 8);
+        const auto exp_size = (bit_width_ + 7) / 8;
+        const auto byte_index = exp_size - 1 - lo / 8;
+        auto bytes = size_t{data_[byte_index]};
+        if (byte_index != 0)  // Prepend the next more significant byte if there is one.
+            bytes |= size_t{data_[byte_index - 1]} << 8;
+        return (bytes >> (lo % 8)) & ((size_t{1} << (hi + 1 - lo)) - 1);
+    }
 };
 
 /// Performs the Almost Montgomery Multiplication (AMM).
@@ -369,27 +382,26 @@ template <>
 }
 
 /// Maximum window width used by the windowed method in modexp_odd.
-constexpr unsigned MAX_WINDOW_WIDTH = 4;
+constexpr unsigned MAX_WINDOW_WIDTH = 5;
+static_assert(MAX_WINDOW_WIDTH <= 8, "Exponent::window() covers at most two adjacent bytes");
 
-/// Number of precomputed values for the max width windowed method.
-constexpr size_t MAX_PRECOMPUTED = (size_t{1} << MAX_WINDOW_WIDTH) - 1;
+/// Number of precomputed base odd powers for the max width windowed method.
+constexpr size_t MAX_PRECOMPUTED = size_t{1} << (MAX_WINDOW_WIDTH - 1);
 
-/// Selects the fixed-window width from the exponent bit length.
-///
-/// TODO: Switch to a sliding window: the table then holds only odd powers. Measured ~3-7%.
-/// TODO: Tune for the densest exponent instead of the average, because gas is charged on
-///   exponent bit length and ignores Hamming weight. The width then collapses to
-///   min(MAX_WINDOW_WIDTH, (bit_width(exp_bits) + 1) / 2). Measured +10.7% worst case.
+/// Selects the sliding-window width from the exponent bit length.
 constexpr unsigned window_width(size_t exp_bits) noexcept
 {
-    // Break-even points for a random exponent, where the 2^w extra table multiplies stop
-    // being repaid: 2^w / ((1-2^-w)/w - (1-2^-(w+1))/(w+1)) = 16, 48, 140 (rounded to 144).
-    if (exp_bits <= 16)
+    // Break-even points for a random exponent, where the table's extra multiply stops being
+    // repaid: 2^(w-1) / (1/(w+1) - 1/(w+2)) = 6, 24, 80, 240. Each narrower width is kept
+    // one bit longer, which measures better on the sparse small exponents seen in practice.
+    if (exp_bits <= 7)
         return 1;
-    if (exp_bits <= 48)
+    if (exp_bits <= 25)
         return 2;
-    if (exp_bits <= 144)
+    if (exp_bits <= 81)
         return 3;
+    if (exp_bits <= 241)
+        return 4;
     return MAX_WINDOW_WIDTH;
 }
 
@@ -408,7 +420,7 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
     const auto exp_bits = exp.bit_width();
 
     const auto w = window_width(exp_bits);
-    const auto table_size = (size_t{1} << w) - 1;
+    const auto table_size = size_t{1} << (w - 1);
 
     // Layout: u[n + base.size()] | table[MAX_PRECOMPUTED*n]
     //       | rem_scratch[2*n + 2*base.size() + 2].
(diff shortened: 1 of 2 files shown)


test/unittests/precompiles_expmod_test.cpp
```diff
@@ -291,58 +291,36 @@ TEST_P(expmod, inputs)
...
-        {"03", "04013929f7999c", "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f",
-            "06f41e370c4ef45a2bc5e1ade1504fbe35e5a42a8f8c2b17ad16a6c657900d48"},
-        // exp_bits=144, w=3: last exponent size before w becomes 4.
-        {"03", "8004cb3ff13151bb9f84a488a5d62e79a680",
+        // Sliding-window exponentiation in modexp_odd. One case per window width w=1..5.
+        // Each exponent is built as: top bit (1) | zero run of w+1 bits | one run of w bits
+        // | trailing zeros, so its windows exercise both the first (b^1) and last
+        // (b^(2^w-1)) precomputed odd powers, with a zero run in between wide enough to
+        // skip a whole window's worth of squarings without a multiply. Modulus is the
+        // secp256k1 field prime: odd, 4 words, so these also cover the mul_amm<4>
```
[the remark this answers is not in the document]
chfast: Correct, and the comment was wrong: the loop squares once per bit unconditionally, so a zero run skips only the multiply. What the w+1 gap actually buys is that the top bit and the one-run cannot fall in the same window, which is what makes the vector hit both b^1 and b^(2^w-1). Reworded to say that.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -433,46 +438,65 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
...
+        if (table_size > 1)
+        {
+            mul_amm<N>(r_tmp, precomputed(1), precomputed(1), m, mod_inv);
+            for (size_t v = 3; v < 2 * table_size; v += 2)
+                mul_amm<N>(precomputed(v), precomputed(v - 2), r_tmp, m, mod_inv);
+        }
+
+        // Scans downward from bit `hi` (which must be set) for the start of the widest
+        // window of at most w bits ending at `hi`, trimming trailing zero bits so the
+        // window value is always odd -- i.e. only odd table entries are ever needed.
```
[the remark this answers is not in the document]
chfast: Right, the loop jumps to the lower bound and increments. Dropped the direction claim; the comment now just states what the lambda returns.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -368,28 +368,33 @@ template <>
     mul_amm_256(r, x, y, mod, mod_inv);
 }
 
-/// Maximum window width used by the windowed method in modexp_odd.
-constexpr unsigned MAX_WINDOW_WIDTH = 4;
+/// Maximum sliding-window width used by modexp_odd. Capped at 5: mainnet's dominant case
```
chfast: Leave the comment as it was. I can see this is capped at 5.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -368,28 +368,33 @@ template <>
...
+/// is 254-bit exponents (the w=5 band below), and width 6 would double the table's cost
+/// (33 vs. 17 words per modulus word) for exponents rare enough not to need it.
+constexpr unsigned MAX_WINDOW_WIDTH = 5;
 
-/// Number of precomputed values for the max width windowed method.
-constexpr size_t MAX_PRECOMPUTED = (size_t{1} << MAX_WINDOW_WIDTH) - 1;
+/// Number of precomputed odd powers b^1, b^3 .. b^(2^w - 1) for the max width.
```
chfast:
```suggestion
/// Number of precomputed base odd powers for the max width windowed method.
```

lib/evmone_precompiles/modexp.cpp
```diff
@@ -368,28 +368,33 @@ template <>
...
+constexpr size_t MAX_PRECOMPUTED = size_t{1} << (MAX_WINDOW_WIDTH - 1);
 
-/// Selects the fixed-window width from the exponent bit length.
+/// Selects the sliding-window width from the exponent bit length.
 ///
-/// TODO: Switch to a sliding window: the table then holds only odd powers. Measured ~3-7%.
 /// TODO: Tune for the densest exponent instead of the average, because gas is charged on
```
chfast: Can we drop this TODO? We decided to tune for the most common cases.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -368,28 +368,33 @@ template <>
...
 ///   exponent bit length and ignores Hamming weight. The width then collapses to
 ///   min(MAX_WINDOW_WIDTH, (bit_width(exp_bits) + 1) / 2). Measured +10.7% worst case.
 constexpr unsigned window_width(size_t exp_bits) noexcept
 {
-    // Break-even points for a random exponent, where the 2^w extra table multiplies stop
-    // being repaid: 2^w / ((1-2^-w)/w - (1-2^-(w+1))/(w+1)) = 16, 48, 140 (rounded to 144).
-    if (exp_bits <= 16)
+    // GMP's win_size() thresholds. The break-even points for a random exponent, where the
```
chfast: Don't mention GMP. Just briefly describe the strategy / method.

lib/evmone_precompiles/modexp.cpp
```diff
@@ -408,7 +413,7 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
     const auto exp_bits = exp.bit_width();
 
     const auto w = window_width(exp_bits);
-    const auto table_size = (size_t{1} << w) - 1;
+    const size_t table_size = size_t{1} << (w - 1);  // Odd powers b^1, b^3, ... b^(2^w - 1).
```
chfast:
```suggestion
    const auto table_size = size_t{1} << (w - 1);
```

lib/evmone_precompiles/modexp.cpp
```diff
@@ -433,46 +438,65 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
         auto r_tmp = std::span<uint64_t, N>{u.first(n)};
         const auto m = std::span<const uint64_t, N>{mod};
 
-        // base_mont^j, for j in 1..table_size.
-        const auto precomputed = [table, n](size_t j) noexcept {
-            return std::span<uint64_t, N>{table.subspan((j - 1) * n, n)};
+        // precomputed(v) = base_mont^v, for odd v in 1 .. 2*table_size - 1.
```
chfast:
```suggestion
        // base_mont^v, for odd v.
```

lib/evmone_precompiles/modexp.cpp
```diff
@@ -433,46 +438,65 @@ void modexp_odd(std::span<uint64_t> result, std::span<const uint64_t> base, Expo
...
+        if (table_size > 1)
+        {
+            mul_amm<N>(r_tmp, precomputed(1), precomputed(1), m, mod_inv);
+            for (size_t v = 3; v < 2 * table_size; v += 2)
+                mul_amm<N>(precomputed(v), precomputed(v - 2), r_tmp, m, mod_inv);
+        }
+
+        // Start of the widest window of at most w bits ending at bit `hi` (which must be
+        // set), with trailing zero bits trimmed so the window value is always odd -- i.e.
+        // only odd table entries are ever needed.
```
chfast:
```suggestion
```

test/unittests/precompiles_expmod_test.cpp
```diff
@@ -291,58 +291,36 @@ TEST_P(expmod, inputs)
...
-            "651aace134976d8456fcc35686a57cf12670b2e596dabecd0ddae9984ced96c4"},
-        // exp_bits=50, w=3, leading window 2 bits.
-        {"03", "0200a6a7ef231d", "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f",
-            "f722a91e1faa3b57f0a19af8d4506b395a0a342e9ee2cbe65cd7a63155d38537"},
-        // exp_bits=51, w=3, leading window 3 bits.
-        {"03", "04013929f7999c", "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f",
-            "06f41e370c4ef45a2bc5e1ade1504fbe35e5a42a8f8c2b17ad16a6c657900d48"},
-        // exp_bits=144, w=3: last exponent size before w becomes 4.
-        {"03", "8004cb3ff13151bb9f84a488a5d62e79a680",
+        // Sliding-window exponentiation in modexp_odd. One case per window width w=1..5.
```
chfast, line 294: This should not remove pre-existing cases. Add new ones if needed.

lib/evmone_precompiles/modexp.cpp
chfast: Restored the original one-liner.

lib/evmone_precompiles/modexp.cpp
chfast: Applied.

lib/evmone_precompiles/modexp.cpp
chfast: Dropped, here and from the follow-up notes.

lib/evmone_precompiles/modexp.cpp
chfast: Rewritten to state the method only: the break-even points and why each narrower width is kept one bit longer.

lib/evmone_precompiles/modexp.cpp
chfast: Applied.

lib/evmone_precompiles/modexp.cpp
chfast: Applied.

lib/evmone_precompiles/modexp.cpp
chfast: Dropped the clause. Removing only that line would have left a dangling "-- i.e." on the line above, so I took the whole sentence; the comment is now two lines. If you meant the entire comment, say so and it goes.

test/unittests/precompiles_expmod_test.cpp
chfast: All 13 restored, values untouched. Their comments could not survive as-is -- they described the leading partial window ((exp_bits - 1) % w + 1), which this PR removes -- so they are now one grouped comment naming the bands they land in: exp_bits 16..18 (w=2), 48..51 (w=3), 144..148 (w=4), the last repeating 148 with a 5-word modulus. Mappings checked against window_width() rather than assumed.

Coverage is better than either set alone: widths 1/2/3/4/5 at 1/4/5/7/2 cases, and the generic mul_amm instantiation now runs at both w=4 (restored case) and w=5 (new case) instead of only w=5.

PR #1634, opened 2026-08-10
test: Recover the blockchain test transaction senders

chfast:
The runner trusted the sender the fixture supplies, so a transaction whose signature the test broke executed as that sender and the block was rejected by whatever rule it happened to break. Recover from the transaction's own encoding instead, and reject it when the signature does not recover. This removes the exemption added in #1632 and the TODO from #1623; the two fixtures added there now exercise the recovery.

`recover_sender()` also had to reject a `v` outside its domain, a rule only `decode_transaction()` enforced and the JSON path never runs. Without it the `v=34` cases of `frontier/validation/bad_v_r_s` recover a stranger's address and fail on its empty balance instead of on the signature — 13 blockchain fixtures.

Recovery is opt-in via `BlockTransitionOptions`, so t8n keeps using the sender it is given.

### DIFF
--- a/test/state/transaction.cpp
+++ b/test/state/transaction.cpp
@@ -164,6 +164,8 @@ std::optional<address> recover_sender(const Transaction& tx, bytes_view txbytes)
 
     // A typed v is {0, 1}. A legacy v is 27 + y_parity, or 35 + 2 * chain_id + y_parity (EIP-155):
     // both bases are odd, so an even v means y_parity 1.
+    if (typed ? tx.v > 1 : (tx.v < 27 || (tx.v > 28 && tx.v < 35)))
+        return std::nullopt;
     const auto y_parity = typed ? tx.v != 0 : tx.v % 2 == 0;
 
     const auto h = keccak256((typed ? bytes{stdx::to_underlying(tx.type)} : bytes{}) +
--- a/test/blockchaintest/blockchaintest_runner.cpp
+++ b/test/blockchaintest/blockchaintest_runner.cpp
@@ -346,22 +346,15 @@ void run_blockchain_tests(std::span<const BlockchainTest> tests, evmc::VM& vm)
                 const auto names_spec_exception =
                     test_block.expected_exception.find("Exception.") != std::string::npos;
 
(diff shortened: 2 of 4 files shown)


test/state/transaction.cpp
```diff
@@ -164,6 +164,8 @@ std::optional<address> recover_sender(const Transaction& tx, bytes_view txbytes)
 
     // A typed v is {0, 1}. A legacy v is 27 + y_parity, or 35 + 2 * chain_id + y_parity (EIP-155):
     // both bases are odd, so an even v means y_parity 1.
+    if (typed ? tx.v > 1 : (tx.v < 27 || (tx.v > 28 && tx.v < 35)))
```
chfast, line 167: Why this is here now?

chfast: Because the JSON path never decodes, and this is the only function that interprets `v`.

`decode_transaction()` bounds `v` while parsing (transaction.cpp:109-112 for legacy, :128 for typed), so a transaction that came from RLP can never reach here out of domain. A blockchain-test transaction is built field by field from `rlp_decoded` JSON, which applies no such bound, and this PR is the first caller to recover a sender from one. Without the check the `v=34` cases of `frontier/validation/bad_v_r_s` derive `y_parity = 1` from an even `v`, recover a stranger's address, and get rejected for its empty balance — `got INSUFFICIENT_ACCOUNT_FUNDS, expected INVALID_SIGNATURE_VRS`, 13 fixtures.

It sits immediately above the `y_parity` derivation because that is what it guards: the line below assumes `v` is one of the forms the comment describes.

It does duplicate the decoder's bound for now. The intended end state is the reverse — drop it from `decode_transaction()` so an out-of-range `v` is reported as `INVALID_SIGNATURE` rather than `INVALID_ENCODING`, which is what would let the over-broad `{INVALID_ENCODING, "TransactionException.INVALID_SIGNATURE_VRS"}` alternative come out of `error_matching.cpp`. That changes state-test behaviour, so I kept it out of this PR. Happy to fold it in if you would rather have it in one go.

PR #1638, opened 2026-08-10
crypto: Share the Fq12 accumulator across pairing pairs

chfast:
pairing_check ran an independent Miller loop per pair and multiplied the
results, paying 64 Fq12 squarings each. The recurrence f ← f²·line is
multiplicative across pairs, so a single accumulator serves all of them
and the squaring is paid once per iteration rather than once per pair.

For N pairs this drops (N−1)×64 squarings plus the N−1 Fq12
multiplications that combined the per-pair results, and lets an
all-infinity input skip the final exponentiation. About 13% off the
ECPAIRING instruction count on the benchmark inputs, which carry 2 to 4
pairs.

Supersedes the draft #1544, reimplemented rather than rebased because it
predates the JacPoint rename and the Point removal. One caveat: the
per-pair state needs a vector, so a single-pair call pays an allocation
and saves no squarings.

### DIFF
--- a/lib/evmone_precompiles/pairing/bn254/pairing.cpp
+++ b/lib/evmone_precompiles/pairing/bn254/pairing.cpp
@@ -5,6 +5,8 @@
 #include "../../bn254.hpp"
 #include "fields.hpp"
 #include "utils.hpp"
+#include <span>
+#include <vector>
 
 namespace evmmax::bn254
 {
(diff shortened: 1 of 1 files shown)


chfast:
Superseded by #1544, which now carries this implementation. Same change; #1544 is the original PR for the idea.

PR #1644, opened 2026-08-10
crypto: Multiply Fq2 with the Karatsuba method

chfast:
multiply(Fq2) was schoolbook, 4 base field multiplications, while Fq6 and
Fq12 already used Karatsuba. At Fq2 the trade is one multiplication for
three additions, which pays because a Montgomery multiplication costs
several times an addition.

About 6% off the ECPAIRING instruction count, measured with paired
interleaved A/B runs of both binaries.

A TODO records the step left on the table: holding the three products at
double width and reducing once per output coefficient instead of after
every multiplication. That needs a wide multiplication and a standalone
Montgomery reduction in ModArith.


chfast:
Reverted because causes performance regressions.

PR #1683, opened 2026-08-29
test: Add --ignore and --collect-only and use them in CI

chfast:
Both flags follow pytest. `--ignore` drops a path from the collection; it is repeatable and matches
whole path components, so `--ignore bc4895` keeps `bc4895-withdrawals`. The path is relative to the
test directory being searched rather than to the working directory as pytest resolves it, which is
what keeps the CI exclusions below one word each. `--collect-only` prints the path of each collected
test, one per line: the file for a test collected from a directory, `path::case` for a file given
directly.

Every path CI excluded through `--gtest_filter` is a plain path, so the patterns port one for one,
and `--collect-only` replaces `--gtest_list_tests` in the three listing tests, which now pin the
collection order across roots instead of gtest's regrouping by suite name. `ethereum-tests` runs the
same 479 ValidBlocks and 221 InvalidBlocks tests as on master, and the execution-spec-tests jobs the
same 8281 state and 8731 blockchain tests.

The last commit drops statetest's built-in skip list: those five names are legacy GeneralStateTests
suites CI has not run since it moved to evmone-blockchaintest, and the list was inert anyway because
the jobs passed `--gtest_filter='*'` over it. After this nothing in the repo passes a `--gtest_*`
flag except `evmone-unittests`, which keeps gtest.


test/blockchaintest/blockchaintest.cpp
```diff
@@ -131,6 +146,18 @@ int main(int argc, char* argv[])
...
+        app.add_option("--ignore", ignored,
+               "Path, relative to a test directory, not to collect tests from. May be given more "
+               "than once. Whole path components are matched, so --ignore bc4895 keeps "
+               "bc4895-withdrawals.")
+            // Without this the option is variadic and swallows the positional paths after it.
+            ->allow_extra_args(false);
+
+        bool collect_only = false;
+        app.add_flag("--collect-only", collect_only,
+            "List the id of each test that would run, one per line, and exit.");
```
chfast: it lists path, not id

test/blockchaintest/blockchaintest.cpp
```diff
@@ -10,6 +10,7 @@
 #include <test/utils/test_files.hpp>
 #include <test/utils/test_report.hpp>
 #include <iostream>
+#include <span>
```
chfast: Skip this include

test/statetest/statetest.cpp
```diff
@@ -10,6 +10,7 @@
 #include <test/utils/test_files.hpp>
 #include <test/utils/test_report.hpp>
 #include <iostream>
+#include <span>
```
chfast: Skip this include

PR #1697, opened 2026-09-04
ci: Ask CodSpeed for allocation measurements too

chfast:
An experiment, to see what CodSpeed's memory mode reports for the baseline VM.

The mode is chosen when the harness is compiled, not when it runs. The first attempt here just added `memory` to the action's `mode` list against the existing simulation build: the runner accepted it, ran twice and uploaded two result sets, and CodSpeed then reported "No results found during the execution". `CODSPEED_MODE` takes one value, and although `instrumentation`, `simulation` and `memory` all compile to the same `CODSPEED_ANALYSIS`, the memory harness behaves differently at run time — it skips the warmup and reports through the allocation hooks.

So the allocation run gets a build of its own, and the two simulation runs are untouched.

Worth watching: whether the extra configure and build cost much wall time, and whether the numbers say anything the instruction counts do not.

https://claude.ai/code/session_016UHPAGwcwXMjqhTLpT31K7

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -13,6 +13,10 @@ on:
   # performance analysis in order to generate initial data.
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 env:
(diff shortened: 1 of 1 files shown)


chfast:
Not worth currently for current benchmarks as memory shows at most single allocation.

PR #1698, opened 2026-09-04
t8n: Report the block gas used that apply_block computes

chfast:
The block's `gasUsed` is the pre-refund transaction gas on Amsterdam ([EIP-7778](https://eips.ethereum.org/EIPS/eip-7778)), which `apply_block` already computes and returns, but `t8n()` ignored that value and summed the receipts' post-refund `gas_used` instead — understating the header by the block's total refund.

Earlier revisions are unaffected: there `block_gas_used` is the same sum of `receipt.gas_used` that `cumulative_gas_used` accumulates, so the output is byte-identical.

`blockchaintest_runner` already checks the header against `res.gas_used`; this makes t8n use the same value.

Verified: `ctest` 1282/1282, `tests@v20.0.1` state 8172 passed, blockchain 8612 passed + 6 skipped.

https://claude.ai/code/session_01SCu2xuYf6PPsvREMdvUxcM

### DIFF
--- a/test/unittests/tooling_t8n_test.cpp
+++ b/test/unittests/tooling_t8n_test.cpp
@@ -485,3 +485,35 @@ TEST(tooling_t8n, receipt_status_reports_failure)
     EXPECT_THAT(run_call_to("0xfe", EVMC_SHANGHAI), HasSubstr("\"status\": \"0x0\""));
     EXPECT_THAT(run_call_to("0x00", EVMC_SHANGHAI), HasSubstr("\"status\": \"0x1\""));
 }
+
+TEST(tooling_t8n, block_gas_used_is_pre_refund_from_amsterdam)
+{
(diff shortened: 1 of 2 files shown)


chfast:
Added a regression test in `test/unittests/tooling_t8n_test.cpp` (e9ec12cf).

The callee (`PUSH0 PUSH0 SSTORE`) clears a pre-set storage slot, so the transaction earns a refund and the block gas parts from the receipt's cumulative gas. The test asserts `gasUsed > receipts[0].cumulativeGasUsed` on Amsterdam and `==` on Osaka, which pins the unchanged pre-Amsterdam output too — no magic gas numbers.

Confirmed it fails on the old code (`Expected: (block_gas) > (cumulative), actual: 21204 vs 21204`) and passes with the fix.

PR #1700, opened 2026-09-04
test: Report a file's fixtures one result each

chfast:
A file collected from a directory was one test recording one outcome, so what happened to the fixtures beside the first had nowhere to go: a file holding two failing fixtures reported one line, and a fixture the loader declined disappeared into the file's verdict.

Running a test now returns one result per fixture, and a file which never gets as far as one — unparsable, or refused by the loader — returns the single result which says so. Collection and the counts are unchanged.

```
FAILED  statetest/tests1/SuiteA/test2_multi.json::test_case_1 - state root
FAILED  statetest/tests1/SuiteA/test2_multi.json::test_case_2 - state root
```

Ordered before #1699, which drops the second collection form: naming a file to see its fixtures listed only pays for itself while this is missing.

https://claude.ai/code/session_016UHPAGwcwXMjqhTLpT31K7

### DIFF
--- a/test/blockchaintest/blockchaintest.cpp
+++ b/test/blockchaintest/blockchaintest.cpp
@@ -15,6 +15,13 @@ using evmone::test::TestCase;
 
 namespace
 {
+/// A test which is one case: the driver takes an array of results, and this is the one.
+TestCase one_case(std::string name, std::function<void(evmone::test::TestReport&)> run)
+{
+    return {name,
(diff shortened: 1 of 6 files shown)


chfast:
Not worth its own PR: the collapse in #1699 does not need it, and evmone test wants per-fixture results in its own run_fixture_file, where the loader can decline one fixture at a time. Measured: this saves 334 lines from #1685 only by moving the same scaffolding upstream, and adds ~180 net.

PR #1701, opened 2026-09-07
test: Collect test files as plain paths

chfast:
A collected test file carried the name of its suite (its directory relative to the search root) from the days when that was the gtest suite it got registered under. Nothing has displayed it since gtest went away, and its only remaining reader had to glue the filename back on to recover the path it actually wanted.

So the file is now just its path, and the root goes to the one function that needs a path relative to it. That derivation is textual rather than `fs::relative`, so collecting a tree no longer stats every file.


test/blockchaintest/blockchaintest.cpp
```diff
@@ -23,18 +23,17 @@ bool collect_tests(std::vector<TestCase>& cases, const fs::path& root,
...
         {
             // Loaded when the test runs: loading a whole tree up front costs far more. A
             // load which throws over an unsupported fixture reaches the driver, which skips.
-            cases.push_back(
-                {file.path.string(), [path = file.path, &vm](evmone::test::TestReport& report) {
-                     std::ifstream f{path};
-                     for (const auto& test : evmone::test::load_blockchain_tests(f))
-                         evmone::test::run_blockchain_test(test, vm, report);
-                 }});
+            cases.push_back({file.string(), [path = file, &vm](evmone::test::TestReport& report) {
```
chfast: rename file→path, this will simplify the capture and readability.

chfast: Done in both runners. The capture loses its initialiser, and naming the loop variable after what it holds reads better than `file` for something that is a path:

```cpp
for (const auto& path : files)
{
    cases.push_back({path.string(), [path, &vm](evmone::test::TestReport& report) {
```

#1699 is rebased on top, where the rename is the only difference from what was there before.

PR #1722, opened 2026-09-18
test: Add dumping post state for state test running.

rodiazet:
This PR extends reporting by printing out the state diff after each state test run when `--trace-summary` flag passed.


chfast (commented):
I'd split the two features. the log hash check and output is no-brainer. for the state diff I have some questions.

rodiazet:
Splited. I make them dependent only because I need to have two changes in one branch. They are independent.

chfast (commented):
Some general comments:
1. Do you want to use the "standardized" state diff JSON format?
2. I think this should rather go to the summary JSON as dedicated field next to state root hash and logs hash. The --statediff param would force --summary.
3. We are reworking this code to be testable via unit tests: this applies here, but is not required: it's better to pass `std::ostream` to run_state_test() to be able to capture summary output and test these in unit tests.

rodiazet:
1.  I wanna stay with this evmone format. That one looks kinda overdone.
2. Will merge it with `--trace-summary` output.
3. Will rewrite the tests if it will be possible.

rodiazet:
Redone

tools/evmone/main.cpp
```diff
@@ -185,7 +189,7 @@ int exec_test_cmd(evmc::VM& vm, std::span<const fs::path> paths, evmone::test::R
     // progress row, as a summary would.
     if (trace)
         std::ios::sync_with_stdio(false);
-    opts.trace_summary |= trace;
+    opts.trace_summary |= (trace || opts.state_diff);
```
chfast: This collapse should happen in statetest_runner.cpp: it decides what to output based on options.

test/utils/statetest_runner.cpp
chfast: Here: options.trace_summary || options.state_diff

tools/evmone/main.cpp
```diff
@@ -172,8 +172,12 @@ const CLI::App& setup_test_cmd(
     cmd.add_flag(
         "--collect-only", opts.collect_only, "List each collected test, one per line, and exit.");
     cmd.add_flag("--trace-summary", opts.trace_summary,
-        "Report each state test's execution summary, as --trace also does. Blockchain tests "
-        "have no summary to report.");
+        "Report each state test's execution summary, as --trace also does, including the "
+        "transaction's StateDiff as JSON, one line per test case, on the trace stream (stderr). "
```
chfast: outdated.

tools/evmone/main.cpp
```diff
@@ -172,8 +172,12 @@ const CLI::App& setup_test_cmd(
...
+        "Blockchain tests have no summary to report.");
+    cmd.add_flag("--state-diff", opts.state_diff,
+        "Same as `--trace-summary`, but additionally reports transaction's state diff. "
```
chfast:
```suggestion
        "Report each state test's execution summary (same as `--trace-summary`) and transaction's state diff. "
```

chfast (changes requested):

chfast (approved):

PR #1723, opened 2026-09-18
test: Report log hash in state test run

rodiazet:
This PR adds reporting log-hash in a case of `--trace-summary` flag passed.


test/utils/statetest_runner.cpp
```diff
@@ -72,6 +72,10 @@ void run_state_test(
 
             const auto state_root = state::mpt_hash(state);
 
+            const auto actual_logs_hash = holds_alternative<state::TransactionReceipt>(res) ?
```
chfast:
```suggestion
            const auto logs_hash = holds_alternative<state::TransactionReceipt>(res) ?
```
and use get_if.

test/utils/statetest_runner.cpp
```diff
@@ -72,6 +72,10 @@ void run_state_test(
...
+                                              logs_hash(get<state::TransactionReceipt>(res).logs) :
+                                              logs_hash(std::vector<state::Log>());
```
chfast:
```suggestion
                                              logs_hash({});
```

test/utils/statetest_runner.cpp
```diff
@@ -84,6 +88,7 @@ void run_state_test(
                         std::clog << R"("pass":false,"error":")" << r.status << '"';
                     std::clog << R"(,"gasUsed":"0x)" << std::hex << r.gas_used << R"(",)";
                 }
+                std::clog << R"("logsHash":"0x)" << hex(actual_logs_hash) << R"(",)";
```
chfast: Is `"logsHash"` the name from the JSON test format?

chfast (changes requested):

test/utils/statetest_runner.cpp
rodiazet: no. They use `logs`, but also they use `hash` for `stateRoot`. I can change it to `logs` to make it consistent in this case at least.

chfast (approved):
Needs clang-tidy fix, but looks good otherwise.

